PR #3, opened 2020-10-21
Separate out environmental variables, improve initial page load test

plocket:
Had to change what was happening a bit because more was relying on them than I thought. Also, the initial page load test needed tweaking so that the file name didn't have to be provided. I think. I'm currently not sure exactly what was going on there, though it was clear to me yesterday.

It now:
1. Makes sure a `test.yml` file is generated
1. Uses that `test.yml` to make sure the server has really recovered from the `pull`

The downside is that if the developer has a custom `test.yml`, they will have to make sure it's is working.

### DIFF
--- a/lib/github_integration/install-branch.js
+++ b/lib/github_integration/install-branch.js
@@ -1,4 +1,5 @@
 const puppeteerutils = require('./puppeteer-utils');
+const env_vars = require('../utils/env_vars');
 
 const setup = async () => {
   let {page, browser} = await puppeteerutils.login();
@@ -28,15 +29,21 @@ const takedown = async () => {
   }
 };
 
-
 const waitForPage = async (page) => {
   // Most reliable way we've found to be sure the server finishes restarting
   // This counts on the fact that docassemble makes a `test.yml` file
   // for every new project. This method doesn not care if the file errors.
   const tries = 20;
-  const test_url = `${puppeteerutils.BASE_INTERVIEW_URL}%3Atest.yml`;
+  
+  // So that we can try to load our interview, we need this file to exist and work
+  // DEVELOPER: Always make sure your test.yml doesn't create an error page
+  await page.goto(`${env_vars.BASE_URL}/playground?project=${env_vars.PROJECT_NAME}`, {waitUntil: 'domcontentloaded'});
+  return
+
+  // test.yml only exists if the developer went to the playground
+  const test_url = `${puppeteerutils.BASE_INTERVIEW_URL}test.yml`;
   console.log("interview url:", test_url);
-  console.log("This isn meant to make sure the interview will load. Sometimes it doesn't work. If this test errors because the interview wouldn't load, you can try re-running this test.");
+  console.warn("This isn meant to make sure the interview will load. Sometimes it doesn't work. If this test errors because the interview wouldn't load, you can try re-running this test.");
 
   await page.goto(test_url, {waitUntil: 'domcontentloaded'});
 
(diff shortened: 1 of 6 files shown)


lib/github_integration/install-branch.js
```diff
@@ -28,15 +29,21 @@ const takedown = async () => {
...
+  
+  // So that we can try to load our interview, we need this file to exist and work
+  // DEVELOPER: Always make sure your test.yml doesn't create an error page
+  await page.goto(`${env_vars.BASE_URL}/playground?project=${env_vars.PROJECT_NAME}`, {waitUntil: 'domcontentloaded'});
+  return
+
+  // test.yml only exists if the developer went to the playground
+  const test_url = `${env_vars.BASE_INTERVIEW_URL}test.yml`;
+  console.log("\ninterview url:", test_url);
+  console.warn("This isn meant to make sure the interview will load. Sometimes it doesn't work. If this test errors because the interview wouldn't load, you can try re-running this test.");
```
niharikasingh, line 46: Spelling error: "isn" -> "isn't"

lib/github_integration/install-branch.js
```diff
@@ -28,15 +29,21 @@ const takedown = async () => {
...
   // for every new project. This method doesn not care if the file errors.
   const tries = 20;
-  const test_url = `${puppeteerutils.BASE_INTERVIEW_URL}%3Atest.yml`;
-  console.log("interview url:", test_url);
-  console.log("This isn meant to make sure the interview will load. Sometimes it doesn't work. If this test errors because the interview wouldn't load, you can try re-running this test.");
+  
+  // So that we can try to load our interview, we need this file to exist and work
+  // DEVELOPER: Always make sure your test.yml doesn't create an error page
+  await page.goto(`${env_vars.BASE_URL}/playground?project=${env_vars.PROJECT_NAME}`, {waitUntil: 'domcontentloaded'});
+  return
```
niharikasingh, line 41: Why are we returning here? Doesn't that mean the subsequent lines will never get executed?

lib/utils/env_vars.js
```diff
@@ -0,0 +1,17 @@
...
+
+const env_vars = {
+  PLAYGROUND_ID: process.env.PLAYGROUND_ID,
+  PLAYGROUND_EMAIL: process.env.PLAYGROUND_EMAIL,
+  PLAYGROUND_PASSWORD: process.env.PLAYGROUND_PASSWORD,
+  BASE_URL: process.env.BASE_URL,
+  BRANCH_NAME: process.env.BRANCH_PATH && process.env.BRANCH_PATH.split('/')[1] || 'test',
+  REPO_URL: process.env.REPO_URL,
+  DEBUG: process.env.DEBUG,
+};
```
niharikasingh, line 11: These `env_vars` just mirror `process.env` variables, so I'm not sure I understand why we're using the former instead of the latter. It would be helpful to have a comment somewhere in here about why this is preferable to using the `process.env` versions.

lib/steps.js
```diff
@@ -78,8 +79,9 @@ Given(/I start the interview at "([^"]+)"[ on ]?(.*)/, async (INTERVIEW_URL, opt
 
   scope.activate = click_with[ scope.device ];
 
+  let interview_url = `${ base_url }${ file_name }.yml`;
```
niharikasingh, line 82: Should this be a `const`?

niharikasingh (approved):

lib/github_integration/install-branch.js
plocket: I have no idea how this worked and now I'm a little worried...

lib/steps.js
plocket: The use of `const` vs. `let` is a whole philosophical topic of its own, but it's legitimate in both ways as far as javascript is concerned. Shall we discuss it separately from this PR?

lib/utils/env_vars.js
plocket: Does the new comment make sense?

PR #4, opened 2020-10-22
Abstract getting choice field, add notes on DOM (conflicts for sure)

plocket:
Will conflict with other branch at least because of version number, but we can work that out after those are merged.

Testing: I haven't yet figured out the process for testing this. I'm pairing it with https://github.com/SuffolkLITLab/docassemble-IntegratedTesting to test functionality, but the branches that match each others' functionality are not clear, so it's hard to see which to test with what. I guess that's a clue as to how to line up testing. Hasn't happened here, though.

I think the matching branch for this at https://github.com/SuffolkLITLab/docassemble-IntegratedTesting is `npmjs`.

Also, these tests probably won't work anyway because they try to use the `test.yml` file and that's not set up in this version of the package.

### DIFF
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -1,3 +1,13 @@
+const { expect } = require('chai');
+
+
+let ordinal_to_integer = {
+  only: 0, first: 0, second: 1, third: 2, fourth: 3, fifth: 4,
+  sixth: 5, seventh: 6, eighth: 7, ninth: 8, tenth: 9,
+  '1st': 0, '2nd': 1, '3rd': 2, '4th': 3, '5th': 4,
+  '6th': 5, '7th': 6, '8th': 7, '9th': 8,'10th': 9,
+};
+
 module.exports = {
   getTextFieldId: async function getTextFieldId(scope, field_label) {
     // field label is optional
@@ -20,7 +30,7 @@ module.exports = {
     }
 
     return field_id;
-  },
+  },  // Ends getTextFieldId
 
   waitForShowIf: async function waitForShowIf(scope) {
     // If something might be hidden or shown, wait extra time to let it hide or show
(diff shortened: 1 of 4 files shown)


lib/scope.js
```diff
@@ -52,5 +62,64 @@ module.exports = {
...
+    * @param {array} role_types 'radio', 'checkbox', or both. The type of field that the developer could be asking for.
+    * 
+    * @returns instance of ElementHandle or test fails
+    * 
+    * Methodologies proposed:
+    *     * Proposed: First dig all the way down to the label, then dig back out
+    *     * Regected: It's not clear how to keep things in order and filtering them
+    *     that way is more confusing, not less. It also takes more lines of code.
+    *     * Proposed: First collect all elements of expected status.
+    *     * Rejected: That ruins getting the correct ordinal
```
niharikasingh, line 82: This doesn't seem to list the methodology you ended up using here. I think it would be awesome to have a summary of how this function works, because it is so complicated.

lib/scope.js
```diff
@@ -52,5 +62,64 @@ module.exports = {
...
+      let [group_label] = await scope.page.$x( identifier_xpath );
+      expect( group_label.constructor.name, `Did not find a "${ group_label_text }" group on this page. ${ group_label.constructor.name }` ).to.equal('ElementHandle');
+
+      group_name = await group_label.evaluate(( elem ) => elem.getAttribute('for'));
+    }
+
+    let role_type = '';
+    for (let role of roles) {
+      role_type += `@role="${ role }"`;
+      if ( roles.indexOf(role) < roles.length - 1 ) { role_type += ' or '; }
```
niharikasingh, line 100: This might be a personal preference, but I think this line is confusing. I think this might be one of the rare cases where a `for (var i = 0; i < roles.length; i++)` loop is clearer, because then this comparison becomes much simpler.

lib/scope.js
```diff
@@ -52,5 +62,64 @@ module.exports = {
...
-  },
+  },  // Ends afterStep
+
+  getSpecifiedChoice: async function getSpecifiedChoice(scope, {ordinal, label_text, group_label_text, roles}) {
+    /* Returns the elementHandle of the field specified by the arguments
+    * 
+    * @param {string} [ordinal='first'] Which instance of the field on the page.
+    * @param {string} [label_text=''] Text of the label for the field, if it has one.
+    * @param {string} [group_label_text=''] Text of the label for the group of fields to which the field belongs, if there is one.
+    * @param {array} role_types 'radio', 'checkbox', or both. The type of field that the developer could be asking for.
```
niharikasingh, line 73: It would be nice to be explicit about which params are optional.

lib/scope.js
```diff
@@ -52,5 +62,64 @@ module.exports = {
...
+    }
+
+    let role_type = '';
+    for (let role of roles) {
+      role_type += `@role="${ role }"`;
+      if ( roles.indexOf(role) < roles.length - 1 ) { role_type += ' or '; }
+    }
+
+    // labels for choices
+    let label_xpath = `//*[@id="daform"]//label[${ role_type }]`;
```
niharikasingh, line 104: I'm not an XPATH expert, so I wasn't able to double-check this XPATH. 😞

niharikasingh (approved):

lib/scope.js
plocket: From what I can tell, it works...

lib/scope.js
plocket: I think the square brackets are the syntax for that, but we can also write it in.

PR #8, opened 2020-11-13
Add/edit steps

plocket:
I think the sibling repo has a `steps_dev` branch. I could be wrong.


niharikasingh (approved):

plocket:
Will have to resolve these conflicts locally I think.

plocket:
Made issue for what we want to keep from here (2 steps) #274

PR #18, opened 2020-11-16
Throw error on setup failure, add new branch name management

plocket:
Working for https://github.com/GBLS/docassemble-MAEvictionDefense currently

### DIFF
--- a/lib/github_integration/install-branch.js
+++ b/lib/github_integration/install-branch.js
@@ -10,7 +10,8 @@ const setup = async () => {
     await waitForPage(page);  // Ensure server has really restarted
   }
   catch (e) {
-    console.log(e);
+    await page.screenshot({ path: `error-setingUp.jpg`, type: 'jpeg', fullPage: true });
+    throw e;
   }
   finally {
     browser.close();
--- a/lib/github_integration/install-branch.js
+++ b/lib/github_integration/install-branch.js
@@ -10,7 +10,7 @@ const setup = async () => {
     await waitForPage(page);  // Ensure server has really restarted
   }
   catch (e) {
-    await page.screenshot({ path: `error-setingUp.jpg`, type: 'jpeg', fullPage: true });
+    await page.screenshot({ path: `error-settingUp.jpg`, type: 'jpeg', fullPage: true });
     throw e;
   }
   finally {
--- a/lib/utils/env_vars.js
+++ b/lib/utils/env_vars.js
@@ -19,9 +19,12 @@ const env_vars = {
 
 // Path could be 'refs/branch_name' or 'refs/head/branch_name' and maybe other things
 // Assume branch name is last part of branch path
-let split = process.env.BRANCH_PATH.split('/');
-if (process.env.BRANCH_PATH) { env_vars.BRANCH_NAME = split[ split.length - 1 ]; }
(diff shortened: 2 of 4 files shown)


lib/utils/env_vars.js
```diff
@@ -13,11 +13,16 @@ const env_vars = {
...
   PLAYGROUND_PASSWORD: process.env.PLAYGROUND_PASSWORD,
   BASE_URL: process.env.BASE_URL,
-  BRANCH_NAME: process.env.BRANCH_PATH && process.env.BRANCH_PATH.split('/')[1] || 'test',
   REPO_URL: process.env.REPO_URL,
   DEBUG: process.env.DEBUG,
 };
 
+// Path could be 'refs/branch_name' or 'refs/head/branch_name' and maybe other things
+// Assume branch name is last part of branch path
+let split = process.env.BRANCH_PATH.split('/');
```
niharikasingh, line 22: Maybe this should be in the if block following. In my Chrome console, the following lead to errors: 

```javascript
x = null 
x.split('/')
// ERROR 
x = undefined 
x.split('/')
// ERROR 
```

niharikasingh (changes requested):

lib/github_integration/install-branch.js
```diff
@@ -10,7 +10,8 @@ const setup = async () => {
     await waitForPage(page);  // Ensure server has really restarted
   }
   catch (e) {
-    console.log(e);
+    await page.screenshot({ path: `error-setingUp.jpg`, type: 'jpeg', fullPage: true });
```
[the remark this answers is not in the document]
plocket: Good catch 👍

lib/utils/env_vars.js
plocket: Makes sense. I was thinking we'd want it to error if the branch path was messed up, but we're providing a fallback already. There are probably pros and cons about _that_ to talk about too, but I'm not sure how crucial that conversation is, so we might as well stick with what we're doing now.

plocket: I suppose we then don't need the check on line 23 to check for the existence of the item on the list as even a single item will be valid.

lib/utils/env_vars.js
```diff
@@ -13,11 +13,19 @@ const env_vars = {
...
   PLAYGROUND_PASSWORD: process.env.PLAYGROUND_PASSWORD,
   BASE_URL: process.env.BASE_URL,
-  BRANCH_NAME: process.env.BRANCH_PATH && process.env.BRANCH_PATH.split('/')[1] || 'test',
   REPO_URL: process.env.REPO_URL,
   DEBUG: process.env.DEBUG,
 };
 
+// Path could be 'refs/branch_name' or 'refs/head/branch_name' and maybe other things
+// Assume branch name is last part of branch path
+if ( BRANCH_PATH ) {
```
niharikasingh, line 22: I think this should be `process.env.BRANCH_PATH`, since the variable by itself doesn't exist yet.

niharikasingh (changes requested):

lib/utils/env_vars.js
plocket: Gorramit. You're right. Good eye!

-

niharikasingh (approved):

PR #23, opened 2020-11-19
Download a file

plocket:
Merge after previous PR (#18) as this uses it.

### DIFF
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -124,7 +124,7 @@ module.exports = {
 
   detectDownloadComplete: async function detectDownloadComplete(scope, endTime) {
     /* Resolve if a download flag has changed. Timeout after a bit less than
-    *    the gobal permitted timeout */
+    *    the global permitted timeout */
 
     // Stop/resolve if too much time has passed
     let expirationTime = 1.9 * 60 * 1000;  // m * s * ms
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -121,4 +121,28 @@ module.exports = {
 
     return elem;
   },  // Ends getSpecifiedChoice
+
+  detectDownloadComplete: async function detectDownloadComplete(scope, endTime) {
+    /* Resolve if a download flag has changed. Timeout after a bit less than
+    *    the gobal permitted timeout */
+
+    // Stop/resolve if too much time has passed
+    let expirationTime = 1.9 * 60 * 1000;  // m * s * ms
+    if ( !endTime ) { endTime = new Date((new Date()).getTime() + expirationTime); }
+    let diff = endTime.getTime() - Date.now();
+    if ( diff <= 0 ) { return false; }
+
+    // If time hasn't expired
+    if ( scope.downloadComplete ) {
+      // If download is complete, reset the flag and resolve
+      scope.downloadComplete = false;
+      return true;
+    } else {
+      // Otherwise wait and then check again
+      await scope.page.waitFor( 100 );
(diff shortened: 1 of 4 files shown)


plocket:
Added downloads to .gitignore in files to push

lib/steps.js
```diff
@@ -65,6 +73,40 @@ Given(/I start the interview at "([^"]+)"[ on ]?(.*)/, async (file_name, optiona
...
     scope.page.setDefaultTimeout(120 * 1000)
+    // Files should get downloaded
+    scope.docs_dir = `downloads_${ scope.scenario_name }`
+    await scope.page._client.send('Page.setDownloadBehavior', {
+      behavior: 'allow',
+      downloadPath: path.resolve( scope.docs_dir ),  // Save in root of this project
+    });
+
+    // ---- EVENT LISTENERS ----
+    // From https://stackoverflow.com/a/51423688
```
niharikasingh, line 84: This block seems like an incantation that I don't entirely understand, but if you've tested and it works, then I'm ready to believe in this spell 🧙

lib/scope.js
```diff
@@ -121,4 +121,28 @@ module.exports = {
 
     return elem;
   },  // Ends getSpecifiedChoice
+
+  detectDownloadComplete: async function detectDownloadComplete(scope, endTime) {
+    /* Resolve if a download flag has changed. Timeout after a bit less than
+    *    the gobal permitted timeout */
+
+    // Stop/resolve if too much time has passed
+    let expirationTime = 1.9 * 60 * 1000;  // m * s * ms
```
niharikasingh, line 130: Not urgent, but I wonder if we should be using a global variable for the permitted timeout, and then this should just be `env_vars.TIMEOUT - 2000` or something.

niharikasingh (approved):

lib/scope.js
```diff
@@ -121,4 +121,28 @@ module.exports = {
 
     return elem;
   },  // Ends getSpecifiedChoice
+
+  detectDownloadComplete: async function detectDownloadComplete(scope, endTime) {
+    /* Resolve if a download flag has changed. Timeout after a bit less than
+    *    the gobal permitted timeout */
```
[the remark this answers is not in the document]
plocket: gobal is such a great word, though!

lib/scope.js
plocket: I really like that. I'll make it an issue.

lib/steps.js
```diff
@@ -274,26 +316,47 @@ Then(/the "([^"]+)" link opens a working page/, async (linkText) => {  // √
...
 
   // Making this a race would probably be more simple
   let elem;
   if (elemType === 'button') {
-    [elem] = await scope.page.$x(`//button/span[contains(text(), "${phrase}")]`);
-    if ( !elem ) {
-      [elem] = await scope.page.$x(`//button[contains(text(), "${phrase}")]`);
+    // [elem] = await scope.page.$x(`//button/span[contains(text(), "${phrase}")]`);
+    // if ( !elem ) {
+      [elem] = await scope.page.$x(`//button//text()[contains(normalize-space(), "${phrase}")]/..`);
```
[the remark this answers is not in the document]
plocket: I think I take care of this in another PR, so I'm not going to mess with it right now in github.

lib/steps.js
plocket: The sorcery of stackoverflow...

PR #33, opened 2020-11-22
Develop language agnostic steps. Close #21.

### LINKED ISSUE #21: Allow using variable name to identify a field
Apparently the field `name` attribute value is the base64 encoded version of the variable name. That can be useful for people handling translations through separate .yml files as it can be language agnostic.

plocket:
Steps that just use var names and choice values. To be honest, this is probably the only type of functionality we really want. I regret everything.

This aligns with https://github.com/plocket/docassemble-MAEvictionDefense/tree/use_var_names

We'll see what the merge conflicts are.

Close #21

### DIFF
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -188,4 +188,27 @@ module.exports = {
     } else { return `cannot be base64ed`; }  // Make sure it's not usable
   },
 
+  detectDownloadComplete: async function detectDownloadComplete(scope, endTime) {
+    /* Resolve if a download flag has changed. Timeout after a bit less than
+    *    the global permitted timeout */
+
+    // Stop/resolve if too much time has passed
+    let expirationTime = 1.9 * 60 * 1000;  // m * s * ms
+    if ( !endTime ) { endTime = new Date((new Date()).getTime() + expirationTime); }
+    let diff = endTime.getTime() - Date.now();
+    if ( diff <= 0 ) { return false; }
+
+    // If time hasn't expired
+    if ( scope.downloadComplete ) {
+      // If download is complete, reset the flag and resolve
+      scope.downloadComplete = false;
+      return true;
+    } else {
+      // Otherwise wait and then check again
+      await scope.page.waitFor( 100 );
+      await detectDownloadComplete(scope, endTime);
+    }
+
+    return false;
+  },
 };
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -120,5 +120,37 @@ module.exports = {
     expect( elem.constructor.name, `Did not find anything where the ${ ordinal } "${ label_text }" ${ roles } should have been. ${ elem.constructor.name }` ).to.equal('ElementHandle');
 
     return elem;
-  },  // Ends getSpecifiedChoice
+  },  // Ends getSpecifiedChoice()
+
+  tapElement: async function tapElement(scope, elem) {
+    /* Tap a given element, detect navigation, and do appropriate related things. */
+    expect( elem, 'That element cannot be found, so it cannot be tapped.' ).to.exist;
+
+    let start_url = await scope.page.url()
+
+    let winner;
+    if (elem) {
+      winner = await Promise.race([
+        Promise.all([  // Error loads page, so no need to detect
+          // Click with no navigation will end immediately
+          elem[ scope.activate ](),
+          scope.page.waitForNavigation({waitUntil: 'domcontentloaded'}),
+        ]),
+        scope.page.waitForSelector('.alert-danger'),
+        scope.page.waitForSelector('.da-has-error'),
+      ]);
+    }
+
+    // Not sure how to detect navigation landing at same page.
+    let end_url = await scope.page.url();
+    let navigated = start_url !== end_url;
+    // Might be able to handle this in `scope.afterStep`
+    if ( navigated ) {  // if (winner.constructor.name === 'ElementHandle') {
+      // If navigation, wait for 200ms to let things settle down
+      await scope.afterStep(scope, {waitForTimeout: 200, navigated: navigated});
+    } else {
+      // If stay on page, wait for `show if` or feedback elements to be move around
+      await scope.afterStep(scope, {waitForShowIf: true, navigated: navigated});
+    }
+  },  // Ends tapElement()
 };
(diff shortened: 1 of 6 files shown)


plocket:
Have to double check something.

plocket:
Ok, we're good-ish. I forgot `tapElement` became a thing. It was when I was developing individual field-based variable name steps and wanted to abstract the button tapping. Not sure it still needs to be abstracted, but I'd rather leave it for now if that's ok.

Just noticed that during the merge conflict resolution I brought back some old unnecessary code, though, so I'll try to get rid of that.

plocket:
That should be good.

plocket:
I don't think I've gotten to fully develop the showifs. I haven't seen them all yet. They may need to be developed a bit more deliberately or as we go.

plocket:
@berit , @niharikasingh - ready for review

lib/scope.js
```diff
@@ -120,7 +120,73 @@ module.exports = {
...
+  },  // Ends getSpecifiedChoice()
+
+  tapElement: async function tapElement(scope, elem) {
+    /* Tap a given element, detect navigation, and do appropriate related things. */
+    expect( elem, 'That element cannot be found, so it cannot be tapped.' ).to.exist;
+
+    let start_url = await scope.page.url()
+
+    let winner;
+    if (elem) {
```
niharikasingh, line 132: this should always exist because of the assertion in line 127

lib/scope.js
```diff
@@ -120,7 +120,73 @@ module.exports = {
...
+      winner = await Promise.race([
+        Promise.all([  // Error loads page, so no need to detect
+          // Click with no navigation will end immediately
+          elem[ scope.activate ](),
+          scope.page.waitForNavigation({waitUntil: 'domcontentloaded'}),
+        ]),
+        scope.page.waitForSelector('.alert-danger'),
+        scope.page.waitForSelector('.da-has-error'),
```
niharikasingh, line 140: comment - this is not meant to detect breaking errors, just user-visible errors for user to fix

lib/scope.js
```diff
@@ -120,7 +120,73 @@ module.exports = {
...
+    *    attribute is made of a combination of its variable's name
+    *    and it's choice value's name, plus a little more.
+    * 
+    * @param {object} scope - The global namespace object
+    * @param {string} var_name - Variable name of the variable being set
+    * @param {string} choice_name - Value that will be set to true for the variable
+    * 
+    * Examples:
+    * checkboxNameAttribute( scope, 'eviction_all_reasons', 'nonpayment' );
+    * // "ZXZpY3Rpb25fYWxsX3JlYXNvbnNbQidibTl1Y0dGNWJXVnVkQSdd"
```
niharikasingh, line 169: fix the return value - it should be an object of 2 strings

lib/scope.js
```diff
@@ -120,7 +120,73 @@ module.exports = {
...
+    *    base64: https://github.com/nodejs/node/issues/6107#issuecomment-207177131
+    * 
+    * Examples:
+    * base64( scope, 'nonpayment' );
+    * // "bm9ucGF5bWVudA"
+    */
+    if ( typeof( text ) === `string`  ) {  // Prevent an error
+      // `Buffer` is a nodejs thing
+      return Buffer(text, `utf8`).toString(`base64`).replace(/[^A-Za-z0-9]/g, '');
+    } else { return `cannot be base64ed`; }  // Make sure it's not usable
```
niharikasingh, line 188: change string to something like "will never match existing dom node"

lib/steps.js
```diff
@@ -353,8 +356,7 @@ When(/I download "([^"]+)"$/, async (filename) => {
 
 When(/I tap the (button|link) "([^"]+)"/, async (elemType, phrase) => {  // √
   /* Taps a button and stores or reacts to what happens:
-  *    navigation, validation error, page error, document download, or just a click.
-  * WARNING: Can only handle one download per scenario currently. */
```
niharikasingh, line 357: add back in?

niharikasingh: add to download function, actually

lib/steps.js
```diff
@@ -373,35 +375,122 @@ When(/I tap the (button|link) "([^"]+)"/, async (elemType, phrase) => {  // √
...
 //#####################################
 
+// -------------------------------------
+// --- VARIABLES-BASED STEPS ---
+
+When(/I set the var(?:iable)? "([^"]+)" to "([^"]+)"/, async (var_name, value) => {
+  /* Use variable names and values to interact with:
+  *  - Buttons that have values
+  *  - Input elements that take text as input
+  *  - Textareas
```
niharikasingh, line 392: needs to talk about select, but another MR might do that already.

lib/steps.js
```diff
@@ -373,35 +375,122 @@ When(/I tap the (button|link) "([^"]+)"/, async (elemType, phrase) => {  // √
...
+  // `yesno`/`noyes` checkbox: `name` = base64 var name, value always 'True'
+  // checkbox: `name` = base64 var name + stuff + base64 choice value + stuff (see function)
+  let to_tap = await Promise.race([
+    scope.page.waitForSelector(`input[name="${base64_var_name}"][value="${ value }"] ~ label`),  // radio
+    // scope.page.waitForSelector(`*[data-saveas="${base64_var_name}"] label`),  // radio showif?
+    scope.page.waitForSelector(`input[name="${base64_var_name}"][value="True"] ~ label`),  // yesno/noyes checkbox
+    // scope.page.waitForSelector(`*[data-saveas="${base64_var_name}"] label`),  // yesno/noyes checkbox showif?
+    scope.page.waitForXPath(`//input[@id="${base64_first_id}"]/ancestor::fieldset//*[contains(@class,"danota-checkbox")]/following-sibling::label`),  // 'none of the above' checkbox
+    // 'none of the above' checkbox showif???
+    scope.page.waitForSelector(`input[name="${name_attr_B}"] ~ label`),  // checkbox
```
niharikasingh, line 480: pointer to docassemble code would be nice, so people know the B/R distinction exists.

niharikasingh (approved):

lib/steps.js
plocket: Confirmed that it does when this is abstracted into `scope` in another PR.

lib/scope.js
plocket: This function is no longer being used anywhere and should be removed! Yay!

PR #35, opened 2020-11-23
Improves feedback for question id step. See MADE matching branch. Fix #17

### LINKED ISSUE #17: When a screen id is not found, add an error message saying what id was found instead (and speed it up)
We might be able to speed it up with XPath or something.

plocket:
You'll have to uncomment two lines, each individually, to see the new error messages in action if you so choose.

Fixes #17

### DIFF
--- a/lib/steps.js
+++ b/lib/steps.js
@@ -147,17 +147,19 @@ When(/I wait (\d*\.?\d+) seconds?/, async (seconds) => {  // √
 /* Here to make writing tests more comfortable. */
 When("I do nothing", async () => { await scope.afterStep(scope); });  // √
 
-// TODO: Switch to getting the id and then checking it against the argument
-//     which will make for more useful test error messages.
 Then('the question id should be {string}', async (question_id) => {  // √
   /* Looks for a santized version of the question id as it's written
-  *     in the .yml. docassemble:
+  *     in the .yml. docassemble's way in python:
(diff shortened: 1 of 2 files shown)


lib/steps.js
```diff
@@ -147,17 +147,19 @@ When(/I wait (\d*\.?\d+) seconds?/, async (seconds) => {  // √
...
-  // Add error message saying what the screen id was
+  let id_as_class = question_id.toLowerCase().replace(/[^A-Za-z0-9]+/g, '-');
+  // `.className` is the node's list of classes as a string, separated by spaces
+  let classes_str = await scope.page.$eval(`body`, function (elem) { return elem.className });
+  let ids_found = classes_str.match(/question-([^ ]+)/);
+
+  // Trying to make helpful error messages
+  let found_any_question_id = ids_found !== null;
+  expect( found_any_question_id , '~ Did not find any question id ~' ).to.be.true;
+  expect( ids_found[1], `~ This page's actual question id is "${ ids_found[1] }" ~` ).to.equal( id_as_class );
```
niharikasingh, line 162: Why are we looking at `ids_found[1]` and not `ids_found[0]`?

niharikasingh (approved):

lib/steps.js
plocket: That's how `.match()` works - the first item on the list is the whole string that's being searched. After that, the items are matches.

Issue #61, opened 2020-12-02
Try "${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY}" for default of current repository


plocket:
REPO_URL="${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY}"

That or make it a secret. Hard to develop on a fork otherwise.

plocket:
Closed by some recent merge...

Issue #53, opened 2020-11-29
Separate some of the interview start options into individual steps


plocket:
I overloaded the 'Given' step a bit. We can possibly break out:
1. Choosing a device: "I start on mobile" or something.
1. Possibly break out choosing a language into its own step too, though it might complicate the automated language selection step, so may not be worth it.

Thoughts?

plocket:
Closed by #67 for now

PR #67, opened 2020-12-07
Make artifact filenames more specific, move device into own step

plocket:
Also uses dynamic `REPO_URL` because of issue MADE was running into.

You can see the behavior of this version and the associated artifacts for (intentional) errors, screenshots, and downloads with their file names at [this action](https://github.com/plocket/docassemble-MAEvictionDefense/actions/runs/405916344) which should be done by 8:40am ET.

### DIFF
--- a/lib/steps.js
+++ b/lib/steps.js
@@ -63,12 +63,6 @@ Before(async (scenario) => {
 
 // Add a check for an error page before each step? After each step?
 
-// I start the interview on mobile
-// I start the interview on pc
-// I start the interview
-// I go to the file "juvenile-sealing" on a mobile
-// I go to "lskdfjlasd.com" on pc
-// I go to the interview at "lksdfjls.com" on pc
 Given(/I start the interview at "([^"]+)"(?: in lang "([^"]+)")?/, async (file_name, language) => {  // √
 
   await scope.start( scope, file_name );
--- a/lib/steps.js
+++ b/lib/steps.js
@@ -56,7 +56,12 @@ Before(async (scenario) => {
   scope.scenario_name = scenario.pickle.name.replace(/[^A-Za-z0-9]/gi, '');
   scope.downloadComplete = false;
   scope.toDownload = '';
+  scope.device = 'pc';
+  scope.activate = click_with[ scope.device ];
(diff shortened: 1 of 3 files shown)


plocket:
Addresses #53, may close #61.

lib/steps.js
```diff
@@ -56,21 +56,29 @@ Before(async (scenario) => {
...
   scope.toDownload = '';
+  scope.device = 'pc';
+  scope.activate = click_with[ scope.device ];
+  scope.filename_suffix = '';
 });
 
+// Add a check for an error page before each step? After each step?
+
 // I start the interview on mobile
 // I start the interview on pc
```
niharikasingh, line 67: Should remove these examples now that device is its own step.

pushing_testing_files/run_form_tests.yml
```diff
@@ -36,8 +36,8 @@ jobs:
       - name: Set languages
         run: echo "EXTRA_LANGUAGES=${{ github.event.inputs.extra_languages }}" >> $GITHUB_ENV
         if: ${{ github.event.inputs.extra_languages != '' }}
-      - name: List languages
-        run: echo "Extra languages to test are $EXTRA_LANGUAGES"
+      - run: echo "REPO_URL=$GITHUB_SERVER_URL/$GITHUB_REPOSITORY" >> $GITHUB_ENV
+      - run: echo "Repo address is $REPO_URL"
```
niharikasingh, line 40: Not sure why "languages" code is replaced with "repo_url" code. Both of those serve different purposes, so shouldn't they both be here?

niharikasingh (approved):

lib/steps.js
plocket: Good eye!

pushing_testing_files/run_form_tests.yml
plocket: The diff is a bit misleading. Previously there was a step (?) to set the languages, then a `run` to list the languages out. I removed the second (since I noticed they get listed usefully in the log anyway) and added a `run` to set the `REPO_URL`.

PR #69, opened 2020-12-07
Make sure new tests start an interview fresh

plocket:
I made a branch on MADE that is deliberately likely to trigger timeout errors. In the past, some of the post-error test would have also errored because the interview would simply resume from where the last test had got it to. None of that should be happening now - just the normal timeout errors. It should be [this action](https://github.com/plocket/docassemble-MAEvictionDefense/actions/runs/406113860) that will probably be done around... 11? Feel free to run it again if you think it's warranted as this is based on race conditions.

### DIFF
--- a/lib/steps.js
+++ b/lib/steps.js
@@ -711,8 +711,9 @@ After(async (scenario) => {
     const name = scenario.pickle.name.replace(/[^A-Za-z0-9]/gi, '');
     await scope.page.screenshot({ path: `error-${name}.jpg`, type: 'jpeg', fullPage: true });
   }
-  // If there is a browser window open, then close it
-  if (scope.page) {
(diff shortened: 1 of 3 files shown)


niharikasingh (approved):

plocket:
I went back to look at the errors and they're not the right errors. Sorry I wasn't more specific.

The errors I saw were because it had the wrong repo name because otherwise the upstream repo tests fail. Trying to make a fix for that now, but it's hanging for over an hour on setup. It might be a long timeout or an infinite loop. I assume the former because the latter makes no sense at all.

To be more specific about what we're looking for - errors in the middle of the interview, not at the start.

plocket:
Ok, now it works

PR #71, opened 2020-12-08
Fix json scripts

plocket:
Fix some scripts and vars

### DIFF
--- a/pushing_testing_files/package.json
+++ b/pushing_testing_files/package.json
@@ -7,10 +7,11 @@
     "test": "tests"
   },
   "scripts": {
-    "langs": "npm run build_langs && npm run cucumber",
-    "build_langs": "node_modules/.bin/da-langs 'tests/features/*.feature'",
+    "test": "npm run langs_setup && npm run cucumber",
+    "langs_setup": "node_modules/.bin/da-langs 'tests/features/*.feature'",
     "cucumber": "./node_modules/.bin/cucumber-js --require ./node_modules/docassemble-cucumber/lib/index.js tests/features/*.feature",
-    "local_langs": "npm run build_langs && npm run local",
+    "langs": "npm run langs_setup && ./node_modules/.bin/cucumber-js --require ./node_modules/docassemble-cucumber/lib/index.js tests/features/*__lang*.feature",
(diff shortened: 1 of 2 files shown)


pushing_testing_files/package.json
```diff
@@ -7,10 +7,11 @@
     "test": "tests"
   },
   "scripts": {
-    "langs": "npm run build_langs && npm run cucumber",
-    "build_langs": "node_modules/.bin/da-langs 'tests/features/*.feature'",
+    "test": "npm run langs_setup && npm run cucumber",
     "cucumber": "./node_modules/.bin/cucumber-js --require ./node_modules/docassemble-cucumber/lib/index.js tests/features/*.feature",
-    "local_langs": "npm run build_langs && npm run local",
+    "langs": "npm run langs_setup && ./node_modules/.bin/cucumber-js --require ./node_modules/docassemble-cucumber/lib/index.js tests/features/*__lang*.feature",
```
niharikasingh, line 12: should this be something like `langs_test` to mirror `test`?

niharikasingh: alternatively, it would be nice to have a description of when one would use each of these commands. not a blocker for this MR, but a nice-to-have piece of documentation.

niharikasingh (approved):

pushing_testing_files/package.json
plocket: I'm not sure how to add a description as I've never done that in JSON (which doesn't have comments). `test` is for github, which runs the workflow, which uses `setup` and `takedown` in their own steps. `local` runs `setup` and `takedown` right in the script itself. The `local_langs`/`langs_local` script uses `local` inside it.

I'd appreciate your thoughts on it, though. I honestly find myself only using `npm run local` very rarely, though it is nice to have when I do use it.

plocket: I'll pull for now and maybe we can talk about this choice later.

niharikasingh: I was thinking in the readme file, we could have a description of all the default commands the package has. I'm not wedded to that idea though, so if you have a better location for this documentation, lmk.

PR #76, opened 2020-12-17
Allow setting of any variable, close #66

### LINKED ISSUE #66: Steps: Make `I set the...` for checkboxes?
Situation:
1. Developer makes a default for a checkbox that makes it 'checked' when the user first comes to the page.
1. Translator accidentally translate the words that set the default, leaving the checkbox without a default, meaning it starts as 'unchecked'. Let's call the language 'Foonian'
1. Test comes along. To test default language behavior, it does not tap the checkbox, leaving it **checked**.
1. Automated language tests come along and build the 'Foonian' test. That test also leaves the checkbox alone, which means the checkbox is **unchecked**.
1. The checkbox determines other behavior in the interview and so the tests in 'Foonian' appear to be broken, even though that default isn't crucial to the functioning of the interview.

It's going to be a bit of a pain to add that. I can't think of another way. It may mean that we really do only need one tests for setting any variable, so that might be a plus...

plocket:
See https://github.com/plocket/docassemble-MAEvictionDefense/actions/runs/428365746 at around 10:30 am.

Closes #66.

### DIFF
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -304,4 +304,102 @@ module.exports = {
     
     return scope.page;
   },  // Ends start()
+
+  getField: async function ( scope, { var_name, choice_name, set_to }) {
+
+    // All the possible variations of name attribute values
+    let base64_name = await scope.base64( scope, var_name );
+    let base64_id_of_first_choice = base64_name + '_0';
+    let { name_attr_B, name_attr_R } = await scope.checkboxNameAttribute( scope, var_name, choice_name );
+
+    // Note: I think `showif`s with `code` don't use `data-saveas`
+    // Note: Choices created without variable names cannot be language agnostic
+
+    // `$x` always gives a list, so just make sure all results are lists
+    // radio: `name` = base64 var name, `choice_name` = var name of choice
+    // `yesno`/`noyes` checkbox: `name` = base64 var name, choice always 'True'
+    // checkbox: `name` = base64 var name + stuff + base64 choice var name + stuff (see function)
+    let all_attempts = [
+      await scope.page.$$(`button[name="${ base64_name }"][value="${ choice_name }"]`),  // button
+      await scope.page.$$(`button[name="${ base64_name }"][value="${ set_to }"]`),  // button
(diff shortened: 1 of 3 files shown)


plocket:
The v0.5.0 might be a good shift from this. Maybe we want to wait for the work with the table/list, though.

lib/steps.js
```diff
@@ -396,109 +396,17 @@ When(/I tap the (button|link) "([^"]+)"/, async (elemType, phrase) => {  // √
 // -------------------------------------
 // --- VARIABLES-BASED STEPS ---
 
-When(/I set the var(?:iable)? "([^"]+)" to "([^"]+)"/, async (var_name, value) => {
-  /* Use variable names and values to interact with:
-  *  - Buttons that have values
-  *  - Input elements that take text as input
-  *  - Textareas
```
niharikasingh, line 403: I think we should still mention that this function covers interactions with buttons, text input, textareas, and radios.

lib/scope.js
```diff
@@ -304,4 +304,102 @@ module.exports = {
...
+
+      } else {
+        await scope.setText( scope, { handle, set_to });
+      }
+
+      await scope.afterStep(scope, { waitForShowIf: true });
+    },  // Ends scope.setField.INPUT()
+  },
+
+  setCheckbox: async function ( scope, { label, set_to }) {
```
niharikasingh, line 391: If we're only using `setCheckbox` in the `scope.js` file, what about starting the function name with an underscore (`_setCheckbox`) or in some other way indicating that this is a helper function that's not going to be used outside this file?

niharikasingh (approved):

lib/steps.js
plocket: Good idea. I think I may have moved that to scope.js along with the related code.

lib/scope.js
plocket: We can definitely start doing that. I think it might be worth making this its own task and go through and do that for all functions that are just used internally.

PR #84, opened 2020-12-21
Add some report capabilities to var_table

plocket:
Saves a primitive report `md` file to artifacts that lists the order in which question ids appeared during each scenario.

I'm trying to figure out what ways we can give useful feedback to the developer since most of their development is meant to be in Github and their ability to get test feedback would be a bit limited.

We might later figure out how to integrate some kind of flexible reporter into the mix, but this is my attempt at an MVP. Also, the data that is printed is saved as an object and I can see it being used in the future to make assertions about the order in which the questions appeared. That would let developers be more selective about what question order they care to test (as opposed to listing out every single step in sequential order).

See action: https://github.com/plocket/docassemble-MAEvictionDefense/actions/runs/434683706

Thoughts?

### DIFF
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -301,7 +301,7 @@ module.exports = {
 
     } catch ( err ) {
       // If it doesn't work, maybe try again
-      if ( attempt_num <= max_attempts ) { await scope.start( scope, file_name, attempt_num ); }
+      if ( attempt_num <= max_attempts ) { await scope.load( scope, file_name, attempt_num ); }
       else {
         // Otherwise throw an error, though there must be a better check we can do
         expect( false, `The interview at ${ interview_url } did not load after ${ max_attempts } tries.` ).to.be.true;
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -408,11 +408,14 @@ module.exports = {
     await handle.type( set_to );
   },
 
-  sign: async function ( scope, { handle }) {
-    /* Draw in a canvas element NOTE: Signature pages do not have the
-    * signature's variable name on the page. That seems to be intentional,
-    * though the reasoning hasn't become clear yet. */
-    let bounding_box = await handle.boundingBox();
+  sign: async function ( scope ) {
+    /* Draw in a canvas element. NOTE: Signature pages do not have the
+    * signature's variable name on the page. That seems to be intentional
+    * in da, though the reasoning hasn't become clear yet. */
+    let canvas = await scope.page.waitForSelector('#dasigcanvas');
+    expect( canvas, `No signature field could be found`).to.exist;
+
+    let bounding_box = await canvas.boundingBox();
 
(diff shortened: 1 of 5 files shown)


plocket:
Should this be saved as a `.md`? Are special programs required to open a `.md`? I didn't even think of that.

lib/scope.js
```diff
@@ -146,7 +146,10 @@ module.exports = {
     let end_url = await scope.page.url();
     let navigated = start_url !== end_url;
     // Might be able to handle this in `scope.afterStep`
-    if ( navigated ) {  // if (winner.constructor.name === 'ElementHandle') {
+    if ( navigated ) {
+      // Save the id that comes next
+      let { question_has_id, id, id_matches } = await scope.examinePageID( scope, 'none to match' );
+      scope.report[ scope.scenario_name ].ids.push( id );
```
niharikasingh, line 152: We're using this `scope.report[ scope.scenario_name ].ids.push( id )` syntax in several places, and I can see us adding it to even more functions in the future. I think this will be more flexible if we replace this with a function `scope.addToReport(msg)` where add to report handles `scope.report[ scope.scenario_name ].ids.push( id )` (or whatever other implementation we end up with. That makes sure we don't repeat the implementation details in several places, and the implementation is easy to change whenever we need to. 

Not a blocker for now, but it's a nice-to-have change.

niharikasingh:
> Should this be saved as a `.md`? Are special programs required to open a `.md`? I didn't even think of that.

Is there any reason to prefer `.md` over `.txt`? Any standard text editor can open both kinds of files, but most non-programmers will be more used to seeing `.txt` files.

lib/steps.js
```diff
@@ -731,7 +737,21 @@ After(async (scenario) => {
   }
 });
 
-AfterAll(async () => {
+AfterAll(async function() {
+  // Create report
+  let report = `# ALKiln REPORT - ${ (new Date()).toUTCString() }\n\n`
```
niharikasingh, line 742: What does `ALKiln` stand for? I'm blanking on that right now.

niharikasingh (approved):

lib/scope.js
plocket: I like that idea, though we may not want it to be simply for pushing ids as we may want to be able to report other things as well, so we may want to think about its interface a bit more. Made this one into an issue too.

lib/steps.js
plocket: lol, sorry, I don't think I mentioned this to anyone here - we're looking at reorganizing the repos that have to do with the Assembly Line project. We're planning to call this one the 'kiln', prefixed by 'AL' for Assembly Line. We can discuss that and change it if needed.

plocket:
I like `.md`s as they can look nicer and sometimes be turned into nice html pages later on if we want to do a fancier reporter in the future. I'm not sold on those reasons, though.

PR #85, opened 2020-12-21
Log report to the console as well

plocket:
Should we log the report to the console? It's the only way I could figure out how to do it that wouldn't require people to download a file. I personally find it really annoying to download from actions. Or in general. Thoughts?

See action: https://github.com/plocket/docassemble-MAEvictionDefense/actions/runs/434724343


niharikasingh:
I prefer console output over downloads for log output as well 👍

niharikasingh (approved):

PR #92, opened 2020-12-28
One-step for the var table with the id name

plocket:
Decided to embrace the id name as it really simplified stuff. (I think both for us and for the developer).

Test: ~https://github.com/plocket/docassemble-MAEvictionDefense/runs/1615705210?check_suite_focus=true~

https://github.com/plocket/docassemble-MAEvictionDefense/actions/runs/448107706

### DIFF
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -444,7 +444,7 @@ module.exports = {
     return { question_has_id: found_any_question_id, id: id, id_matches: ids_match };
   },  // Ends scope.examinePageID()
 
-  processVar: async function ( scope, { how_many, var_data, id, attempt=1 }) {
+  processVar: async function ( scope, { how_many, var_data, id }) {
     /* Non-recursive function to set as many variables as possible before
     * reaching the terminal page. Important to avoid recursion because of async ops. */
 
@@ -455,7 +455,6 @@ module.exports = {
     // Loop through setting all vars that can be set. Also collect all
     // remaining unused vars for later.
     for ( let row of var_data ) {
-
       // Try to find this varible on the current page
       let var_name = row.var;
       let choice_name = row.choice;
@@ -468,11 +467,8 @@ module.exports = {
       // of how we must detect the need to hit 'continue', all remaining vars
       // will be looped through and get 'found', removing them from the table.
       if ( set_to.includes('/sign') ) {
-        // Wait extra time for a signature. For some reason that's become an issue recently...?
-        await scope.page.waitFor( 10 );
         let is_signature_page = await scope.page.$('.dasignature');
         if ( is_signature_page ) {
-          process.stdout.write(`got signature page`);
           a_var_was_found = true;  // Even though it doesn't have a proper var name
 
           let handle = await scope.page.waitForSelector(`#dasigcanvas`);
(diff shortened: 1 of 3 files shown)


lib/scope.js
```diff
@@ -471,27 +472,34 @@ module.exports = {
...
-      }
+      if ( set_to.includes('/sign') ) {
+        let is_signature_page = await scope.page.$('.dasignature');
+        if ( is_signature_page ) {
+          a_var_was_found = true;  // Even though it doesn't have a proper var name
+          await scope.sign( scope );
+          process.stdout.write(`\x1b[36m${ '~' }\x1b[0m`);  // signed
+        } else {
+          // Remember any variable that wasn't found yet
+          remaining_vars.push( row );
```
niharikasingh, line 483: Just for my own understanding, does `remaining_vars` act as a queue? That is, it pops an element from the front of the queue and tries to finish the question related to that element. If it can't, it adds the element to the back of the queue.

lib/steps.js
```diff
@@ -64,6 +64,7 @@ Before(async (scenario) => {
   scope.toDownload = '';
   scope.device = 'pc';
   scope.activate = click_with[ scope.device ];
+  scope.target_question_id = 'null';
```
niharikasingh, line 67: Where do we use this variable? 

Also, it seems weird to use 'null' as the default string. I think it would be better practice to use `null` as a keyword or to use an empty string.

lib/steps.js
```diff
@@ -180,66 +178,42 @@ When(/^(some|all)(?: of)?(?: the(?:se)? var(?:iable)?s are used)?/, {timeout: -1
...
   }
 
-  let total_vars = var_data.length;  // For a check at the end
-
-  expect( scope.final_question_id, `You need to give the id of the question that should be reached when ${ how_many } of the variables in the table are set. Use the step 'When the final question id is "your question id"'` ).to.exist;
-  // See if this is the terminating case already!
-  let { question_has_id, id, id_matches } = await scope.examinePageID( scope, scope.final_question_id );
-
   // Until we reach the final page or hit an error
+  let original_num_vars = var_data.length;  // For a check at the end
```
niharikasingh, line 182: We never use this variable anymore

lib/steps.js
```diff
@@ -180,66 +178,42 @@ When(/^(some|all)(?: of)?(?: the(?:se)? var(?:iable)?s are used)?/, {timeout: -1
...
   }  // ends while !id_matches
-
-  let remaining_json = JSON.stringify(remaining_vars);
-  let target_id = scope.final_question_id
-
-  // if the id does not match we did not reach the desired screen
-  let target_not_reached_msg = `We needed to get to the question id "${ target_id }",` +
-    ` but all the variables that can be used have been used and the question id is only "${ id }".` +
-    ` The remaining variables are ${ remaining_json }`
-  expect( id_matches, target_not_reached_msg ).to.be.true;
```
niharikasingh, line 227: Do we not display this error message anymore? I don't see this error message in the code anymore.

lib/scope.js
plocket: No, it just keeps track of what variables haven't been used yet. Everything stays in the same order. Here's the idea with all the stuff in between removed:
```py
original_list_of_vars = [ 'var1', 'var2' ]
fresh_list_of_vars = []
for a_var in original_list_of_vars:
  if found_on_page( a_var ):
    set_var()
  else:
    fresh_list_of_vars.push( a_var )
```

Does that make sense?

lib/steps.js
plocket: 1. Good catch.
1. That was definitely not supposed to be a string
1. It should not be needed anymore. Something just got lost in translation.

lib/steps.js
plocket: Good catch

lib/steps.js
plocket: The target id being reached is the only way the loop will stop, so if the target can't be reached, something else will have to halt the loop, such as an error. I _think_ I've managed to throw an error for all the cases that would occur, but if you see something I've missed, let me know. Remember that when a variable can't be found on the page, the code tries to hit a 'continue' type button. I _think_ these are the situations that I tried to cover, though I may be missing something that I thought of and addressed in the code:
1. The continue button is pressed, but the user can't continue - they get a validation message or a system error.
1. The code tries to hit the continue button, but it doesn't exist.
1. The code tries to set a variable, but its element can't be interacted with (it's hidden or something).
1. Something hangs (the page takes too long to load or something).

niharikasingh (approved):

PR #93, opened 2020-12-28
Update some good stuff from main

plocket:


niharikasingh (approved):
This seems fine, and I don't think it needs a detailed review since all of these changes were reviewed before being merged into `main`.

Issue #54, opened 2020-11-29
Research using tags to filter which scenarios or features to run (including on manually run tests)


plocket:

plocket:
Would we call this done?

Issue #19, opened 2020-11-18
Add a manual trigger with the ability to set env vars


plocket:

plocket:
Discussing allowing the following inputs:
1. `language`: Pick just one language to run. How do we make sure that 'default' will run the default language without error? Can probably use it to set environment variables. Alternatively, an `if` in the workflow or in the test?
1. `filepath`: Pick one specific test file. Would probably affect the file path given as the argument to the script. Can that script use environment variables?
1. `url`: Specify a different url to test at (`INTERVIEW_URL` environment variable)

Other ideas?

plocket:
`extra_languages` might be a better name than `languages`

plocket:
* Manual trigger has been added.
* `extra_languages` is now available, though language-agnostic functionality is not quite ready.

plocket:
`extra_languages` has been added.

plocket:
Remaining:
- [ ] `url`
- [x] `tags`
- [ ] `fileglob` (?)

plocket:
At this point, we have decided against adding these other env vars.

- `tags`: Authors manage this now in their workflow file
- `languages`: Can be controlled with a [cucumber `Examples` block](https://cucumber.io/docs/gherkin/reference/#examples) in a `Scenario Outline` for the test. We'll see if it's possible to use a Background action so authors can avoid writing those examples in every test.
- `url` Can be given on a per-Scenario, or even per-Step, basis.
- `filepath`: Actually better taken care of by  `tags`

PR #96, opened 2020-12-29
add tags to generated language tests

berit:


plocket (approved):
Ran it, tooled around, looks good! We need to publish this as soon as we merge it.

plocket:
I also didn't realize `npm run cucumber_base -- "$@" tests/features/*.feature` could run tests that don't have tags. Very thoughtful of the developers.

plocket:
Rather, `npm run langs_setup && npm run cucumber -- $@`

plocket (commented):
Hmm, also, does this need to change the action file that it pushes to other repos? (looking at the PR on MADE)
[The workflow file and the package.json, I think]

plocket:
Will close #54, research tags

plocket:
Will also address #19

berit:
@plocket What does the comment about "isn't ignoring reports" mean? What reports?

plocket:
We have files that get saved now that are called reports. They have `_report_` in the name. They give the tester some extra information about what happened during the tests.

I thought that we added `_report_` to the script that gives a repository its .gitignore, but something must have gotten lost along the way.

plocket (approved):
I think this is working correctly. See it being used with `extra_languages=Español` and `tags=@fast`: https://github.com/plocket/docassemble-MAEvictionDefense/runs/1676964971?check_suite_focus=true

Issue #74, opened 2020-12-15
DECISION: (Story tables) List/State/'Scenario'-based questions


plocket:
[Decision: We've agreed that these would be useful, but do not replace tests that test the order of questions, as that is sometimes important too.]

Discussion regarding a test such as this (tables can have multiple columns):

```
  Given I set these vars:
    | some_var | some value or other |
    | other_var | 3 |
```
[Data table cucumber docs](https://cucumber.io/docs/cucumber/api/#step-arguments)

### Cons
1. no order of questions
1. more flexible/permissive
1. no human language cues
1. no intermediate/question-id-type checks (error should include id)
1. easier to read for humans

### Pros
1. no order of questions
1. more flexible/permissive
1. no human language needed
1. faster to write for coders perhaps
1. easier to read for coders
1. more behavior-testing-like than what we currently have

### Possible variations
- include a version that tests that _all_ those values come up and get set
- include a version that tests that _any_ of those values come up and get set

plocket:
The story table has been implemented with #107.

PR #107, opened 2020-12-30
Add 'stories' through table of variable values

plocket:
Also includes reports for the table stories.

Addresses decision in #74, though not sure if we should close it or what. I don't think we made other issues for this.

This is already published. Also, this is a merged version of `main` and `var_table` that's tested at https://github.com/plocket/docassemble-MAEvictionDefense/actions/runs/453094859.

I know there's a lot in here, but I hope we've gone over a lot of pieces already along the way.


plocket:
Published 0.4.67 along with that last commit.

niharikasingh (approved):

PR #112, opened 2021-01-10
Fix bug with field-finding error detection, improve err msg. (quick PR after #107)

plocket:
We'll deal with merge conflicts after the previous PR is merged so we can update to an appropriate version number. Maybe we should look into automating publishing on merge to `main` (https://github.com/plocket/docassemble-cucumber/issues/105).


niharikasingh (approved):
Nice and easy :)

PR #121, opened 2021-01-14
Add function to format the test reports for printing.

plocket:
This was it working a couple weeks ago: https://github.com/plocket/docassemble-MAEvictionDefense/actions/runs/452943821

Also:
1. Switches to .txt instead of .md
1. Adds a function for storing report data.

### DIFF
--- a/lib/steps.js
+++ b/lib/steps.js
@@ -58,7 +58,7 @@ BeforeAll(async () => {
 });
 
 Before(async (scenario) => {
-  // Names
+  // Names for files and reports
   let scenario_name = scenario.pickle.name.replace(/[^A-Za-z0-9]/gi, '');
   let tags = ''
   for ( let tag of scenario.pickle.tags ) {
@@ -66,12 +66,14 @@ Before(async (scenario) => {
   }
   // With tag names automatically added to langauges, this should work to make
   // each language unique.
-  scope.safe_id = `_${ scenario_name }${ tags }`
+  let new_id = `_${ scenario_name }${ tags }`
 
   // Should we add a counter? It would have to be unique to each scenario name.
-  if ( scope.report[ scope.safe_id ] ) {
-    scope.safe_id += `_1`;
+  // Not sure of another good way to indicate the _name_ is a duplicate
+  while ( scope.report[ new_id ] !== undefined  ) {
+    new_id += '_';
   }
+  scope.safe_id = new_id;
 
   // Downloads
   scope.downloadComplete = false;
--- a/lib/steps.js
+++ b/lib/steps.js
@@ -52,20 +52,36 @@ let click_with = {
 };
 
 BeforeAll(async () => {
+  scope.timeout = 120 * 1000;
+  scope.showif_timeout = 500;
   scope.report = {};
-
 });
 
 Before(async (scenario) => {
-  scope.timeout = 120 * 1000;
-  scope.showif_timeout = 500;
-  scope.scenario_name = scenario.pickle.name.replace(/[^A-Za-z0-9]/gi, '');
+  // Names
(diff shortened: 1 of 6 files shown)


lib/steps.js
```diff
@@ -52,20 +52,36 @@ let click_with = {
...
+  for ( let tag of scenario.pickle.tags ) {
+    tags += tag.name.replace(/[^A-Za-z0-9]+/g, '_').replace(/(\P{L})+/gu, '_');
+  }
+  // With tag names automatically added to langauges, this should work to make
+  // each language unique.
+  scope.safe_id = `_${ scenario_name }${ tags }`
+
+  // Should we add a counter? It would have to be unique to each scenario name.
+  if ( scope.report[ scope.safe_id ] ) {
+    scope.safe_id += `_1`;
```
niharikasingh, line 73: Hmm... I'm not sure this works in all cases. What if you have keys `abc` and `abc_1` in `scope.report` already, and the current `scope.safe_id = 'abc'`? I think this would end up creating a duplicate key in that case. 

Not sure if this is worth worrying about right now, since this could be a rare edge case that we can afford to ignore for now.

lib/steps.js
```diff
@@ -52,20 +52,36 @@ let click_with = {
...
+  }
+
+  // Report
+  scope.report[ scope.safe_id ] = { ids: [] };
```
niharikasingh, line 77: Why not let `scope.addToReport` handle this initialization? Doing it here seems to duplicate work.

lib/steps.js
```diff
@@ -80,15 +96,15 @@ Before(async (scenario) => {
 Given(/I start the interview at "([^"]+)"(?: in lang "([^"]+)")?/, async (file_name, language) => {  // √
   await scope.load( scope, file_name );
 
-  scope.filename_suffix = `_${scope.scenario_name}`;
-  if ( language ) { scope.filename_suffix = `${scope.filename_suffix}_${ language }`; }
+  // This won't be needed after tags get added in
```
niharikasingh, line 99: Are you referring to https://github.com/plocket/docassemble-cucumber/pull/96/files?

niharikasingh (approved):

lib/steps.js
plocket: You're totally right. I _knew_ there was something wrong with that. I didn't want to bother adding a count, though.

Bleh, if it's not going to work properly should we just remove it and leave a note saying that we'll take care of it later? It seems useless to make it iterate and just keep adding `_1` anyway. We could just add an underscore each time... Something like this:
```
...
while ( scope.report[ new_id ] !== undefined  ) {
  new_id += '_';
}
scope.safe_id = new_id;
```

It doesn't seem very clear, though.

lib/steps.js
plocket: I like it. Maybe we can group the changes we decide on so that we don't have to test this a million times. Or maybe they should be separate. They're not really related... Guess I'm just getting lazy in my old age.

lib/steps.js
plocket: Thanks! I made this branch before the tags branch was merged in. The tags already add the language name, so I can remove this.

lib/steps.js
niharikasingh: I think the above code snippet will work, and we could add a comment explaining what the code snippet does. We don't need to add the new code snippet in right now, but I do think we should remove what currently exists, or make a ticket explaining it needs to be removed.

plocket:
New tests: https://github.com/plocket/docassemble-MAEvictionDefense/actions/runs/490392678

lib/steps.js
plocket: Eh, I did it anyway. It's not ideal, but it is what it is.

plocket:
It does pass, and this was approved before, so I think I'll just merge and we'll deal with fallout if there is any.

PR #122, opened 2021-01-14
Make getField more flexible for generated tests. See details. Close #113

### LINKED ISSUE #113: Allow mixing choice names and values for possible generated regression tests
Bear with me here. I was requested to explore the possibility of giving developers the ability to automatically generate tests. They would go through the interview, taking a certain path, and then they would save the resulting JSON object representing the variables. The automatic test generator could [use that JSON object to generate a test for that particular path](https://repl.it/@plocket/generatetestsfromobj).

It seemed impossible because the JSON object doesn't allow differentiating between some types of inputs. Sometimes a value should be the name of a choice in a set of radio buttons. Other times it should be the value to give to an input element.

Looking at our code and the da DOM, though, I think we may be able to make our code flexible and permissive enough to make up for that without causing incorrect behavior.

There is only one type of field that actually needs all three pieces - variable name, choice name, and value. The `.getField()` could use both choice name and value when looking for the element because the incorrect option won't accidentally find a field. When setting the value, if the given data only has two values, it can use the second value as the 'setting' value and otherwise behave as it does now (sometimes humans might include all three values when not strictly necessary because they're confused or a sentence sounds incomplete without all three values).

One of my main concerns is with the prospect of generated tests. I'm not sure we can build something robust enough to work consistently enough that a human doesn't have to edit the code and that the generated code might be too confusing for a human to edit to fix the problem.

A lesser concern is that It might  seem to provide guidance to the human about how to write the variable table, but it would be slightly wrong. Not sure it would be wrong enough to really cause an issue.

On a possible plus side, if the generated tests aren't super off-base, they could give the developer a foundation for creating other tests that take a similar path.

plocket:
Closes #113
- Make getField more flexible so table columns can be
interchangable for generated tests.
- Find only enabled fields so out-of-order table steps
will not try to fill in disabled fields.

Test run: https://github.com/plocket/docassemble-MAEvictionDefense/runs/1704204877?check_suite_focus=true

[This version numbering thing is getting complicated. We should figure out a standard way to handle it.]

### DIFF
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -329,29 +329,31 @@ module.exports = {
 
     // `$x` always gives a list, so just make sure all results are lists
     let all_attempts = [
-      await scope.page.$$(`button[name="${ base64_name }"][value="${ choice_name }"]`),  // button
-      await scope.page.$$(`button[name="${ base64_name }"][value="${ set_to }"]`),  // button
-      await scope.page.$$(`*[data-saveas="${ base64_name }"] input`),  // text input showif
-      await scope.page.$$(`*[data-saveas="${ base64_name }"] textarea`),  // textarea showif
-      await scope.page.$$(`*[data-saveas="${ base64_name }"] select`),  // select showif/hideif
+      await scope.page.$$(`button[name="${ base64_name }"][value="${ choice_name }"]:not([disabled])`),  // button
+      await scope.page.$$(`button[name="${ base64_name }"][value="${ set_to }"]:not([disabled])`),  // button
+      await scope.page.$$(`*[data-saveas="${ base64_name }"] input:not([disabled])`),  // text input showif
+      await scope.page.$$(`*[data-saveas="${ base64_name }"] textarea:not([disabled])`),  // textarea showif
+      await scope.page.$$(`*[data-saveas="${ base64_name }"] select:not([disabled])`),  // select showif/hideif
       // Radios and checkboxes
       // radio: `name` = base64 var name, `choice_name` = var name of choice
       // `yesno`/`noyes` checkbox: `name` = base64 var name, choice always 'True'
       // checkbox: `name` = base64 var name + stuff + base64 choice var name + stuff (see function)
-      await scope.page.$$(`input[name="${ base64_name }"][value="${ choice_name }"]`),  // radio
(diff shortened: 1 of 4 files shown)


lib/scope.js
```diff
@@ -327,47 +327,59 @@ module.exports = {
...
       }
-    }
 
-    // If we get out of the for loop, we didn't find the element
-    let err_message = `Cannot find the input with the variable name "${ var_name }"`;
-    if ( choice_name ) {
-      throw new ReferenceError( `${ err_message } and the value "${ choice_name }. Is it hidden?"` );
-    } else {
-      throw new ReferenceError( `${ err_message }. Is it hidden?` );
+      let other = document.querySelector(`:not(button)[name="${ base64_name }"]:not([disabled])`);
```
niharikasingh, line 368: Add the previously-existing `others (not showif)` above this line, so we don't lose what element this is looking for.

niharikasingh (approved):

lib/scope.js
plocket: Above this line would be before `firstChoiceInput`. The '`if ( choice_name )...` stuff is being deleted. Did you mean to describe putting it above `firstChoiceInput`? If so, we should discuss that more because my thinking is that this would be a catch all for anything that was left over that could possibly match the given specifiers.

niharikasingh: I meant: 

```javascript
      // Put these two last so they'll come last in the loop and won't override other options  <----- this talks about *two* final options 

      // 'none of the above' checkbox  <------ this is the first final optiion 
      let firstChoiceInput = document.querySelector(`input#${ base64_id_of_first_choice }`);
      if ( firstChoiceInput ) {
       let fieldset = firstChoiceInput.closest('fieldset');
       let noneOfTheAboveCheckbox = fieldset.querySelector('input.danota-checkbox:not([disabled])');
       if ( noneOfTheAboveCheckbox ) { return noneOfTheAboveCheckbox; } 
      }

      // <------ need a comment here explaining what the second of the two final options is 
      let other = document.querySelector(`:not(button)[name="${ base64_name }"]:not([disabled])`);
      if ( other ) { return other; }

      return null;
    }, { selectors, base64_id_of_first_choice, base64_name });
```

plocket: Ah, I finally understand. Thanks!

PR #136, opened 2021-01-21
Fix continue button fields need a linear order to work. Fix #115

### LINKED ISSUE #79: race condition breaks/subverts table-based 'story'/scenario tests
[Edit]
Upgrading this to a bug.

### How to replicate
1. If the testing project already exists on the dev server, run `npm run takedown`.
1. Run `npm run setup` to set up a brand new project.
1. Run tests with (I believe) 80cc3edf8ab064f74e5b790219a83af543c9d59a.
1. The second `signature` question id page will time out.
See https://github.com/plocket/docassemble-MAEvictionDefense/runs/1742130312?check_suite_focus=true

### Why?
The `signature` canvas page times out because the `rental agreement` page eats up the signature rows for some unknown reason. This is how it happens:
1. `rental agreement` loops through the first time. The selectors in `getField()` find the most specific fields possible based on the values given. The 'continue' button is hit.
1. The fact that the next page hasn't finished loading - that we're still on the `rental agreement` page - is not detected. The code thinks the 'continue' action was completed successfully.
1. The `rental agreement` page goes through the loop again. This time, any `/sign` rows somehow catch something. There is no `var_name` or `choice_name` value, just the `set_to` value of `/sign`. I think the others are `undefined`. I'm not sure what they catch, though. I do know they are `INPUT` fields with `type="hidden"`. That field uses up any and all `/sign` rows in 'setting' the value of that field. The 'continue' button is 'hit' again.
1. The `rental agreement` page goes through the loop a third time. This time it catches no variables and hits 'continue' again.
1. The next page is reached.
1. When the `signature` page is found, there's no `/sign` row for it.

### Notes
1. We need to be able to find `hidden` fields. Most fields are `hidden` so that they can be interacted with, but so that only pretty stuff is actually visible.
1. There's something that's not being awaited. If you put delays in certain places, an error will be caused by the page navigation even though the process of finding the nodes has not completed and the code should still be waiting for the next page DOM content to load. I've checked all our functions. None of the ones affecting this were missing `awaits`. When I added `awaits` to the ones we (I) had missed, this didn't solve the problem.

---
[Orignal post]
To make tables work as 'stories', we need the variable rows to be able to be listed out of order. That's fine most times, but we've run into at least one situation where we can't track down a race condition, requiring the developer to break the table into two parts. It means they need a `When the terminal question id is "some id"` step to stop early, then put in their own `And I wait 2 seconds` before creating another table for the rest of the variables.

For those tests order _does_ matter, which is not what we want from stories.

Can we somehow allow some way for developers to influence this situation? An extra wait after hitting 'continue' for some specific pages? An extra wait at the start of some pages? An extra wait after setting a field?

### LINKED ISSUE #115: Test with a table where cont field is set before other vars
Possible problem with var table.
- [x] What if a continue button field variable value appears in the table before other field values that are not required and should lead to a branching path? Will the interview continue?
- [x] Does a continue button field variable appear in the JSON variables? That is, is this in danger of happening arbitrarily?

[Edit]
To clarify: Most likely to happen with generated tests, but humans might do it too.

If these things are true:
1. The 'continue' button has a variable name.
1. There are also other fields being set on the page.
1. The continue button's variable row comes before the row of one of the other fields.

Then the page will try to continue before all the field values are set.

### LINKED ISSUE #139: `scope.getField()` `set_to` doesn't find all the fields
I've fixed this one locally while working on #136, but I'm putting this here in case of buses.

`set_to` was being made safe for html for finding fields using that value because buttons and such don't like the kind of strings that need to be passed in for `textarea`s. The problem is that it also gets rid of pretty legitimate values, like things with spaces in them (which docassemble allows in values). New solution:

Use a try/catch when looking for the selector so that `set_to` can stay accurate.

### Test to write:
1. Include multi-line text
1. Include a button with non-html-safe `set_to` value, like something with a space in it.

plocket:
[Fix #79, fix #115, fix #139] Would love some better ideas for handling signing. It's a bit all over the place. Seems too coupled to me. Relatedly, I recommend reading scope.processVar by itself instead of the diff. Easier to see what's going on.

Sadly, as a consequence of these changes the tilde symbol for the signature step is gone :(

Test started at 8:40: https://github.com/plocket/docassemble-MAEvictionDefense/runs/1742130312?check_suite_focus=true

### DIFF
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -427,7 +427,7 @@ module.exports = {
     if ( !disabled ) {
       let tag = await handle.evaluate( elem => { return elem.tagName; });
       await scope.setField[tag]( scope, { handle, set_to });
-    }
+    } else { console.log( set_to, disabled ); }
 
     return disabled;
   },
@@ -598,23 +598,26 @@ module.exports = {
     // each field could reveal other fields.
     // +1 to trigger `while` the first time around
     let prev_num_disabled = disabled_fields.length + 1;
+    console.log( disabled_fields.length );
     while ( disabled_fields.length > 0 && prev_num_disabled > disabled_fields.length ) {
       prev_num_disabled = disabled_fields.length;
 
-      let found_indexes = [];
+      let enabled_indexes = [];
       for ( let d_i = 1; d_i < disabled_fields.length; d_i++ ) {
         let { row, handle } = disabled_fields[ d_i ];
-        let was_disabled = await scope.setVariable( scope, { handle, set_to: row.set_to });
+        console.log( 'loop disabled:', row.set_to );
+        let is_still_disabled = await scope.setVariable( scope, { handle, set_to: row.set_to });
         // Prepare to quickly remove fields that were found and set
-        if ( !was_disabled ) {
-          found_indexes.push( d_i );
+        if ( !is_still_disabled ) {
+          enabled_indexes.push( d_i );
           process.stdout.write(`\x1b[36m${ '*' }\x1b[0m`);  // var was set
         }
       }  // ends for every remaining disabled field
 
-      for ( let d_i of found_indexes ) {
+      // Remove fields that have now been set
+      for ( let d_i of enabled_indexes ) {
         // Fast way to remove item from list. See https://stackoverflow.com/a/638404.
-        disabled_fields[ d_i ] = disabled_fields[ disabled_fields.length - 1] // last item
+        disabled_fields[ d_i ] = disabled_fields[ disabled_fields.length - 1 ] // last item
         disabled_fields.pop();
       }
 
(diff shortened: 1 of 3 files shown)


plocket:
Welp, there seems to be some race condition going on with the signature screen. I'm not sure why. I think I'm using the same code as before to wait for the canvas to appear and it was working fine then. Ugh.

plocket:
Leave that test alone - DO NOT rerun it. We can run a new test to see what happens after we... do something...

plocket:
Started https://github.com/plocket/docassemble-MAEvictionDefense/runs/1751842206?check_suite_focus=true at 5:40. May actually solve the problem by giving a fake `var` column to rows without anything in there.

niharikasingh (approved):
There is a lot going on here, but I don't see any obvious logical flaws that jump out at me. Of course, because of the sheer size/scope of the changes, it's very possible that I've missed something. If there is a tricky portion of these changes that you specifically want me to look closely at, let me know.

plocket:
I'd love to re-examine the signature handling, @niharikasingh , but I don't think it's causing any practical issues at the moment.

PR #143, opened 2021-01-26
add .env-example

### LINKED ISSUE #2: Make .env-example file
For local developers. Might as well make it available.

berit:
closes #2

### DIFF
--- a/pushing_testing_files/.env-example
+++ b/pushing_testing_files/.env-example
@@ -1,11 +1,20 @@
+# Username and password to log in to a docassemble playground to run tests
+# PLAYGROUND_ID can be found in the url of an interview launched from that playground
+# It should look something like this:
+# https://www.docassemble-server-url.com/interview?i=docassemble.playground[PLAYGROUND_ID]%3Ainterview.yml#page1
 PLAYGROUND_EMAIL=example@example.com
 PLAYGROUND_PASSWORD=12345
 PLAYGROUND_ID=1
 
-BASE_URL=https://www.example.com
-REPO_URL=https://github.com/example/example
+BASE_URL=https://www.docassemble-server-url.com
+REPO_URL=https://github.com/user_name/repo_name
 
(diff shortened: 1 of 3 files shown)


berit:
@plocket I wasn't sure if this was what you wanted or how to test it

pushing_testing_files/.env-example
```diff
@@ -0,0 +1,11 @@
+PLAYGROUND_EMAIL=example@example.com
+PLAYGROUND_PASSWORD=12345
+PLAYGROUND_ID=1
+
+BASE_URL=https://www.example.com
```
plocket, line 5: How about https://www.docassemble-server-url.com?

pushing_testing_files/.env-example
```diff
@@ -0,0 +1,11 @@
...
+REPO_URL=https://github.com/example/example
```
plocket, line 6: How about https://github.com/user_name/repo_name?

pushing_testing_files/.env-example
```diff
@@ -0,0 +1,11 @@
...
+
+BRANCH_PATH=heads/master
```
plocket, line 8: How about repo_name/main?

plocket (approved):
This is looking pretty darn good. You're spot on. Not blocking, but I think we could improve the names in `.env` to guide the developer a bit more.

berit:
@plocket updated the defaults mostly to your requests. Also added some comments

plocket (approved):
Double checked on `BRANCH_PATH` syntax and the one in here is right. We should make an issue to change the convention for the name to `BRANCH`. It probably wouldn't need an explanation at that point.

plocket:
Did not know that comments in `.env` are now a Thing. Will wonders never cease...

plocket:
Need to remember to add this to CHANGELOG on next version bump or something. And maybe to documentation.

PR #147, opened 2021-02-01
Move logic to field gathering loop, report on remaining vars

plocket:
I ended up moving the signature logic to the loop that was gathering fields. The outer, `while`, loop meant a lot of complications. Tell me what you think.

Also added remaining vars to the report. It's not quite in the right spot, but it does let us see if the signature solution is working. See https://github.com/plocket/docassemble-MAEvictionDefense/actions/runs/528828913

### DIFF
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -562,7 +562,16 @@ module.exports = {
         set_to: row.value,
       });
       if ( handle ) {
-        fields_to_set.push({ row, handle, tag, type });
+        if ( (tag === 'CANVAS') && row.value === '/sign' ) {
+          // A signature field on a signature page is the only field on that page
+          fields_to_set.push({ row, handle, tag, type });
+          // Get the rest of the variables before...
+          remaining_vars = remaining_vars.concat( var_data.slice( var_data.indexOf(row) + 1 ));
+          // ...stopping the loop so no more /sign rows are gathered
+          break;
+        } else {
+          fields_to_set.push({ row, handle, tag, type });
+        }
       } else {
         remaining_vars.push( row );  // keep remaing vars around for future pages
       }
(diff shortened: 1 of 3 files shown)


lib/scope.js
```diff
@@ -575,11 +584,10 @@ module.exports = {
...
 
       prev_length = fields_to_set.length;
-      // Go through each field on the page, seeing if you can set its
-      // variable
+      // Go through each field on the page, seeing if you can set its variable
       let remaining_fields = [];
-      for ( let i = 0; i < fields_to_set.length; i++ ) {
-        let field = fields_to_set[i];
+      for ( let field_i = 0; field_i < fields_to_set.length; field_i++ ) {
+        let field = fields_to_set[field_i];
```
niharikasingh, line 590: We could even go back to the old style loop here, since we don't need to use the index var `i` anymore (you're using the `var_data.indexOf(row)` function instead.

lib/scope.js
```diff
@@ -562,7 +562,16 @@ module.exports = {
...
+        if ( (tag === 'CANVAS') && row.value === '/sign' ) {
+          // A signature field on a signature page is the only field on that page
+          fields_to_set.push({ row, handle, tag, type });
+          // Get the rest of the variables before...
+          remaining_vars = remaining_vars.concat( var_data.slice( var_data.indexOf(row) + 1 ));
+          // ...stopping the loop so no more /sign rows are gathered
+          break;
+        } else {
+          fields_to_set.push({ row, handle, tag, type });
+        }
```
niharikasingh, line 574: This works for me! This is actually what I originally had, but I was waffling between including this code here or in the next loop.

lib/scope.js
```diff
@@ -562,7 +562,16 @@ module.exports = {
         set_to: row.value,
       });
       if ( handle ) {
-        fields_to_set.push({ row, handle, tag, type });
+        if ( (tag === 'CANVAS') && row.value === '/sign' ) {
+          // A signature field on a signature page is the only field on that page
+          fields_to_set.push({ row, handle, tag, type });
+          // Get the rest of the variables before...
+          remaining_vars = remaining_vars.concat( var_data.slice( var_data.indexOf(row) + 1 ));
```
niharikasingh, line 569: Is it a safe assumption to make that we always just want the first `/sign` in the var_data array? It seems safe enough, but wanted to flag it in case this raises any concerns for you.

niharikasingh (approved):
Approving provisionally, assuming tests pass 🤞 I'll check back later tonight to see if they have.

lib/scope.js
plocket: Eh, this works too!

lib/scope.js
plocket: Yeah, it took some experimentation (which you really couldn't do, sorry)

lib/scope.js
plocket: All `/sign` rows are the same currently, so it matters even less than usual (since tables are supposed to not care about order).

Issue #158, opened 2021-02-28
Refactor: get all fields on the page then compare to story table


plocket:
Edit 2 (03/23):
Turning this into a reworking of the framework. This need was highlighted by the trouble calculating variable names of proxy variables (x, i, etc).

Changes ~for v2~:
- [x] Story table will have three columns that are basically `| var name | value | checked |`. We need another word for `checked`. This will make it much easier for developers to know what to put in each column. It will also make it more straightforward to check against those values. Not completely sure this part will work out.
- [x]  Support previous format of story table. Still not sure what 'checked' should be.

Refactors ~for v2~:
- [x] First use the string of the html to collect the data about the page (sought var, field selectors, field values), then check _that_ against the table, setting values appropriately.
- [x] Use sought var to figure out actual var names and values for proxy vars (x, i, j...).
- ~Clean up `scope.js`. Are there outdated functions in there? Do we need to split it into multiple files?~ This is holding up this issue and really belongs in its own issue.

Testing changes ~for v2~:
- [x] Add unit tests based on data, like html strings. Simpler to maintain. Less conflict with different tests trying to use the same project name (though the fix for that is in an issue already).

---
Edit 1:
We may have to do a big refactor to account for this newly discovered situation. That may include getting rid of one column in the table. Which of the three columns in the table need what information is pretty confusing as it is anyway.

---
https://docassemble.org/docs/fields.html#index%20variables. Also relevant to generic objects: https://docassemble.org/docs/modifiers.html#generic%20object.

This poses a particular problem. DOM uses literally 'i', not var name. This is fine as long as the answers are as interchangeable as this implies. If the specific variables' answers matter later on, it's more of a problem. It also poses a problem for testing generated PDFs because the order in which the vars might be set is not controlled, so you have the potential of coming out with a different PDF just because of a small change. It wouldn't even make a difference if we sort the rows alphabetically or something. If the developer changes the name of the variable, it will get out of whack. What is the solution here?

A simple example: `your_past_benefits['State Veterans Benefits']`. Here we can just replace what's inside the brackets with an 'i'. If something gets more complicated (e.g. `thing[i]['bar']`), or `thing['bar'][i]` or `thing[i][j]` do we create all combos?

Trying to discuss this upstream.

plocket:
Experiments going on here: https://repl.it/@plocket/findi#index.js

plocket:
[...]

PR #160, opened 2021-03-03
Re #158, temp fix for simple index_var detection

plocket:
Vars might be set out of order. Addresses #158, but does not close it. This will let some tests work under some circumstances. [This test shows the result](https://github.com/plocket/docassemble-DeadBrokeDads2/runs/2023969728?check_suite_focus=true) (passes).

### DIFF
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -431,13 +431,19 @@ module.exports = {
 
       // Put these two last so they'll come last in the loop and won't override other options
 
+      // TODO: Require the developer to use 'None' when they mean 'none of the above'??? Or
+      // maybe something more obscure so they're less likely to have written that variable
+      // intentionally, like NOTA or DA-NOTA or NONE.
       // 'none of the above' checkbox
+      // Only one (or none) of the below selectors will find a field. Either
+      // `base64_id_of_first_choice` or `base64_id_of_index`
       let firstChoiceInput = document.querySelector(`input#${ base64_id_of_first_choice }`);
       if ( firstChoiceInput ) {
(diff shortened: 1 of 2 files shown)


lib/scope.js
```diff
@@ -402,17 +438,29 @@ module.exports = {
        let noneOfTheAboveCheckbox = fieldset.querySelector('input.danota-checkbox');
```
niharikasingh, line 443: Make clear somehow that only one of `base64_id_of_first_choice` and `base64_id_of_index` will exist. The way we check for both made it seem to me like both of these might be true somehow. Same for `base64_var_name` and `base64_index`.

lib/scope.js
```diff
@@ -380,9 +390,35 @@ module.exports = {
       // 'none of the above' checkbox showif???
       `input[name="${ name_attr_B }"]`,  // checkbox
       `input[name="${ name_attr_R }"]`,  // checkbox alt
+
+      // simple index var case as described above
+      `button[name="${ base64_index }"][value="${ choice_name }"]`,  // button
```
niharikasingh, line 395: I think we can avoid all these repetition in selector clauses by setting a variable, e.g., `base_64_var_or_index` to either a variable name or an index variable. We have the ability to detect whether a given argument `var_name` is a variable name or an index name. So once we know which one it is, we can just use that.

niharikasingh (approved):

lib/scope.js
plocket: This would be negated by using the second comment to clarify the code. Thanks!

PR #168, opened 2021-03-13
Fix infinite loop caused when table button var causes user errors

plocket:
Test for this branch should error (user got an error) instead of getting stuck in
infinite loop. https://github.com/plocket/docassemble-cucumber/runs/2099929359?check_suite_focus=true

We should set up some ways to assert errors happen for this repository. Not sure how exactly. Some kind of step that causes a try/catch to throw vs. not throw or something?

### DIFF
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -679,7 +679,22 @@ module.exports = {
     for ( let data_i = 0; data_i < fields_to_set.length; data_i++ ) {
       let { handle, tag, type, row } = fields_to_set[ data_i ];
       if ( tag === 'BUTTON' && type === 'submit' ) {
+        let page_url = await scope.page.url();
         await scope.setVariable( scope, { handle, set_to: row.value });
+      
+        // Have to catch errors right now or may get stuck in an infinite loop?
+        // TODO: Don't catch errors - maybe the developer expected/wanted an error
+        // For now they'll have to stop the story table early and do the rest the slow way
+        // TODO: Discuss: add the option of throwing when an error happens? Per page?
+
+        // Make sure no system or user error messages appear
+        // Discuss moving this out of here. Having to pass in `id` just for
+        // these messages smells bad. Con: Already a lot of code out there.
+        let error = await scope.waitUntilContinued( scope, { url: page_url, id, remaining_vars });
+        if ( error.was_found ) {
+          await scope.throwPageError( scope, { remaining_vars: remaining_vars, id: id });  
+        }
+
         already_continued = true;
         process.stdout.write(`\x1b[36m${ 'v' }\x1b[0m`);  // pressed button successfully
 
@@ -699,6 +714,7 @@ module.exports = {
     // If didn't already press a continue button, press one now.
     if ( !already_continued ) {
       // All fields on this page should be taken care of
+      let page_url = await scope.page.url();
       await scope.continue( scope );
       // Have to catch errors right now or may get stuck in an infinite loop?
       // TODO: Don't catch errors - maybe the developer expected/wanted an error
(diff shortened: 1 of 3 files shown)


lib/scope.js
```diff
@@ -679,7 +679,22 @@ module.exports = {
     for ( let data_i = 0; data_i < fields_to_set.length; data_i++ ) {
       let { handle, tag, type, row } = fields_to_set[ data_i ];
       if ( tag === 'BUTTON' && type === 'submit' ) {
+        let page_url = await scope.page.url();
         await scope.setVariable( scope, { handle, set_to: row.value });
+      
+        // Have to catch errors right now or may get stuck in an infinite loop?
+        // TODO: Don't catch errors - maybe the developer expected/wanted an error
```
niharikasingh, line 686: I think this will be a useful future step. 

In Python testing, the tests always throw errors. If they are expected, the user can say something like: 
```python
def test_passes():
    with pytest.raises(Exception) as e_info:
        x = 1 / 0
```
If the user doesn't realize an error will happen, then the test will fail. 
```python
def test_fails():
    x = 1 / 0
``` 
We could try something analogous. 

BUT that seems like a nice-to-have feature, not a must-have feature right now (at least to me).

lib/steps.js
```diff
@@ -177,6 +177,7 @@ When(/^the user gets to ?(?:the)?(?: ?question)?(?: id)? "([^"]+)" with this dat
   // still have to be some automatic continuing.
   // It also seems harsh, but more reliable.
   // Ask developers what they'd like.
+  let count = 0;
```
niharikasingh, line 180: Is this being used for anything?

tests/features/generated_story_tables.feature
```diff
@@ -0,0 +1,65 @@
...
+- [x] Basic direct fields
+- [x] Able to select first choice (not thorough)
+- [x] Able to select second choice (not thorough)
+- [x] Able to select third choice (not thorough)
+- [x] Hidden fields
+- [x] Nested hidden fields listed out of order
+- [x] Continue button field listed before unrequired fields
+
+@generated @slow @1 @out_of_order
+Scenario: Fields are listed out of order
```
niharikasingh, line 33: Specify that "Fields are listed out of order **so this test will generate an error**" or something similar?

lib/scope.js
```diff
@@ -741,15 +775,32 @@ module.exports = {
...
     }
 
-    return { error_found, error_handlers };
-  },
+    return { was_found, error_handlers };
+  },  // Ends scope.checkForError()
+
+  throwPageError: async function ( scope, { error_handlers = {}, remaining_vars = {}, id = '' }) {
+    /* Throw a cucumber error with the error that appeared on the page. */
+    let was_system_error = error_handlers.system_error !== undefined;
```
niharikasingh, line 792: Can `error_handlers.system_error` be `false`? If so, should the comparison be `error_handlers.system_error === true` instead? 

(Looks like we're not using the error_handlers variable right now, so this might be a moot point.

niharikasingh (approved):

lib/scope.js
plocket: Good thoughts and issue made!

lib/steps.js
plocket: Reply made in #171!

tests/features/generated_story_tables.feature
plocket: Fields out of order should not generate an error. We were just testing that this particular occasion. Our code (not the test) caused an error incorrectly, which then triggered this infinite loop. This was just a happy circumstance that let us fix the problem with the infinite loop. That's why I kept the error around instead of making a PR with them both fixed - we could see this behaving properly. In the future we'll have to work out testing our framework's errors (which is different than the errors that developers should be testing).

Does that make sense? I'm not sure I articulated it very well. We can discuss and clarify more if needed and add it in here posthumously.

lib/scope.js
plocket: Really good point. Turned into #172.

tests/features/generated_story_tables.feature
niharikasingh: Yeah it makes sense now :) This is one of the testing-framework-tests we should write then at some point in the future.

PR #169, opened 2021-03-13
Fix `.setVariable()` some elems too far out of view, add some debugging

plocket:
[`.focus()` seemed to work locally, but doesn't seem to work on GitHub for some reason. I'll try `.scrollIntoView(false)` next.]

1. I've been getting a bit tired of adding elaborate debugging stuff every time there's weird behavior. I added stuff that will be silent, but that I use all the time - coloring the DOM nodes. That should show up on screenshots if `DEBUG=1` in SECRETS, which no one knows about yet, so it shouldn't confuse anyone.
2. Puppeteer tries to scroll up to make an element visible before interacting with it. First, [that's known to not always be successful](https://github.com/puppeteer/puppeteer/issues/6462#issuecomment-717163409). Secondly, docassemble's header actually blocks the fields when they're at the top of the screen, so we need `false` in `.scrollIntoView(false)` to scroll things to the bottom of the screen instead. This concentrates on fixing it for the story table, though. It fixes it in `.setVariable()` and the functions called from it. I _think_ it It won't fix it for some other types of sentences.

Also added to the change log to prep for changes.

Just FYI, this passes, but actually shouldn't pass. The `showif_checkboxes_other` should be selecting `showif_checkboxes_nota_2`, but it's selecting `showif_checkboxes_nota_1`. No idea why yet.

Great to have this interview to test against!

### DIFF
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -679,7 +679,22 @@ module.exports = {
     for ( let data_i = 0; data_i < fields_to_set.length; data_i++ ) {
       let { handle, tag, type, row } = fields_to_set[ data_i ];
       if ( tag === 'BUTTON' && type === 'submit' ) {
+        let page_url = await scope.page.url();
         await scope.setVariable( scope, { handle, set_to: row.value });
+      
+        // Have to catch errors right now or may get stuck in an infinite loop?
+        // TODO: Don't catch errors - maybe the developer expected/wanted an error
+        // For now they'll have to stop the story table early and do the rest the slow way
+        // TODO: Discuss: add the option of throwing when an error happens? Per page?
+
+        // Make sure no system or user error messages appear
+        // Discuss moving this out of here. Having to pass in `id` just for
+        // these messages smells bad. Con: Already a lot of code out there.
+        let error = await scope.waitUntilContinued( scope, { url: page_url, id, remaining_vars });
+        if ( error.was_found ) {
+          await scope.throwPageError( scope, { remaining_vars: remaining_vars, id: id });  
+        }
+
         already_continued = true;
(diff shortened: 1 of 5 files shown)


lib/steps.js
```diff
@@ -177,6 +177,7 @@ When(/^the user gets to ?(?:the)?(?: ?question)?(?: id)? "([^"]+)" with this dat
   // still have to be some automatic continuing.
   // It also seems harsh, but more reliable.
   // Ask developers what they'd like.
+  let count = 0;
```
niharikasingh, line 180: My comments on the other PR #168 carry over here too, since this one builds on that one.

lib/scope.js
```diff
@@ -483,6 +483,14 @@ module.exports = {
       let { tag, type } = await handle.evaluate( elem => {
         return { tag: elem.tagName, type: elem.getAttribute('type') };
       });
+
+      if ( env_vars.DEBUG ) {
```
niharikasingh, line 487: This is a good idea. I like this!

lib/scope.js
```diff
@@ -504,6 +512,9 @@ module.exports = {
     });
 
     if ( !disabled ) {
+      // da's header covers the top of the section containing the fields which means
+      // sometimes it blocks nodes from clicks. Scroll item to the bottom instead.
+      await handle.evaluate( elem => { elem.scrollIntoView(false); return elem.scrollTop });
```
niharikasingh, line 517: Do we need to `return elem.scrollTop` here? We don't seem to be using that value.

niharikasingh (approved):

lib/steps.js
plocket: Made #171

lib/scope.js
plocket: Added to #171, remove unused dev code

Issue #173, opened 2021-03-15
Found: New inputs with useful data at the bottom of each page. How to use?


plocket:
Found hidden input elements at the bottom of the page. It's not clear how much use the data can be to us, but it does have the sought var for #158 at the very least. They come in `name`/`value` attributes. I've noted them below with an example and some notes. [The below is an example of the direct standard fields testing page, I think, and what those nodes look like and how they might correspond to data we need on the page.]

### name = _checkboxes:
value =
'eyJZMmhsWTJ0aWIzaGxjMTk1WlhOdWJ3IjogIkZhbHNlIiwgIlkyaGxZMnRpYjNobGMxOXZkR2hsY2x0Q0oxa3lhR3haTW5ScFlqTm9abUl6VW05YVdFcG1Zak5DTUZoNlJTZGQiOiAiRmFsc2UiLCAiWTJobFkydGliM2hsYzE5dmRHaGxjbHRDSjFreWFHeFpNblJwWWpOb1ptSXpVbTlhV0VwbVlqTkNNRmg2U1NkZCI6ICJGYWxzZSIsICJZMmhsWTJ0aWIzaGxjMTl2ZEdobGNsdENKMWt5YUd4Wk1uUnBZak5vWm1JelVtOWFXRXBtWWpOQ01GaDZUU2RkIjogIkZhbHNlIiwgImNtRmthVzlmZVdWemJtOCI6ICJGYWxzZSJ9'

'{"Y2hlY2tib3hlc195ZXNubw": "False", "Y2hlY2tib3hlc19vdGhlcltCJ1kyaGxZMnRpYjNoZmIzUm9aWEpmYjNCMFh6RSdd": "False", "Y2hlY2tib3hlc19vdGhlcltCJ1kyaGxZMnRpYjNoZmIzUm9aWEpmYjNCMFh6SSdd": "False", "Y2hlY2tib3hlc19vdGhlcltCJ1kyaGxZMnRpYjNoZmIzUm9aWEpmYjNCMFh6TSdd": "False", "cmFkaW9feWVzbm8": "False"}'
checkboxes_yesno: 'False'
checkboxes_other[B'Y2hlY2tib3hfb3RoZXJfb3B0XzE']: 'False'
checkbox_other_opt_1
checkboxes_other[B'Y2hlY2tib3hfb3RoZXJfb3B0XzI']: 'False'
checkbox_other_opt_2
[...]

PR #176, opened 2021-03-21
#158 Start refactor for parsing proxies and other improvements

plocket:
Addresses #158 and refactoring. These functions don't affect anything right now. I'm just working out how they may work and want to bring people on the journey. Want to avoid one huge PR all at once.

The idea is to parse the DOM first, as a string. That seems like it will be more well-defined than guessing at selectors and figuring out the correct order to search for them every time we need to add another discovery. Later, this will be compared against the table, which is going to change, in order to match the right variables.

These functions also currently only handle (and can be locally tested with) direct standard fields (e.g. no `show if`s and no proxy/alias variables like `thing[i]`). Test locally with `npm run unit`. I hope to incorporate those tests into the github workflow soon. We will add more functionality once this functionality has been reviewed.

Starts to use new nodes found (discussed in #173). 

I'm not sure what to call `var_name`/`var_names` as they could be proxy names, not the actual variable name that will match up in the table.

P.S. Also adds accidentally a test that currently doesn't do anything. `getMatchingRow.test.js` and its fixtures, `tables.fixtures.js`. 'Fixtures' are files for data that's needed for tests.


niharikasingh (approved):
I think the tests are a good idea 👍 

The logic makes sense to me, but the entire paradigm shift is so large that I'm sure I've missed side-effects of this change that will become clearer later. I think continuing to use tests will mitigate bad side-effects to a large extent.

PR #177, opened 2021-03-23
#158 Match a table row to  a field

plocket:
Addresses #158 refactor for v2. One of the next steps is being able to match a row in the table to a field. The new columns in the story table have been changed to basically `| var name | value | checked |`, though now that I think about it, 'checked' doesn't quite make sense. I'm now thinking maybe 'select' would be better, but it still doesn't seem clear for some reason.

Need to remember to normalize character case when first processing the table.

### DIFF
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -511,7 +511,9 @@ module.exports = {
     *  TODO: Not sure where to bring in cheerio
     *  TODO: Should this also return the sought variable for proxy stuff?
     *  TODO: Pass in `$` instead?
-    *  TODO: Switch story columns so true/false are the only values in the third column. */
+    *  TODO: Switch story columns so true/false are the only values in the third column.
+    *  TODO: Instead of getting tag, type, and selector here, should we just
+    *     pass around the node object itself? That would be hard to test for. */
     
     /*
     *  Development plan:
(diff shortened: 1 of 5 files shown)


lib/scope.js
```diff
@@ -805,8 +813,80 @@ module.exports = {
...
+          let names_and_values_match = names_match && table_row.value === guess.value;
+
+          switch ( field.tag ) {
+            case 'canvas':
+              if ( names_and_values_match ) { rows_match = true; }
+              break;
+
+            case 'button':
+              if ( names_and_values_match ) { rows_match = true; }
+              break;
```
niharikasingh, line 860: Since button and canvas have the same code, you could combine them

```javascript
case 'canvas': 
case 'button': 
    # something 
```

lib/scope.js
```diff
@@ -805,8 +813,80 @@ module.exports = {
...
+
+            case 'textarea':
+              if ( names_match ) { rows_match = true; }
+              break;
+          }
+
+          if ( rows_match ) {
+            matches.push({ field: field, table: table_row });
+            continue;
```
niharikasingh, line 882: If there is only one match for each table and field row, you could probably break here instead of continuing.

niharikasingh (approved):
I love the tests!

lib/scope.js
plocket: True! Making into an issue.

lib/scope.js
plocket: Good eye! Made #184 with some discussion.

PR #178, opened 2021-03-23
#158 add simple show if field tests (after #177)

plocket:
#158 refactor. Add tests for simple 'show if' fields. No proxies, etc. Though it does test for `_field_n` aliases, though. I'm not sure what vocabulary to put those in. I guess they're proxies to in a way? Yet they have to be handled differently in a completely different section.


lib/scope.js
```diff
@@ -616,12 +622,16 @@ module.exports = {
...
       field_data.rows = await scope.getPossibleTableRows( scope, { node, var_names, values });
       page_data.fields.push( field_data );
 
     }  // ends for every node
 
-    // console.log( JSON.stringify( page_data, null, 2 ));
+    if ( page_data.sought_var[0] === `direct_showifs`) {
+      // console.log( field_like_names );
+      // console.log( JSON.stringify( page_data, null, 2 ));
+    }
```
niharikasingh, line 634: This looks like debugging code that isn't doing anything anymore.

lib/scope.js
```diff
@@ -805,8 +821,83 @@ module.exports = {
...
+            continue;
+          }
+
+        }  // ends for each table row
+      }  // ends for each field row possibility
+    }  // ends for each field
+
+    if ( page_data.sought_var[0] === `direct_showifs`) {
+      // console.log( JSON.stringify( matches, null, 2 ));
+    }
```
niharikasingh, line 899: more debugging code?

niharikasingh (approved):

lib/scope.js
plocket: I should have mentioned this and thanks for spotting unused code. It's not doing anything here anymore, but I continue to use it to help build further tests. I log the objects created, check over them to make sure they look right, then use them to build the `page_data.fixtures.js` values.

lib/scope.js
plocket: Same comment on continuing to use this for building test fixtures.

PR #180, opened 2021-03-23
#158 custom continue buttons that are not yesno or yesnomaybe

plocket:
#158 Add and test custom continue buttons that are not yesno or yesnomaybe

### DIFF
--- a/tests/unit_tests/getMatchingRow.test.js
+++ b/tests/unit_tests/getMatchingRow.test.js
@@ -42,7 +42,7 @@ it("matches the right table and field rows for yesnomaybe buttons", async functi
 });
 
 
-it("matches the right table and field rows for yesnomaybe buttons", async function() {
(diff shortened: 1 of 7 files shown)


tests/unit_tests/getMatchingRow.test.js
```diff
@@ -42,3 +42,18 @@ it("matches the right table and field rows for yesnomaybe buttons", async functi
 });
 
 
+it("matches the right table and field rows for yesnomaybe buttons", async function() {
```
niharikasingh, line 45: I think this should say "... for other buttons"

niharikasingh (approved):

tests/unit_tests/getMatchingRow.test.js
plocket: Thanks!

PR #185, opened 2021-03-25
#158, implement new code for live tests

plocket:
Looks like it's working! I'm a bit suspicious because it was surprisingly easy to convert. Will integrate uint tests into the workflow after adding support for the old story table format. #158


lib/scope.js
```diff
@@ -838,8 +841,10 @@ module.exports = {
...
     /* Given html_data and a list of objects describing fields and values,
-    * returns the matching fields and the data necessary to set those values.
-    * May also handle proxy vars (e.g. x, i...)
+    * returns a list of the matching fields and the data necessary to set
+    * those values: `[{ field: {}, table: {} }, ...]`. See
+    * test/unit_tests/matches.fixtures.js
+    * May in future also handle proxy vars (e.g. x, i...)
     * 
     * @param story_table {arr} - fields and the values they can be set to
     * @param story_table[n].var_name {str} - name of variable to set
```
niharikasingh, line 850: I think the `var_name` in this comment needs to be updated to `var`. 

Also, just out of curiosity, what is the reasoning behind renaming `var_name` to `var`?

lib/steps.js
```diff
@@ -217,7 +214,7 @@ When(/^the user gets to ?(?:the)?(?: ?question)?(?: id)? "([^"]+)" with this dat
         try {
           let { question_has_id, id, id_matches } = await scope.examinePageID( scope );
           // Some errors happen in there to be closer to the relevant code. We'd have to try/catch anyway.
-          let leftovers = await scope.processVar(scope, { var_data: remaining_vars, id });
+          let leftovers = await scope.processVar(scope, { var_data, id });
```
niharikasingh, line 217: `processVar` is returning errors now, not remaining variables. I think that changes how this part of the code works.

niharikasingh (changes requested):

lib/scope.js
plocket: `var_name` was the renaming - it started as `var`. I just forgot what it was when I was writing the new code.

lib/steps.js
plocket: You're right, `leftovers` isn't really useful anymore. Done!

niharikasingh (approved):

PR #186, opened 2021-03-25
#158 support v1 story table (after #185)

plocket:
#158. I think this will fully support v1 tables, though it is a bit awkward and its testing isn't super thorough yet because we don't have even a single signature page in the online interviews. Once that's done and added to the tests, I think we can be fairly confident this will,  except for the primitive proxy var (x, i, j...) setting,  support everything that the we were previously able to support.

Next I think we need to add unit tests to workflows and get started on proxy vars themselves.

### DIFF
--- a/lib/steps.js
+++ b/lib/steps.js
@@ -199,8 +199,8 @@ When(/^the user gets to ?(?:the)?(?: ?question)?(?: id)? "([^"]+)" with this dat
       // See bug at https://github.com/plocket/docassemble-cucumber/issues/79
       // May be able to remove this once converted to `getPageData`
       var: row.var || 'al_no_var_name',
-      value: row.choice || row.value,
-      checked: row.checked || row.value,
+      value: row.value || row.choice,
+      checked: row.checked || row.choice,
     }
     supported_table.push( result );
   }
--- a/lib/steps.js
+++ b/lib/steps.js
@@ -181,30 +181,6 @@ When(/^the user gets to ?(?:the)?(?: ?question)?(?: id)? "([^"]+)" with this dat
 
   let supported_table = await scope.normalizeTable( scope, { var_data: raw_var_data.hashes()});
 
-  // // If table doesn't have headers, give it headers
-  // let sample = var_data[0];
-  // if ( Array.isArray( sample )) {
-  //   let rows = raw_var_data.raw();
-  //   var_data = rows.map(( row ) => {
-  //     return { var: row[0], value: row[1], checked: row[2] };
-  //   });
-  // }
-
-  // // Support v1 tables where the columns were var, choice, value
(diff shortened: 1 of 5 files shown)


niharikasingh:
I changed the base branch to remove duplicate commits from the diff, but once you merge the other 158 branch into `main`, you can change this MR's base branch to `main`.

lib/steps.js
```diff
@@ -183,18 +183,26 @@ When(/^the user gets to ?(?:the)?(?: ?question)?(?: id)? "([^"]+)" with this dat
...
   for ( let row of var_data ) {
-    // We need to take care of this before saving this row in the report
-    row.var = row.var || 'no alkiln story table value';
+    let result = {
+      original: row,
+      // Give fake data if needed to prevent some kind of hidden INPUT field from being found
+      // See bug at https://github.com/plocket/docassemble-cucumber/issues/79
+      // May be able to remove this once converted to `getPageData`
+      var: row.var || 'al_no_var_name',
+      value: row.choice || row.value,
```
niharikasingh, line 202: Should this be the other way around? The old table headers are `var/value/checked` but here the old table's header appears second.

niharikasingh (approved):

lib/steps.js
plocket: I think you're totally right, wow. Thanks for spotting that.

plocket: I think this needs more changes. `.checked` should be `.checked` or `.choice` to get the true/false values. I think.

plocket: We need to write test cases for these, though I can't think of what they would be atm.

niharikasingh (changes requested):
Need to figure out how to combine old and new tables.

lib/steps.js
plocket: Created a test in the end and it tests some things ¯\_(ツ)_/¯

tests/unit_tests/tables.fixtures.js
```diff
@@ -82,4 +82,283 @@ tables.buttons_event_action = [
 ];
 
 
+// =================
+// Converting from old table to new table
+// =================
+tables.original_formatting = [
```
niharikasingh, line 88: The amount of repetition in this test setup makes me slightly nervous that copy+paste errors while writing the test could result in spurious test failures. Is it possible to write tests where the input is just the original table (as the developer would write it) and it gets compared to the processed version of the new-style table (again, as the developer would write it)?

niharikasingh (approved):

tests/unit_tests/tables.fixtures.js
plocket: That's a complex question. Created #188 for it.

PR #192, opened 2021-04-04
#158 refactor (needs publishing) Handle proxy vars, fix signatures

### LINKED ISSUE #158: Refactor: get all fields on the page then compare to story table

plocket:
#158, I think. This should [almost] finish out the refactor to collect all fields and _then_ match to the story table, adding tests for a lot of functionality.

This particular commit adds handling proxy vars. It also handles signatures, which were missed in previous commits. I would have done them separately, but I'd added them to the testing interview already, having planned to implement tests for signatures before proxy vars (and then apparently didn't). https://github.com/plocket/docassemble-cucumber/runs/2264474918?check_suite_focus=true.

This is not yet published on npm.

Edit:
Not sure if it's more helpful or harmful, but this is based on code mocked up here considering how the code had to be adjusted to work in this repo: https://replit.com/@plocket/index-vars#index.js

Edit:
I'm not sure this handles 'None of the above' yet.

### DIFF
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -11,6 +11,8 @@ let ordinal_to_integer = {
   '6th': 5, '7th': 6, '8th': 7, '9th': 8,'10th': 9,
 };
 
+let proxies_regex = /\b[xijklmnop]\b/g;
+
 module.exports = {
   addToReport: async function( scope, { key, value }) {
     /* Add an item to a specific list in a scenario's report.
@@ -548,20 +550,28 @@ module.exports = {
 
     // Get sought var
     let encoded_sought_var = $($( `input[name="_event"]` )).attr( `value` );
+    // TODO: Change to `sought_name`?
     let sought_var_str = await scope.fromBase64( scope, { base64_str: encoded_sought_var });
-    page_data.sought_var = JSON.parse( sought_var_str );  // Why is this a list? Can multiple vars be sought?
+    page_data.sought_var = JSON.parse( sought_var_str )[0];  // Why is this a list? Can multiple vars be sought?
     // Some `show if` fields have `names` like '_field_n' instead of their var name. Ex: _field_1
     let field_like_names = await scope.getFieldNamesDict( scope, { $: $ });
 
     // Will deal with `option` once inside `select`
     let all_nodes = $( `canvas, fieldset button, .daquestionactionbutton, .form-group input, .form-group select, .form-group textarea` );
+    let proxies_to_names_map = {};
 
-    page_data.fields = [];
+    let temp_fields = [];
     for ( let node of all_nodes ) {
       // Decsion: Do not abstract the below. There's too much data to pass around for it to make sense
 
       let $node = $( node );
-      let field_data = { selector: '', tag: node.name, rows: [], type: '' };
+      let field_data = {
+        node: node,
+        selector: '',
+        tag: node.name,
+        rows: [],
+        type: '',
+      };
       if ( node.attribs.type ) { field_data.type = node.attribs.type; }
 
       // canvas elements are special
@@ -569,11 +579,11 @@ module.exports = {
         field_data.selector = `canvas`;
         field_data.tag = `canvas`;
         field_data.rows = [{
-          var_name: sought_var,
+          var_name: page_data.sought_var,
           value: '/sign',  // Our custom value for signatures
           checked: '',
         }];
-        page_data.fields.push( field_data );
+        temp_fields.push( field_data );
         continue;
       }
 
@@ -582,6 +592,7 @@ module.exports = {
 
       // Often used in the variable name
       let encoded_name = $node.attr( `name` );
+      // Action buttons keep their var name in a unique place
       if ( $( node ).hasClass( 'daquestionactionbutton' ) ) {
         let href = node.attribs.href;
         let match = href.match( /action=([^&]*)/ );
@@ -597,8 +608,8 @@ module.exports = {
       // ====================
       // Default values
       let { var_names, values } = await scope.getPossibleVarNames( scope, {
-          encoded: encoded_name,
-          field_like_names,
+        encoded: encoded_name,
+        field_like_names,
       });
       
       if ( node.name === `select` ) {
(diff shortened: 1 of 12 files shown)


lib/scope.js
```diff
@@ -698,14 +758,19 @@ module.exports = {
...
     // or less than one match, this code will fail later.
     if ( matching_nodes.length === 1 ) { return selector; }
-    // Use `expect` instead. Maybe add link to issues page.
-    else { throw `${ matching_nodes.length } nodes match the '${ selector }' selector. Talk to the developers.`; }
+    // TODO: Use `expect` instead? Maybe add link to issues page. [which issue?]
+    // Submit buttons are a dime a dozen on da pages and aren't used to match variables.
+    // Everything else should be unique.
+    else if ( selector !== '#daquestion button' ) {
+      throw `${ matching_nodes.length } nodes match the '${ selector }' selector. Talk to the developers.`;
+    }
```
plocket, line 773: Signature fixes.

lib/scope.js
```diff
@@ -1035,6 +1219,8 @@ module.exports = {
       await scope.afterStep(scope, { waitForShowIf: true });
     },
     CANVAS: async function ( scope, { handle, set_to }) {
+      // `<canvas>` takes a while to load
+      await scope.page.waitForSelector('#dasigcanvas');
```
plocket, line 1223: Signature stuff.

plocket: Waits here too because of the possibility that this may be called from elsewhere.

lib/scope.js
```diff
@@ -1133,13 +1319,20 @@ module.exports = {
...
     let fields_to_set = [];
     for ( let row of matches ) {
       let field = row.field;
 
+      // Only wait for `<canvas>`. It takes a while to load, but will never be
+      // hidden. Other elements could be hidden, though, and will never show up.
+      if ( field.tag === 'canvas' ) {
+        await scope.page.waitForSelector('#dasigcanvas');
+      }
+
```
plocket, line 1335: Signature stuff. What we're doing here is getting the handle to send to `.setVariable()`. To get the handle for the canvas, we have to wait. I'm debating putting this in `.setVariable()` instead. If we want to start removing puppeteer from this function, that might help. We haven't yet assessed whether that's possible or not, though.

plocket: It just occurred to me this may affect other things that use `.setVariable()`. Shoot. Not enough bandwidth for this...

tests/features/generated_story_tables.feature
```diff
@@ -90,3 +96,4 @@ Scenario: Format of story table is v1
     | textarea | | Multiline text\narea value |
     | show_3 | True | true |
     | show_2 | True | true |
+    |  |  | /sign |
```
plocket, line 99: I guess this should technically have two signatures. The table rows no longer get removed, though, so it's not technically necessary. If people were still going to use old tables, this may or may not be considered a problem. Have to work on the generator to see how it needs to be updated.

lib/scope.js
plocket: I can't find anything else using `.setVariable()`, so we should be safe.

lib/scope.js
```diff
@@ -593,32 +595,42 @@ module.exports = {
...
       // Decsion: Do not abstract the below. There's too much data to pass around for it to make sense
-
       let $node = $( node );
-      let field_data = { selector: '', tag: node.name, rows: [], type: '' };
+      let field_data = {
+        node: node,
+        selector: '',
+        values: [],
+        var_names: [],
+        name_patterns: [],
```
niharikasingh, line 616: Might be useful to document somewhere what these three values mean. The `field_data` structure seems pretty important for our testing framework, so it would be good to have documentation about what it should contain.

lib/scope.js
```diff
@@ -668,19 +681,66 @@ module.exports = {
...
+        // Get the data that will help us construct the intrinsic name later and
+        // look for the sought_var match now.
+        let pattern = await scope.getSoughtVarPatterns( scope, { field_name: name, });
+        field_data.name_patterns.push( pattern );
+        // Try to pull out the intrinsic name parts
+        let maybe_map = await scope.getProxiesToNamesMap( scope, {
+          sought_var: page_data.sought_var, pattern_data: pattern,
+        });
+        // If that works, it's the right map for the proxy values
+        if ( maybe_map ) { proxies_to_names_map = maybe_map; }
```
niharikasingh, line 705: I don't understand the advantage of using `maybe_map` here instead of just assigning directly to `proxies_to_names_map`

lib/scope.js
```diff
@@ -883,6 +992,83 @@ module.exports = {
...
+          rebuilt.push( betweens[ indx/2 ] );
+        } else {
+          rebuilt.push( pattern_data.proxies[ Math.floor( indx/2 ) ]);
+        }
+      }  // end for rebuilding
+      pattern_data.name_parts = rebuilt;
+    }  // ends if has proxy parts (x, i, j, etc.)
+
+    return pattern_data;
+  },  // Ends scope.getSoughtVarPatterns()
```
niharikasingh, line 1046: Are we planning (hypothetically, at some point in the future) to write unit tests for the different functions in this file? This seems like a prime candidate for a lot of unit testing.

lib/scope.js
plocket: #194!

lib/scope.js
plocket: Good question! #196! Though having trouble of thinking of a succinct way to document it.

lib/scope.js
plocket: That question has been on my mind a lot too. I made #197 for it.

PR #201, opened 2021-04-05
Too many fixes. Fixes #190, but other bugs found. Outlined in CHANGELOG.

### LINKED ISSUE #190: New race condition: url changes, but page is not a new page
Currently we're checking for the url to change. Why is that? Because we had an issue where puppeteer thought the next 'page' had loaded, but it was still the same page. (This begs the question of whether we should even use `.waitForNavigation()`, but maybe that's a different issue.)

Now we're running across the url changing while the same page is loaded (or reloaded?) before it move on to the next page. We need something unique to detect a true change in a page.

Current suggestion is the sought var in the `_events` hidden input field `value`. That _should_ be unique, though we cannot be sure of that.

Note: We do not have a clear way to reproduce this bug, but it did happen consistently [ in [DBD2](https://github.com/caroRob/docassemble-DeadBrokeDads2/tree/add_tests_to_separating_interview_code_etc) ].

Note: We thought of using this to detect an infinite screen (a da-specific bug) - when the sought var doesn't change, but the url changes, it may be an 'infinite screen'. That won't work exactly because of the case we just ran into. Maybe detecting multiple consecutive page changes with the same sought var would be sufficient.

plocket:
Fixes #190 and other items listed in CHANGELOG.md. We can go through this together if needed. In trying to get DBD2 to pass, I'm running into bugs and needed behavior changes in the new code. It's the perfect interview with lots of weird edge cases!

I thought it would just a couple things and lost track. There's more to do, but I'm making this PR before this gets bigger. I've added a couple unit tests to cover some of the fixes and I'm making issues on the testing interview repo for adding fields to test against a live interview.

### DIFF
--- a/lib/steps.js
+++ b/lib/steps.js
@@ -711,6 +711,10 @@ When(/I set the address of ?(?:the var(?:iable)?)? "([^"]+)" to "([^"]+)"/, asyn
 
 After(async function(scenario) {
   if (scenario.result.status !== 'passed') {
+    if ( env_vars.DEBUG ) {
+      console.log( 'Error occurred while running test' );
+      await scope.page.waitFor( 60 * 1000 );
+    }
     await scope.addToReport(scope, { key: 'ids', value: '~ scenario stopped here ~' });
     await scope.page.screenshot({ path: `error${ scope.safe_id }.jpg`, type: 'jpeg', fullPage: true });
   }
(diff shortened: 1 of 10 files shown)


lib/steps.js
```diff
@@ -711,6 +711,10 @@ When(/I set the address of ?(?:the var(?:iable)?)? "([^"]+)" to "([^"]+)"/, asyn
 
 After(async function(scenario) {
   if (scenario.result.status !== 'passed') {
+    if ( env_vars.DEBUG ) {
+      console.log( 'Error occurred while running test' );
+      await scope.page.waitFor( 60 * 1000 );
```
niharikasingh, line 716: What's the reason behind waiting _after_ an error happens?

niharikasingh (approved):

lib/steps.js
plocket: Good question. It's so the developer (developing locally) has time to examine the page before it closes. I'll add that comment in the next PR.

PR #203, opened 2021-04-06
Fix incorrect truthy test, some small cleanup (sorry)

plocket:
Adds unit tests as well. Turns out our map was overriding itself, exactly as we had been trying to avoid...

### DIFF
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -712,7 +712,9 @@ module.exports = {
           sought_var: page_data.sought_var, pattern_data: pattern,
         });
         // If that works, it's the right map for the proxy values
-        if ( maybe_map ) { proxies_to_names_map = maybe_map; }
+        if ( Object.keys( maybe_map ).length ) {
+          proxies_to_names_map = maybe_map;
+        }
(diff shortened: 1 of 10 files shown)


lib/scope.js
```diff
@@ -712,7 +712,9 @@ module.exports = {
           sought_var: page_data.sought_var, pattern_data: pattern,
         });
         // If that works, it's the right map for the proxy values
-        if ( maybe_map ) { proxies_to_names_map = maybe_map; }
+        if ( Object.keys( maybe_map ).length ) {
+          proxies_to_names_map = maybe_map;
```
niharikasingh, line 716: Do we expect to have valid `maybe_map` values multiple times in this loop? If not, we can break here, since this code makes sure only one value of `maybe_map` gets written to `proxies_to_names_map`.

niharikasingh (approved):

lib/scope.js
plocket: The loop is needed for setting up the nodes, so we can't break. Regarding the possibility of multiple values, I've made #206.

PR #205, opened 2021-04-07
Use a unique project name

### LINKED ISSUE #20: Prevent failure from project already existing (e.g. two pushes close together)
Add a random suffix to the project name. Rationale: If you make two pushes really close together, it shouldn't be a reason for your tests to fail.

niharikasingh:
Closes #20 

This is what it looks like when I create multiple projects with the same branch:

Screen Shot 2021-04-06 at 8 29 54 PM

### DIFF
--- a/lib/github_integration/puppeteer-utils.js
+++ b/lib/github_integration/puppeteer-utils.js
@@ -31,18 +31,26 @@ const navigateToManageProject = async (page) => {
 
 const createProject = async (page) => {
   // Check if a project with this name already exists
+  // If it does, come up with a new project name
   await navigateToManageProject(page);
-  const projectLinkSelector = `[href="/playground?project=${env_vars.PROJECT_NAME}"]`;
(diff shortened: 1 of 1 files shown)


plocket (approved):
Thanks! This will definitely give a unique name.

A nice future tweak would be to increment the count as if it were counting up. That is, right now it does:
1. `proj_name`
2. `proj_name0`
3. `proj_name01`
4. `proj_name012`
5. `proj_name0123`

and so on. It might be a little neater to have `proj_name1`, `proj_name2`, etc.

It's happening because we're using `env_vars.PROJECT_NAME` as the base even though it's constantly changing. What we'd need is some kind of `base_name` variable that will hold on to `proj_name` so we can add to that instead of adding repeatedly to the end of the most recent `env_vars.PROJECT_NAME`.
```js
env_vars.PROJECT_NAME = base_name + nameIncrementer;
```

PR #213, opened 2021-04-12
Close #207, move test files to da Sources for new projects

### LINKED ISSUE #207: Store test files in da `Sources` directory instead
That way developers can edit them when they're editing the interview and their tests don't necessarily have to fail.  [This will only affect newly created projects, of course. Past projects can either update or keep what they have.] [~Allow storing test files in each location.~]

Research: Can da handle `.feature` files? I may be misremembering, but experiments in https://github.com/plocket/docassemble-ALAutomatedTestingTests may have shown it does not allow proper editing and/or proper saving. [Answer: It now allows editing of those files.]

Will add [a sibling issue](https://github.com/plocket/docassemble-ALAutomatedTestingTests/issues/102) for this in https://github.com/plocket/docassemble-ALAutomatedTestingTests [for the test-setup interview].

plocket:
Close #207. This will allow developers to update tests while editing in the da Playground instead of locally or in github.

Sorry for the extra comments added.


niharikasingh (approved):
Just to confirm, tests will now exclusively be stored in the `sources` directory, correct? The issue suggests that people can store them in both places (`sources` AND `tests`) but the `package.json` scripts only refer to the former.

plocket:
Yes, just `sources`. I forgot to change the issue. This will only affect newly created projects, of course. Past projects can either update or keep what they have.

plocket:
I've just noticed, our workflow doesn't save the test file to the right location anymore. Should we just delete that workflow at this point since we have the new interview we can do it with? That might be best - less duplication.

niharikasingh (approved):

PR #221, opened 2021-05-05
Fix mix of proxy vars and non-proxy vars not finding proxy var

plocket:
I'm a bit concerned that the regex here is too broad, but I'm not sure how else to approach this problem. I also can't think of a situation where it would get us into trouble, but I'm almost certain it's out there.

On a page with proxy vars, usually all variables use the proxy var in them. When a developer uses `code:` to generate fields on a page, the code may create fields that substitute the actual variable as opposed to using the proxy variable.

In the below code, this results in a field with the decoded variable value of `children[1].name.first`. Our old proxy pattern finder would be looking for something of the format `children[i].name.first` and would not find it. We somehow need to match `children[1].name.first` (the sought var) and `children[i].birthdate`, etc.

Created unit tests for it.

Example question:
```yml
sets: 
  - children[i].name.first
  - children[i].name.last
comment: |
  Changed code to list only first names of children in question, but then choices for children_of_both in q. id: custody_children became their full names instead of only first names even after I added '.familiar()' to the end.
# `ordinal()` says 'first', even if there's only one :(
id: children user is parent of
question: |
  Your ${ children.as_noun()}
subquestion: |  
  % if children.target_number == 1:
  Tell us about your child
  % else:
  % if i == 0:
  Tell us about your oldest child first.
  % else:
  You told us
  about ${comma_and_list([child.familiar() for child in children.complete_elements()])}. 
  Tell us about your ${ordinal(i)} child. 
  % endif
  % endif
     
  % if i == 0:
  % if children_cares_for.target_number >= 1:
  For now, just name your own ${ children.as_noun()}.
  % endif
  % endif
 
fields:
  - code: |
      children[i].name_fields()
  - Birthdate: children[i].birthdate
    datatype: BirthDate
    min: ${ today().minus(years=18) }
    max: ${ today() }
    validation messages:
      min: |
        The court asks only for children under 18.
  - Is ${other_parties.familiar()} this child's other parent?: children[i].defendant_is_parent
    datatype: yesnoradio
    show if:
      code: |
        defendant_and_plaintiff_are_parents_of_child # means parents of any children
validation code: |
  if children[i].birthdate < today().minus(years=18):
    validation_error("The court asks only for children under 18", field="children[i].birthdate")
How question came to be asked
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,7 +38,9 @@ Format:
 ### Security
 - 
 -->
-<!-- ## [Unreleased] -->
+## [Unreleased]
+### Fixed
+- Does not handle proxy vars when fields produced by code are included on a page and they skip the proxy var (using the actual value instead). That is, there is a mix of the use of proxy vars and non-proxy vars both referring to the same objects.
 
 ## [1.3.5] - 2021-05-04
 ### Added
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -1036,6 +1036,7 @@ module.exports = {
       name_parts: [ field_name ],  // Entire name by default
     };
 
+    // This uses a word break to make sure we get an isolated x, i, j, ...
     if ( proxies_regex.test( field_name )) {
       pattern_data.uses_proxies = true;
       /* substitute, placeholder, proxy, envoy, mediary, wildcard, gap var, catcher, ~catch all~, reserve, generic, space */
@@ -1044,9 +1045,20 @@ module.exports = {
 
       // The regex that will discover and capture the actual var names
       let escaped = [];
-      for ( let tween of betweens ) { escaped.push(
-        tween.replace(/[\-\[\]{}()*+?.,\\\^$|#\s]/g, "\\$&")
-      );}
+      for ( let tween of betweens ) {
+        // Save things like brackets, but replace every other group with .*
+        // Example need:
+        // Sought var is `people[1].name.first`
+        // Because the question is using `code:`, the sought var is there as that exact match: `people[1].name.first`
+        // A second field is using `i`, though: `people[i].birthdate`
+        // Matching only to 'people[', 'i', '.name.first' will not work as there's no such text on the page
+        // Here we have to be able to discover 'i' using any field that may
+        // contain 'i', no matter what the other text surrounds it. Will this be too permissive?
+        let sans_periods = tween.replace(/\./g, "");
+        let safe = sans_periods.replace(/[\-\[\]{}()*+?.,\\\^$|#\s]/g, "\\$&")
+        let alpha_replaced = safe.replace(/[A-Za-z0-9_]+/g, ".*");
+        escaped.push( alpha_replaced );
+      }
       let full_regex = '^' + escaped.join( '(.+)' ) + '$';
       pattern_data.names_regex = new RegExp( full_regex );
 
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "docassemble-cucumber",
-  "version": "1.3.5",
+  "version": "1.3.5-fx-proxy-mix.1",
   "description": "Integrated automated end-to-end testing with docassemble, puppeteer, and cucumber.",
   "main": "lib/index.js",
   "scripts": {
--- a/tests/unit_tests/getMatchingRow.test.js
+++ b/tests/unit_tests/getMatchingRow.test.js
@@ -83,6 +83,15 @@ it(`matches the right table and field rows for a multi-proxy name (x[i])`, async
 it(`matches the right table and field rows for a proxy name when a non-match comes after a match`, async function() {
   let result = await getMatchingRow( scope, { page_data: page_data.proxies_non_match, story_table: tables.proxies_non_match });
   expect( result ).to.deep.equal( matches.proxies_non_match );
+});
+
+// Proxy vars used on a page with a mixture of fields created by
+// `code:` and normal fields
+// - code: children[i].name_fields()
+// - Birthdate: children[i].birthdate
+it(`matches the right table and field rows for a proxy name when using normal fields and those created with \`code\` `, async function() {
+  let result = await getMatchingRow( scope, { page_data: page_data.proxies_code_fields_mix, story_table: tables.proxies_code_fields_mix });
+  expect( result ).to.deep.equal( matches.proxies_code_fields_mix );
   // console.log( JSON.stringify( result ) );
 });
 
--- a/tests/unit_tests/getPageData.test.js
+++ b/tests/unit_tests/getPageData.test.js
@@ -20,6 +20,11 @@ const getPageData = scope.getPageData;
 * NOTE: scope.getField should only be concerned with the stuff in `#daquestion`
 */
 
+before(function () {
+  scope.report = {};
+  scope.safe_id = 'unit_tests';
+});
+
 // ============================
 // Standard fields - no proxies
 // ============================
@@ -76,7 +81,6 @@ it(`creates the right data for an event action button`, async function() {
 // ============================
 // x[i].name.first
 it(`creates the right data for a multi-proxy name (x[i])`, async function() {
-  // `field:` and `action buttons:`
   let result = await getPageData( scope, { html: html.proxies_xi });
   expect( result ).to.deep.equal( page_data.proxies_xi );
 });
@@ -84,9 +88,17 @@ it(`creates the right data for a multi-proxy name (x[i])`, async function() {
 // your_past_benefits[i].still_receiving
 // your_past_benefits['State Veterans Benefits'].still_receiving
 it(`creates the right data for a proxy name when a non-match comes after a match`, async function() {
-  // `field:` and `action buttons:`
   let result = await getPageData( scope, { html: html.proxies_non_match });
   expect( result ).to.deep.equal( page_data.proxies_non_match );
+});
+
+// Proxy vars used on a page with a mixture of fields created by
+// `code:` and normal fields
+// - code: children[i].name_fields()
+// - Birthdate: children[i].birthdate
+it(`creates the right data for a proxy name when using normal fields and those created with \`code\` `, async function() {
+  let result = await getPageData( scope, { html: html.proxies_code_fields_mix });
+  expect( result ).to.deep.equal( page_data.proxies_code_fields_mix );
 });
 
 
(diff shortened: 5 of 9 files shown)


plocket:
This won't work yet. I'll add notes later.

plocket:
I finally got some space to think about some possible scenarios and how to possibly match them:

| proxy var | possible way to match |
|-|-|
| `foo[i].bar[j]` | `foo[i].bar[j]anything`? |
| `foo[i].bar[j].zoo` | `foo[i].bar[j].anything`? |
| `foo[i].bar[j][1]` | `foo[i].bar[j][anything`? |
| `foo[i].bar[j]['zoo']` | `foo[i].bar[j]['anything`? |
| `x.foo.bar` | `x.anything` |
| `x['foo']` | `x['anything` |
| `x[0]` | `x[anything` |
| `x` | `anything` but with no punctuation |

The current solution works for `foo[i]` type situations, but for `x` situations it does not. `x.foo` regex would come out to `/^(.+).*$/`. That's not anything. We need some delimiters somehow. We can't keep all periods or the scenario outlined in this PR in general won't work.

One thought is to match everything other than the last group exactly (remember groups are split by the proxy vars themselves):
``` js
// foo[i].bar[j].zoo
/^foo\[(.+)\].bar\[(.+)\].*$/
// x.foo.baz
/^(.+)\..*$/
```

That wouldn't work if both `foo[i].bar[j].zoo` and `foo[i].chair.hair[j]` could be on the page at the same time. With that possibility, maybe something that could work is an exact match to the first group with any immediately following punctuation:
``` js
// foo[i].bar[j].zoo
/^foo\[(.+).*(.+).*$/
// x.foo.baz
/^(.+)\..*$/
```

Maybe keep brackets around `i`, `j`, etc. type proxy vars. Maybe this can be restated as keeping punctuation that directly surrounds any proxy vars.
``` js
// foo[i].bar[j].zoo
/^foo\[(.+).*\[(.+)\].*$/
// x.foo.baz
/^(.+)\..*$/
```

Do any of these work if the sought var is `x` and the `code:` specifier creates `foo` instead? Maybe we need to use a \b instead of any kind of punctuation? Maybe we use word break with `x` and brackets with everything else...?

Would appreciate thoughts on this.

niharikasingh:
I'm not sure I understand the problem fully. 

The initial PR was fixing the problem where the page would have the variable `children[1].name.first` but we were only looking for `children[i].name.first`. What is the corresponding equivalent for the `x.foo` case? Will the page have `x.foo` but we are currently looking for `y.foo`?

What happens if our regex that looks for `x.foo` is over-permissive and finds too many potential matches on the page? Is that necessarily a bad thing? For example, our code will think there are 5 fields on the page while in reality, there are only 3. It'll fill out the fields that the user provided and click to the next page. The extra fields should just get treated like non-required fields, right? It's not perfect, but the tests should still work if I'm thinking about this correctly.

plocket:
You're getting close to what I'm trying to describe, so let me try to re-word. Btw, the current PR works fine in the case of `i`, but not for `x`. It's not an easy process to understand and I've found it hard to articulate, but I'll try to walk through the process with an example situation.

The relevant YAML of an interview that would work fine:
``` yml
generic object: Individual
fields:
  - "Hair color": x.hair_color
  - "Number of toes": x.num_toes
```

The story table is:
```
| user.hair_color | Blue |  |
| user.num_toes | 6 |  |
```

Our logic
1. Find out the sought var is `user.hair_color`
1. Match `x.hair_color` to `user.hair_color` (with `.hair_color`) to discover that wherever `x` appears alone, it represents `user`
2. Use that info to create regex that then converts all variables that use `x.whatever` to `user.whatever`
3. Now we have a way to match the variable names in the table to the fields on the page.

So. That's the summary of the process. Now on to where this breaks and the solution in this PR.

---

The YAML:
```
  - code: |
    x.hair_color_field
  - "Number of toes": x.num_toes
```

This represents the variables as they would appear on the page once the code has been run to generate the page.
```
  - "Hair color": user.hair_color
  - "Number of toes": x.num_toes
```

Note that instead of `x.hair_color`, the page now has `user.hair_color`.

The sought variable is still `user.hair_color` and the story table is still
```
| user.hair_color | Blue |  |
| user.num_toes | 6 |  |
```

Outcome before this PR:

No proxy variable would be identified. `.hair_color` would match a variable, but it would match exactly. There's no `x.hair_color` so no proxy variable would be detected. There would be no regex created and `x.num_toes` would literally stay `x.num_toes` and would not match anything in the table.

Outcome after this PR:

The regex to identify `x` is much more permissive, but I think it's fine. It would detect that `x` is present and where it is. The _conversion_ regex is _way_ too permissive, though (`/^(.+).*$/`). It works fine for `i`, but not for `x`. With the `x` examples above I think it would turn every field variable name into `user.num_toes` or something like that, which means that every field would be set to `6`.

---

I'm hoping to find a way to create conversion regex that can handle most, if not all, combinations of `x`, `i`, `j`, etc. that may crop up.

Does that clarify the situation a bit? If not, that's ok.

plocket:
Quick reminder that we're no longer trying to match table variables to DOM nodes. We're instead collecting all fields and recording the variable names of each field. We then later try to match those discovered variable names to names that appear in the story table.

niharikasingh:
I like the idea of "an exact match to the first group with any immediately following punctuation." It seems like the simplest way to solve this issue to me, without adding additional issues. However, we need to make sure that the first group will include `x` in the `x.foo.bar` case. Would it make sense to look at the two cases as follows?

1. `foo[i].bar[j].zoo` case: where the proxy variable **is not** the first character in the string. (Alternatively, the condition could be that the proxy variable is surrounded by brackets, or is not touching a period.) In this case, we can replace the proxy variable with `(.+)`, just like you did right now, and keep anything that precedes the first proxy variable. 

1. `x.foo.bar`: where the proxy variable **is** the first character in the string. In this case, we want to keep the proxy variable around in the regex, or we risk making the regex too permissive. So the regex will be something like `/^x\..*$/`. 

I'm not sure if this condition will cover all possible cases... If the "is/is not first character in string" condition doesn't cover all possible cases, would any of the other two conditions I mention in the first point?

plocket:
Unfortunately `/^x\..*$/` will only match things that start with `x.` literally. To put it another way, `x` is a stand-in for another string - the actual variable name. There will be no row in the story table that will look like `| x.foo | something |  |`. It will be something like `| children.foo | something |  |`.

plocket:
I also worry about the use of `i` on a page like to set both `foo[i]` and `bar[i]`, so I'm not sure we can keep the first characters exact either. Ugh.

plocket:
~Thinking about it now, I think we do need to keep even more punctuation. Turning `foo[i].bar[j]` into `/^foo\[(.+).*(.+).*$/` would not work as it would again capture too much.~

~I think I am not currently able to make something that will be able to differentiate every eventuality. It may even not be possible for a human to do so if there's too much variety. I think we're going to have to settle on some things that we just don't cover. I'm not sure what those are, though.~

Possible lead for using the API (which would take more testing setup, but not sure how to avoid that):

Using the GET request described at https://docassemble.org/docs/api.html#interview_data you can get you objects like:
```
      {
        "hide": true, 
        "to_insert": "children[0].name.first", 
        "var": "children[0].name.first", 
        "var_base": "children[i].name.first", 
        "var_type": "possibly_not_used"
      },
```

Could it solve all the problems ever involved in proxy variables? That would be freakin' amazing. We could throw away all the current proxy var code. Well, maybe not all of it. Ok, I don't know which bits we could throw away, but there would be changes and there's hope!

plocket:
Looking at the creation of `var_base`, I'm no longer sure we can use it. The file isn't loading on github for me right now, but the link to that line would be
```
            for var in sorted(names_used):
                var_trans = re.sub(r'\[[0-9]\]', '[i]', var)
                var_trans = re.sub(r'\[i\](.*)\[i\](.*)\[i\]', r'[i]\1[j]\2[k]', var_trans)
                var_trans = re.sub(r'\[i\](.*)\[i\]', r'[i]\1[j]', var_trans)
                info = dict(var=var, to_insert=var)
                if var_trans != var:
                    info['var_base'] = var_trans
```

Let me know if I'm wrong, but that looks to me like it's replacing every digit in brackets with an index variable regardless of whether the question uses an index variable or not. It also seems to have a couple other things off about it. This would require some pretty extensive testing to work out what the outcomes would be under various circumstances.

If that's the case, I'm not sure we can use this to determine the evaluated value of a proxy variable.

plocket:
`interview_data` is a dead end. It's meant for the sidebar in the playground and isn't useful for runtime stuff. Back to the drawing board.

plocket:
I think I described the problem poorly beforehand. I may have been confused. I will try to restate:

I think the problems start when we try to find the actual variable name. That is, the name of the variable that is in the same position as the index variable.

For example, using `x.hair_color` to find `user.whatever`. Creating regex to find `user` is the part that is tripping us up.

plocket:
Maybe the below will help a bit. I'm not sure whether the second scenario would ever happen, but it seems like it could.

#### Possible I think
Given the strings `users['foo'].hair_color` and `x[i].toes.count`, match `x` to `users` and `i` to `'foo'`.

#### Impossible?
Given the strings `users['foo'].hair_color`,  `parents[0].height`, and `x[i].toes.count`, match `x` to `users` and `i` to `'foo'`.

There are other scenarios that may be harder to figure out. We may need to sit down and list out as many scenarios as we can come up with.

plocket:
We could possibly try to match both scenarios. We already do that a bit when dealing with encrypted names. We could add both `parents[0].toes.count` and `users['foo'].toes.count` and just see if either of them appear on the table. That could run into problems, of course, if both of those are listed in the same table with different values.

plocket:
Latest idea is to add a column to the story table, probably to write the sought var - that can be used as the unique identifier for the page.

plocket:
I'm not sure I wrote a clear description of the fail case above. This is the problem:

This works:
```
sought var: parents[0].child.foo
fields:
x.foo
x.bar
```
`x` is clearly `parents[0].child`

This does not work
```
sought var: parents[0].child.foo
fields:
parents[0].child.foo
x.bar
```
What is `x`? `parents[0].`? `parents[0].child`?

plocket:
I’m debating two paths for setting development priorities:
1. Developing this new v2/v3 table where you include the sought var
2. Improving and testing the other sentences we have (the non-story table stuff) so that developers can use those in the meantime when all else fails. They can still use story tables, they may just have to be in pieces, interrupted by other sentences that care about the order of questions.

The downside of opt. 1 is that it’s a breaking change and I’m also not sure how long it will take to work out the details. I anticipate creating lots of new bugs.

If folks have input, I’d appreciate that.

P.S. I'm not sure quite what will happen with this PR at the moment. It may be possible to improve the regex somewhat for people that are sticking with this version of the framework.

niharikasingh:
Based on the above, sounds like option 2 will be the easier path. Both paths seem equal to me from the test writer's perspective (both involve some small amount of pain), so I'm in favor of the easier path.

plocket:
That's the way I'm leaning too.

plocket:
See #256 for the path we took.

PR #252, opened 2021-06-07
Close #241, close #251, interactive steps 'set' and 'sign'. Also,

### LINKED ISSUE #241: Test or change the Step 'I set'
`When(/I set the ?(?:"([^"]+)" choice of)? var(?:iable)? "([^"]+)" to "([^"]+)"/, async ( choice_name, var_name, set_to ) => {`

I think this is going to have to use the sought var as well, so this may either come later or be a good place to start testing that out.

### LINKED ISSUE #251: Test the Step 'I sign'
`Then('I sign', async () => {`

plocket:
fix 'sign' step as I found it to be broken. Need to note that none
of the tests currently test the action button in the testing
interview.

[After these test are written, we can convert this step to use the new field setting methodology that the table is using.]

### DIFF
--- a/tests/features/interactive_steps.feature
+++ b/tests/features/interactive_steps.feature
@@ -1,0 +1,61 @@
+@interactive
+Feature: Interactive steps
+
+# In tag names, 'i' is for 'interactive'
+
+@generated @fast @i1
+Scenario: I set text-type values
+  Given I start the interview at "all_tests"
+  And I set the "True" choice of var "checkboxes_yesno" to "true"
+  And I set the "checkbox_other_opt_1" choice of var "checkboxes_other" to "true"
+  And I set the var "dropdown_test" to "dropdown_opt_2"
+  And I set the "False" choice of var "radio_yesno" to "true"
+  And I set the "radio_other_opt_3" choice of var "radio_other" to "true"
+  And I set the var "text_input" to "Regular text input field value"
+  And I set the var "textarea" to "Multiline text\narea value"
+  When I set the "True" choice of var "direct_standard_fields" to "true"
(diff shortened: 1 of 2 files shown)


tests/features/interactive_steps.feature
```diff
@@ -0,0 +1,61 @@
...
+  And I set the "True" choice of var "checkboxes_yesno" to "true"
+  And I set the "checkbox_other_opt_1" choice of var "checkboxes_other" to "true"
+  And I set the var "dropdown_test" to "dropdown_opt_2"
+  And I set the "False" choice of var "radio_yesno" to "true"
+  And I set the "radio_other_opt_3" choice of var "radio_other" to "true"
+  And I set the var "text_input" to "Regular text input field value"
+  And I set the var "textarea" to "Multiline text\narea value"
+  When I set the "True" choice of var "direct_standard_fields" to "true"
+  # Next page
+  Then the question id should be "showifs"
```
niharikasingh, line 18: Is this a good place to re-test the "then I arrive at the next page" statement? I know we have it in another test, but we could test it twice since it's a trickier statement.

niharikasingh (approved):

tests/features/interactive_steps.feature
plocket: It's a good thought to keep in mind, but I don't think repeating it here will serve a purpose because this doesn't trigger one of the tricky situations so we'd basically be testing the same thing. We haven't yet been able to replicate the tricky situations that have come up against. If/when we can, I think those would be useful tests. I'd appreciate your thoughts on this, though, if you have others.

Issue #235, opened 2021-06-06
Test or change the Step 'link opens in'?


plocket:
`Then(/the "([^"]+)" link opens in (a new window|the same window)/, async (linkText, which_window) => {`

Is this really a useful test? Really only for escape buttons, but maybe that's enough reason. Also, should it really use the visible text to identify the link? Would it be better or worse to use the url?

Issue #234, opened 2021-06-06
Test or remove Step `link leads to`


plocket:
`Then(/the "([^"]+)" link leads to "([^"]+)"/, async (linkText, expected_url) => {`

Uses the text on the page. Is it important that certain text match up to a certain URL? Is it more important that the url itself is present on the page? See #233 too.

plocket:
Closed in favor of #236

PR #255, opened 2021-06-08
Add language-agnostic link-url checking step and its test

plocket:
Addresses #234 and #235. I realize this adds a step, but when we make sure the other two link checking steps aren't being used, I think it'd be best to remove them and stick with this one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@ Format:
 -->
 ## [Unreleased]
 ### Added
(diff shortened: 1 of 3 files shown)


niharikasingh:
I think the new step is good, but to close the issues mentioned above, we should either remove those steps or test them. Otherwise, this MR doesn't close those issues.

niharikasingh (approved):

PR #258, opened 2021-06-11
Close #256, use new way to identify sought var and table vals

### LINKED ISSUE #256: V2: more dev input about the sought var for each page so stories can work
This is the basic problem:

This works:
```
sought var: parents[0].child.foo
fields:
x.foo
x.bar
```
`x` is clearly `parents[0].child`

This does not work
```
sought var: parents[0].child.foo
fields:
parents[0].child.foo
x.bar
```
What is `x`? `parents[0].`? `parents[0].child`? da does not tell us this in the DOM.

The only way we can think of getting around this is to use the sought var, the one unique value on the page. If the developer lets us know what the sought var is, as well as the variable name that actually appears on the page (which is more complicated than you might think, but we'll go into that another time), then we can know which row is truly for which page.

Makes it much harder on the developer, and means that the generator can't really generate a fully working test by itself, but it is what it is.

[See PR #221 for a long and confusing discussion]

plocket:
We can talk through this one.

Close #256


tests/features/generated_story_tables.feature
```diff
@@ -29,73 +33,69 @@ The below scenario covers story table tests for:
...
-    | showif_text_input | Show if text input value |  |
-    | showif_textarea | Show if\nmultiline text\narea value |  |
-    | showif_yesnoradio | True | true |
-    | text_input | Regular text input field value |  |
-    | textarea | Multiline text\narea value |  |
-    | show_3 | True | true |
-    | show_2 | True | true |
-    | signature_1 | /sign |  |
-    | signature_2 | /sign |  |
+    | var | value | checked | sought |
```
plocket, line 41: New `sought` column (should we call this `trigger`?)

tests/features/generated_story_tables.feature
```diff
@@ -29,73 +33,69 @@ The below scenario covers story table tests for:
...
+    | direct_showifs | True | true |  |
```
plocket, line 42: `sought` column does not need a value so that simple interviews don't have to deal with it. They don't even need a `sought` column`.

tests/features/generated_story_tables.feature
```diff
@@ -29,73 +33,69 @@ The below scenario covers story table tests for:
...
+    | button_continue | True | true |  |
+    | buttons_other | button_2 |  |  |
+    | buttons_yesnomaybe | True | true |  |
+    | checkboxes_other | checkbox_other_opt_1 | true |  |
+    | checkboxes_other | checkbox_other_opt_2 | true |  |
+    | checkboxes_other | checkbox_other_opt_3 | false |  |
+    | checkboxes_yesno | True | true |  |
+    | direct_standard_fields | True | true |  |
+    | dropdown_test | dropdown_opt_2 |  |  |
+    | x[i].name.first | Proxyname1 |  | proxy_list[0].name.first |
```
plocket, line 52: `x[i].name.first` is the string in the YAML for this variable name, and it's the info that will be in the DOM. 
https://github.com/plocket/docassemble-ALAutomatedTestingTests/blob/main/docassemble/ALAutomatedTestingTests/data/questions/all_tests.yml#L80

tests/features/generated_story_tables.feature
```diff
@@ -29,73 +33,69 @@ The below scenario covers story table tests for:
...
-    |  |  | /sign |
+  Then I arrive at the next page
+
+# TODO: Test og table with `sought` col added?
+
+@fast @stf3
+Scenario: Table is format 2 (no sought var)
+  Given I start the interview at "all_tests"
+  And the user gets to "showifs" with this data:
+    | var | value | checked |
```
plocket, line 93: The main columns right now are `| var | value | checked |`. In this new version, I want to combine `| value |` and `| checked |`. I think it's possible. I think the wording needs to be worked on. I think this is the next priority so that we don't have to do this yet again after a bunch of interviews start using this version in case we find that it's a breaking change.

plocket: The key is checkboxes. Right now, you say 'this checkbox option is true (checked)' (or false for unchecked). The shift would be that the developer would only list the checkboxes that are checked and we'd make sure the other, unnamed, checkboxes were unchecked. This is a bit of a different paradigm, but one we can do now that we collect all the elements on the page. I'm not completely sure it's possible.

tests/unit_tests/html.fixtures.js
```diff
@@ -535,8 +574,12 @@ html.proxies_non_match = `
 // ============================
 // Signature
 // ============================
+// I think signature pages won't have sought var node on mobile
```
plocket, line 577: TODO: I need to remove this comment. It's solved as long as the interview developer does the thing that needs doing.

On the other hand, maybe we need to handle situations where the sought var is not present on the page to allow for more simple interviews to still be tested.

lib/scope.js
```diff
@@ -435,22 +435,7 @@ module.exports = {
...
-    // No repos we know of has used the array version of this table as of 03/27/21,
-    // so we shouldn't need to support the old formatted tables in here.
-    let sample = var_data[0];
-    if ( Array.isArray( sample )) {
-      let rows = raw_var_data.raw();
-      var_data = rows.map(( row ) => {
-        return { var: row[0], value: row[1], checked: row[2] };
-      });
-    }
+    * Table must have headers. Arrays are not supported. */
```
plocket, line 438: I made an issue for properly implementing supporting arrays, but as no one's using them right now, I don't feel it's urgent.

lib/scope.js
```diff
@@ -518,30 +506,30 @@ module.exports = {
...
 
-    // All these pieces of data feel very divorced from the field itself and the action happening.
-    // Surely there's a better way.
-
     // Get sought var
-    let encoded_sought_var = $($( `input[name="_event"]` )).attr( `value` );
-    // TODO: Change to `sought_name`?
-    let sought_var_str = await scope.fromBase64( scope, { base64_str: encoded_sought_var });
-    page_data.sought_var = JSON.parse( sought_var_str )[0];  // Why is this a list? Can multiple vars be sought?
+    let encoded_sought_var = $($( `#sought_variable` )).data( `variable` );  // This should come out to 'None' at the very least
```
plocket, line 510: This is the new sought var element that developers will have to add to their interviews. AL will have them added automatically unless someone overrides the values there.

lib/scope.js
```diff
@@ -518,30 +506,30 @@ module.exports = {
...
+    // TODO: Change to `sought_name` or `trigger`?
+    page_data.sought_var = await scope.fromBase64( scope, { base64_str: encoded_sought_var });
+    if ( !page_data.sought_var ) {
+      // TODO: Add documentation and then link to it in this message.
+      await scope.addToReport(scope, { key: 'ids', value: `Warning: You are missing an element containing the data for the page's sought/trigger variable. Some fields, like loops with index variables, may not work correctly in these tests. Talk to the folks at the AssemblyLine project to learn more.` });
```
plocket, line 515: This gets added to the report for every page, but I'm not sweating the small stuff right now.

lib/scope.js
plocket: This is how the developer can do it in their interview. If done this way (there are others, but this is one), this element should be on mobile as well. Not all methods do that.
https://github.com/plocket/docassemble-ALAutomatedTestingTests/blob/main/docassemble/ALAutomatedTestingTests/data/questions/all_tests.yml#L6-L10

lib/scope.js
```diff
@@ -552,17 +540,16 @@ module.exports = {
...
-          field_name: page_data.sought_var,
-        });
-        field_data.name_patterns = [ canvas_pattern ];
-        temp_fields.push( field_data );
+        field_data.rows = [{
+          var: page_data.sought_var,
+          value: '/sign',  // Our custom value for signatures. TODO: discuss
+          checked: '',
+        }];
+        page_data.fields.push( field_data );
```
plocket, line 548: This is for the signature field/canvas. Just restoring it to almost what it was before.

lib/scope.js
```diff
@@ -605,72 +592,26 @@ module.exports = {
...
-        let pattern = await scope.getSoughtVarPatterns( scope, { field_name: name, });
-        field_data.name_patterns.push( pattern );
-        // Try to pull out the intrinsic name parts
-        let maybe_map = await scope.getProxiesToNamesMap( scope, {
-          sought_var: page_data.sought_var, pattern_data: pattern,
-        });
-        // If that works, it's the right map for the proxy values
-        if ( Object.keys( maybe_map ).length ) {
-          proxies_to_names_map = maybe_map;
+        if ( proxies_regex.test( name )) {
```
plocket, line 606: Will find `x`, `i`, `j`, etc - the proxy vars. If any field has a proxy var, the whole page data will store that.

plocket: From @niharikasingh 
> Do we need to add tests to make sure all appearances of proxy vars are caught?

From me:
I think we can. I think we may not need to because the regex finds an isolated letter of the type described, so it's pretty simple. It can fail if someone uses `foo.i` and we can make that more robust, but that's bad practice. I think they can't use `x` for other purposes in da because `x` is special. The also can't use `[i]`, `[j]`, etc because those are special. This is very stringent and I think it's one thing we really can count on.

plocket: Add to proxy var regex - check for brackets.
^ to make an issue with.

lib/scope.js
```diff
@@ -1040,12 +882,11 @@ module.exports = {
     */
     // There should only ever be one match for any table row and for any field row.
     // Not sure how to check on this.
-    let matches = [];
+    let matches_by_field = {};
```
plocket, line 885: This will be an object where each field has a key and its value will be an array of matching table rows.

lib/scope.js
```diff
@@ -1056,7 +897,8 @@ module.exports = {
             case 'canvas':
               if ( names_and_values_match ) { rows_match = true; }
               // Support v1 tables
-              else if ( table_row.checked === `/sign` ) { rows_match = true; }
+              else if ( table_row.checked === `/sign` ) { rows_match = true; }  // TODO: Add unit and feature tests for this
```
plocket, line 900: Just adding support for old tables back in.

lib/scope.js
```diff
@@ -1086,15 +928,57 @@ module.exports = {
...
+
             continue;
           }
 
         }  // ends for each table row
       }  // ends for each field row possibility
     }  // ends for each field
 
-    return matches;
+    // Some fields will have multiple matches from the story table.
```
plocket, line 946: The only reason we would care about multiple rows is if there is a possible proxy var conflict.

lib/scope.js
```diff
@@ -1086,15 +928,57 @@ module.exports = {
...
+    // has a sought var described that matches the page sought var, the
+    // program knows to use that row to set the value.
+
+    // For now, we're being flexible about it - if we don't think we
+    // need that sought var because the situation is not as described
+    // above, we don't prevent data getting through.
+    // e.g. If an interview has a list of `users`, but they only set
+    // data for `users[0]`, there's no need to check the sought var.
+    // That way devs with simple cases can ignore the sought var column.
+    de_duped_matches = [];
```
plocket, line 963: TODO: Need to add `let` here. I do not want a global var.

lib/scope.js
```diff
@@ -1086,15 +928,57 @@ module.exports = {
...
+    for ( let field_key in matches_by_field  ) {
+      let field_matches = matches_by_field[ field_key ];
+      if ( page_data.uses_proxies && field_matches.length > 1 ) {
+        // If multiple story table rows match one field, only
+        // keep the one(s) that match this particular page (sought var)
+        for ( let match of field_matches ) {
+          if ( match.table.sought === page_data.sought_var ) {
+            de_duped_matches.push( match );
```
plocket, line 971: TODO: Add a message to the dev if this leave us with multiple matches.

lib/scope.js
```diff
@@ -1086,15 +928,57 @@ module.exports = {
...
+          }
+        }
+      } else {
+        // If proxy vars aren't relevant here or if only
+        // one table row matched this field, use that match obj
+        de_duped_matches.push( field_matches[0] );
```
plocket, line 977: Should we add all matches here? Not just the first one? I don't know when that would happen in legit code, but maybe it could and we're more likely to see that at some point if we don't hide it here. I can definitely see a dev accidentally adding double rows. We should therefore `concat()` and report if there are multiple matches. This will also make things more transparent to us, so it's an awesome idea.

lib/scope.js
```diff
@@ -605,72 +592,26 @@ module.exports = {
...
+        // TODO: Add links to docs about da YAML values that devs never fill in themselves and
+        // so don't know what they are. E.g. yesnomabye -> True/False/None.
         // TODO: Text inputs and textareas must ignore the value, etc. How?
         values.push( node.attribs.value );
       }
-      // TODO: add default of empty string for values?
 
-      field_data.var_names = var_names;
-      field_data.values = values || [];
+      // ******************
```
plocket, line 601: TODO: Remove old dev line

niharikasingh (approved):

tests/unit_tests/html.fixtures.js
plocket: It seems I've already handled this. See the `.addToReport()` a bit further on.

plocket: Comment adjusted.

lib/scope.js
plocket: First comment added, along with link in warning message. Doesn't talk about mobile, but put that as something to add to docs in #259.

lib/scope.js
plocket: Made #260

lib/scope.js
plocket: Comment added. #261 made for brackets regex.

lib/scope.js
plocket: Comment made.

lib/scope.js
plocket: Comment added.

lib/scope.js
plocket: Made #262.

lib/scope.js
plocket: Made #262

Issue #264, opened 2021-06-12
v2: Drop support for format 1 and 2 tables. Retain support for missing 'sought' column.


plocket:
Format 1 table: `| var | choice | value |`
Format 2 table: `| var | value | checked |`
Format 3 table: `| var | value | sought |` or `| var | value |`

Format 1 tables had `| var | choice | value |`. For both radio buttons and checkboxes, it was appropriate to give `choice` a value of 'false'. With checkboxes, that meant the checkbox would be unchecked. With radio buttons, it was not necessary but it was sometimes there because it was easier than remembering how its rules were different.

The same is true of format 2 tables for the 'checked' column.

In format 3, checkboxes with 'false' are removed from the table all together. That should not happen to radios, but there's no way for the code to tell the difference. We could require all format 1 tables to update themselves to remove those type of radio columns, but we could just as easily tell everyone to update to format 3.

This does truly put us into framework v2 territory, though.

plocket:
See #242 for discussion on this kind of format where unmentioned things are made 'empty'.

plocket:
Framework 2 achieved today. We're using a column called `trigger`.

Issue #242, opened 2021-06-06
Can we reduce checkbox descriptors (to shorten sentences/the table)


plocket:
That is, just have the user put in the variable name and:
- for multiple choice, list what should be checked
- for yes/no and no/yes, specify if the box should be checked or not

That would reduce sentences in some ways, though checkboxes may need their own sentence at that point. I'm not sure we could use 'I set the ___ var to ___' because you can't simply set a checkbox to true/false because of no/yes checkboxes. We could do 'I set the ___ var to checked/unchecked' I suppose, but that doesn't seem very clear. We could try it, though.

Maybe we change the sentence to 'I set the ___ field to ___'. Would that work?

plocket:
I think we can. We might use 'checked/unchecked' specifically, so there isn't confusion about true/false with [noyes fields](https://docassemble.org/docs/fields.html#yesno).

plocket:
For multiple choice checkboxes, the dev would list values that need to be checked. Unlisted values would be unchecked if needed. We'd need to make the table sentence somehow more clear about how it represents the 'positive' state of the page - these would be the only values on the page. Not sure what that sentence can look like.

plocket:
Trying to figure out how to do field/story table matching now. Here's a simplified explanation of the conundrum and a vague idea of a solution. Maybe we should add this to some docs if this becomes implemented.

We have a form with fields.

I start the logic by collecting all the fields on the form
Each field has `id`s that may be encoded, sometimes multiple times. They also might not be encoded. (it’s more complicated than `id`s, but I think that’s sufficient for now). To be sure to get at least one correct decoded field `id`, I make a list with three versions of a field (not always, but let’s just assume always for now), one for each of three decodings.

I also have data with rows containing human-written `id`s that I want to try to match with the field `id`s. A field `id` might have multiple matching row `id`s, but then I have an additional piece of data that can ensure I get the right row (unless someone put in the wrong data and I do want to log a warning when that happens).

This may not be a problem even if it’s true: I am a little worried that a row `id` might have multiple matching field `id`s, but I don’t know how to break that down and actually figure out for a fact what the deal is. I would want to log that finding as a warning too.

The real problem is that non-matches are significant too. So if I have a field that has three `id`s and one matches and the other two don’t, I don’t want to accidentally use the non-matches when I should be using the match. That is, I need to know which actual fields (not just field `id`s) match which rows in the table and which actual fields match nothing at all.

How do I make sure to not over-match?

[...]

PR #265, opened 2021-06-13
Address #242, reduce to 3 columns

plocket:
This is also a big one. It may need a conversation too. Let me know if you're interested in/have bandwidth for that. Again, a couple things snuck in that didn't need to. I've been bad about that recently.

Related changes:
- Added `sought` column for story table.
- No longer support formats 1 or 2 of tables. Definitely v2 time.
- Removed `checked` column from table. Only put a checkbox in the table if you want it to be checked. Everything not appearing in the table will be given 'empty' values. For checkboxes, that means ensuring they're unselected. For fields, that means emptying. Dropdowns unselect everything. RADIO BUTTONS ARE AN ISSUE because it's not clear what to do there. If there's a default selected, the user wouldn't be able to select nothing.
- Matching of fields to story table rows works differently to line up with the removal of `checked`.
- Updated tests to match new functionality.
- Tried to make element selectors more picky.
- Recognize anchor tag.

Unrelated changes:
- Attempt to handle extra space chars in dropdown menu item values. Should use trim or something instead with everything eventually. See issue.
- Start scenario id with timestamp.
- Allow more languages in scenario name.

I'm still not sure about this 3-column arrangement, but I guess we'll see.

I also don't know what sentence to make for it.

Addresses #242, but not sure it totally closes it. Also addresses #264.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,7 +47,7 @@ Format:
 - `sought` column for story table.
 
 ### Changed
-- Removed `checked` column from table. Now only put a checkbox in the table if you want it to be checked. Everything not appearing in the table will be given 'empty' values. For checkboxes, radios, etc, that means ensuring they're unselected. For fields, that means emptying. Not sure what it means for dropdowns yet.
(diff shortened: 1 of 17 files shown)


plocket:
Could still keep some of this. The matching, for example, is a bit more robust and informative than what we had before. Would just have to change how fields do and don't get set.

plocket:
That might mean this could be broken out into two parts - I can put the matching in by itself.

plocket:
We did decide to go with the other approach.

PR #268, opened 2021-06-14
Make node specifiers more specific, add another debug log.

### LINKED ISSUE #262: Add message to the dev when multiple table rows match a field
Should we add all matches here? Not just the first one? I don't know when that would happen in legit code, but maybe it could and we're more likely to see that at some point if we don't hide it here. I can definitely see a dev accidentally adding double rows. We should therefore `concat()` and report if there are multiple matches. This will also make things more transparent to us, so it's an awesome idea.

_Originally posted by @plocket in https://github.com/plocket/docassemble-cucumber/pull/258#discussion_r650303021_

Also handle it here: https://github.com/plocket/docassemble-cucumber/pull/258#discussion_r650302443

plocket:
Also some cosmetic stuff.

Basically, starts making moves towards PR #265. Taking it a bit slower.

[Sorry, I meant the last changes to apply to a different branch. Let me know if I should split this.]
[Close #262, message to dev about matching rows.]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,8 @@ Format:
 - Tests for some interactive Steps.
 - Tests for Assembly Line package-specific tests.
 - Move debug logs for getMatchingRow and getPageData to appropriate places.
+- Start testing report to make matching tests work.
+- Give developer warning when they have multiple rows in the table for one variable.
 
 ### Changed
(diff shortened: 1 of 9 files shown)


plocket:
I forgot that PR was merging in to here, sorry. May be easier to look at the two first commits in here independently now, @niharikasingh .

niharikasingh (approved):

PR #281, opened 2021-06-18
Close #263, remove use of `scope.getField()`

### LINKED ISSUE #263: Remove use of `getField()`
We now want to use `getPageData()`. It's not a 1-to-1 replacement - logic will have to be changed up to accommodate the change in functionality - but it's more reliable and tested.

plocket:
Close #263

### DIFF
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -974,7 +974,7 @@ module.exports = {
     return { question_has_id: found_any_question_id, id: id, id_matches: ids_match };
   },  // Ends scope.examinePageID()
 
-  processVar: async function ( scope, { var_data, id, non_table=false }) {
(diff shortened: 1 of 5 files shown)


lib/scope.js
```diff
@@ -1132,18 +972,16 @@ module.exports = {
     return { question_has_id: found_any_question_id, id: id, id_matches: ids_match };
   },  // Ends scope.examinePageID()
 
-  processVar: async function ( scope, { var_data, id }) {
+  processVar: async function ( scope, { var_data, id, non_table=false }) {
```
niharikasingh, line 975: `non_table=false` is a double-negative, which is hard for me to parse at first glance. You could change to `table=true` if that doesn't break other things.

niharikasingh (approved):

PR #284, opened 2021-06-20
Close #220 MVP, support a `.there_is_another` loop, but only with explicit dev values

### LINKED ISSUE #183: Combine the matching logic of elements that have the same checks, e.g. button and canvas
Since button and canvas have the same code, you could combine them

```javascript
case 'canvas': 
case 'button': 
    # something 
```

_Originally posted by @niharikasingh in https://github.com/plocket/docassemble-cucumber/pull/177#discussion_r600081391_

### LINKED ISSUE #220: Story table: How do we handle `.there_is_another` loops?
When asking about cases for example, these are the separate screens that may be shown:

What is docket number for the first case? `cases[0].docket_number` (A: 1234)
Do you have another case? `cases.there_is_another` (A: yes)
What is the docket number for the second case? `cases[1].docket_number` (A: 5678)
Do you have another case? `cases.there_is_another` (A: no)

In the first instance, the answer should be ‘yes’, but in the second instance, the answer should be ‘no’. It’s the same variable both times, though. Is it possible to handle that in a story table somehow?

plocket:
This is just the MVP for supporting this kind of loop, but it'll be sufficient for now. The developer will have to add this row in this way to their table:
```
|  x.there_is_another  | 3 | some_var.there_is_another |
```
This will indicate to make 3 `some_var` items. Literally, it means the test will answer 'yes' to the `.there_is_another` question two times (the question isn't asked the first time).

It's kind of confusing whether to give the number of items you want or the number of button presses you want. I'm not sure which to pick or if there's a better way to indicate it.

Closes #220 MVP

[Also starts to move `sought` to `trigger` - that's a conversion that will happen next/soon.]

Close #183.

### DIFF
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -678,6 +678,9 @@ module.exports = {
           let names_match = table_row.var === guess.var;
           let names_and_values_match = names_match && table_row.value === guess.value;
 
+          // If we were always using tables, we could just use `sought` to match
+          // every field, but we get Steps too, which don't have that info
+
           switch ( field.tag ) {
             case 'canvas':
               if ( names_and_values_match ) { row_matches = true; }
@@ -708,6 +711,14 @@ module.exports = {
             case 'a':
               if ( names_match ) { row_matches = true; }
               break;
+          }
+
+          // --- Special cases ---
+          // `.there_is_another` buttons won't always match `value`
+          if ( table_row.var.includes(`.there_is_another`)
+              && table_row.var.includes(`.there_is_another`)
+              && names_match ) {
+            row_matches = true;
           }
 
           if ( row_matches ) {
@@ -778,7 +789,6 @@ module.exports = {
     // da's header covers the top of the section containing the fields which means
     // sometimes it blocks nodes from clicks. Scroll item to the bottom instead.
     await handle.evaluate(( elem ) => { elem.scrollIntoView(false); return elem.scrollTop });
-    let set_to = row.value;
 
     if ( env_vars.DEBUG && handle ) {
       let name = await handle.evaluate(( elem ) => {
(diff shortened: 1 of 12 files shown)


plocket:
It is failing for some reason. I'll have to run it locally again.

plocket:
For some reason it thought a file didn't exist that does actually exist. No idea what that was about.

plocket:
Unless we update the ALTestingTests interview, we can't run tests on this again. Sorry, I forgot we needed to keep it steady for this. That said, it should be pretty easy to change it back for this and update it again later.

lib/scope.js
```diff
@@ -633,167 +623,213 @@ module.exports = {
...
+
+    // If field uses proxies and there has been more than one match
+    // we now use the trigger var to narrow down the matches.
+    if ( uses_proxies && matches.length > 1 ) {
+      // Multiple matches means this came from a story
+
+      // Feedback for devs - we need the trigger var but we don't have it (field data)
+      // TODO: Add docs and link to docs on how to add the trigger var
+      if ( field.trigger === undefined ) { await scope.addToReport(scope, { key: 'ids', value: `WARNING: A field on this page uses an index variable or a generic object, but the page does not contain an element with the trigger variable's name. Tell the developer to add an element to the page with the id "sought_variable"` }); }
+
```
plocket, line 743: If `field.trigger` is undefined, stop here. This probably wouldn't be expected behavior for the developer for this to continue and not collect matches.

lib/scope.js
```diff
@@ -633,167 +623,213 @@ module.exports = {
...
     }
 
+    let { tag, type } = field;
+    // We use the last match. We wrote a warning for the user if there were
+    // mutliple matches in the story that the last value will be used. In future,
+    // Steps will be able to set this data gradually and we don't want the user
+    // to get confused thinking that an earlier Step was used instead of a later one.
+    // TODO: Add the `was_used` prop to the row so it can be used in the report.
+    let row = matching_input_rows[ matching_input_rows.length - 1 ];
+    let set_to = row.value;
```
plocket, line 826: For `there_is_another`, this is usually a button, so it shouldn't usually matter, but this is not actually always `True` or `False` - the value needs to get mutated or something. And then mutated back??? (from the original row) We want to keep the list of rows referring to the original rows, right? Maybe we don't have to, though, if we collect used rows separately.

plocket: Send back the right row values from `matches`, but from `setVariable` send back the original row that was used...? Is that possible? Maybe the fake row can have a reference to the original row in it?

niharikasingh (approved):

PR #286, opened 2021-06-20
Close #285, switch sought to trigger

### LINKED ISSUE #285: Change `sought`/`sought_var` to `trigger`
I think that will be more intuitive for people. I could be wrong. It also will help with name consistency in the code I think.

plocket:
Close #285


niharikasingh (approved):
This seems like it just renames the `sought` string to `trigger`, right? I don't see any substantive code changes, but let me know if I missed anything.

PR #288, opened 2021-06-20
Allow 'I get to' for the story table Step. Needed for better docs

### LINKED ISSUE #129: Pick 'I' or 'the user' step language and align everything to that
I suggest `I`, but mostly because it's shorter.

[We can grandfather in the previous versions of steps so we don't have to make a new major version for the breaking change, but deprecate it in the CHANGELOG so that in the next major version we can get rid of it.]

plocket:
Closes #129

### DIFF
--- a/tests/features/observation_steps.feature
+++ b/tests/features/observation_steps.feature
@@ -23,7 +23,7 @@ Scenario: I see user errors
 @slow @o3
 Scenario: I check navigation
   Given I start the interview at "all_tests"
-  And the user gets to "showifs" with this data:
+  And I get to "showifs" with this data:
     | var | value | trigger |
     | checkboxes_other['checkbox_other_opt_1'] | true |  |
     | dropdown_test | dropdown_opt_2 | |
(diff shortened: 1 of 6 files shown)


tests/features/observation_steps.feature
```diff
@@ -23,7 +23,7 @@ Scenario: I see user errors
 @slow @o3
 Scenario: I check navigation
   Given I start the interview at "all_tests"
-  And the user gets to "showifs" with this data:
+  And I get to "showifs" with this data:
```
niharikasingh, line 26: Since the tests now support both sentences ("the user gets..." and "I get..."), should one of the tests still retain "the user gets..." language to make sure that is still tested?

niharikasingh (approved):

tests/features/observation_steps.feature
plocket: Good point. It should probably be its own test. Made #292. In fact, we should figure out all the variations possible in these steps and test all of them. Really good point.

PR #290, opened 2021-06-22
Switch how loop incrementation is done.

plocket:
#220. Moves the responsibility into `scope.processVar()`. Let me know what you think.

### DIFF
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -665,7 +665,7 @@ module.exports = {
         // I've found no better way to get this to work than to put it in here
         let table_value = var_datum.value;
         if ( var_datum.var.includes( `.there_is_another` )) {
-          table_value = await scope.getThereIsAnotherValue( scope, { value: table_value, from_table: from_table });
+          table_value = await scope.getThereIsAnotherValue( scope, { row: var_datum, from_table: from_table });
         }
         let names_and_values_match = names_match && table_value === guess.value;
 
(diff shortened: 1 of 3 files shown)


lib/scope.js
```diff
@@ -1125,28 +1119,34 @@ module.exports = {
...
     * safe value to set for this round. Give the developer a warning if
     * it's an inappropriate value. Avoid mutation.
     *
     * WARNING: Remember to decrement the actual value later.
     */
-    if ( parseInt( value ) < 1 || value === `False` || value === `false` ) {  // Loop will end
+    let value = row.value;
+    let int_val = parseInt( value );
+
+    if ( row.times_used >= int_val ) {  // Loop will end
```
niharikasingh, line 1133: In my head, this condition would fit better if it came after the `isNaN` condition. That way the top two conditions (value is a boolean, most likely from non-table test) form one group, and the bottom three (value is an integer, most likely from table) form one group.

lib/scope.js
```diff
@@ -817,11 +814,13 @@ module.exports = {
     let { tag, type } = field;
     // We use the last match. We wrote a warning for the user if there were
     // mutliple matches in the story that the last value will be used. In future,
-    // Steps will be able to set this data gradually and we don't want the user
+    // Steps will be able to set this data gradually and we don't want the developer
     // to get confused thinking that an earlier Step was used instead of a later one.
-    // TODO: Add the `was_used` prop to the row so it can be used in the report.
     let row = matching_input_rows[ matching_input_rows.length - 1 ];
     let set_to = row.value;
+    if ( row.var.includes( `.there_is_another` )) {  // Seemed cleaner to just repeat
```
niharikasingh, line 821: Clever!

niharikasingh (approved):

PR #293, opened 2021-06-23
Uses `.target_number` to handle loops. Close #220.

### LINKED ISSUE #220: Story table: How do we handle `.there_is_another` loops?

plocket:
Closes #220 again. See [description of behavior there](https://github.com/plocket/docassemble-cucumber/issues/220#issuecomment-866418428). If you have thoughts, I'd welcome an opinion on this alternative alongside the opinions described there.

### DIFF
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -665,7 +665,7 @@ module.exports = {
         // I've found no better way to get this to work than to put it in here
         let table_value = var_datum.value;
         if ( var_datum.var.includes( `.there_is_another` )) {
-          table_value = await scope.getThereIsAnotherValue( scope, { value: table_value, from_table: from_table });
+          table_value = await scope.getThereIsAnotherValue( scope, { row: var_datum, from_table: from_table });
         }
         let names_and_values_match = names_match && table_value === guess.value;
 
@@ -752,7 +752,7 @@ module.exports = {
         }
       }
       // Feedback for devs - we are missing the matching triggers we need
-      if ( matches.length === 0 ) { await scope.addToReport(scope, { key: 'ids', value: `WARNING: A field on this page uses an index variable or a generic object. The test gave multiple matches for the story data for the field, but none match the trigger variable. Is the story table missing a row? For now, the test will have to ignore this field. The rows' data is ${ JSON.stringify( matches_for_report )}. The field's data is ${ JSON.stringify( field )}` }); }
+      if ( matches.length === 0 ) { await scope.addToReport(scope, { key: 'ids', value: `WARNING: A field on this page uses an index variable or a generic object. The test gave multiple matches for the story data for the field, but none match the trigger variable. Is the story table missing a row? For now, the test will have to ignore this field. The rows' data is ${ JSON.stringify( original_matches )}. The field's data is ${ JSON.stringify( field )}` }); }
     }
 
     // Dev may have accidentally included a row more than once. Give them feedback
@@ -822,6 +822,8 @@ module.exports = {
     // to get confused thinking that an earlier Step was used instead of a later one.
     // TODO: Add the `was_used` prop to the row so it can be used in the report.
     let row = matching_input_rows[ matching_input_rows.length - 1 ];
+    // TODO: right now this is wrong because it's the original `row.value`. `.there_is_another`
+    // sometimes needs to be an adjusted value. Maybe we could just recalculate that value in here.
     let set_to = row.value;
 
     // Detect a custom type and get the element of that custom type if it exists
@@ -956,6 +958,7 @@ module.exports = {
     // Depending on the current value/status of a checkbox and the desired
     // value to set, either taps the checkbox or leaves it alone.
     let status = await label.evaluate( elem => { return elem.getAttribute('aria-checked'); });
+    // `.there_is_another` won't work for this yet (because `set_to` should not be the original `row.value` sometimes)
     if ( set_to.toLowerCase() !== status ) {
       await label[ scope.activate ]();
     }
@@ -1042,14 +1045,7 @@ module.exports = {
         // Try to set other types of fields
         let row_used = await scope.setVariable( scope, { field, var_data, from_table });
         if ( row_used ) {
-
-          // If the `.there_is_another` field was set and it's an integer
-          // mutate the actual row obj to increment the loop progress
-          if ( row_used.var.includes(`.there_is_another`) && parseInt( row_used.value )) {
-            row_used.value = `${ parseInt( row_used.value ) - 1 }`;
-          }
-
-          row_used.times_used += 1;
+          row_used.times_used += 1;  // Esp important for `.there_is_another`
           process.stdout.write(`\x1b[36m${ '*' }\x1b[0m`);  // assumes var was set if no error occurred
         } else {
           // If the field wasn't ready, keep it around
@@ -1071,13 +1067,8 @@ module.exports = {
 
         let row_used = await scope.setVariable( scope, { field: fields_to_set[ data_i ], var_data, from_table });
         if ( row_used ) {
-          // If the `.there_is_another` field was set and it's an integer
-          // mutate the actual row obj to increment the loop progress
-          if ( row_used.var.includes(`.there_is_another`) && parseInt( row_used.value )) {
-            row_used.value = `${ parseInt( row_used.value ) - 1 }`;
-          }
+          row_used.times_used += 1;  // Esp important for `.there_is_another`
 
-          row_used.times_used += 1;
           // TODO: special continue/navigation button function that handles url and all?
           
           // Have to catch errors right now or may get stuck in an infinite loop?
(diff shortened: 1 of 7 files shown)


plocket:
I actually think this is broken. I'm going over the code again and it seems to be like `.there_is_another` only gets set once, which is wrong.

plocket:
Ah, I forgot that `processVar()` should really be called `processPage()`. It's called on every page, which is where the table values need to change. So it's fine. Whew. Trying to write some tests for it now.

lib/scope.js
```diff
@@ -1052,8 +1030,11 @@ module.exports = {
         }
 
         // Try to set other types of fields
-        let was_set = await scope.setVariable( scope, { field, var_data, from_table });
-        if ( was_set ) {
+        let row_used = await scope.setVariable( scope, { field, var_data: ensured_var_data });
+        if ( row_used ) {
+          // Esp important for `.target_number`
+          if ( row_used.source ) { row_used.source.times_used += 1 }
+          else { row_used.times_used += 1; }
```
niharikasingh, line 1037: Won't all rows with `target_number` as a variable have a source attribute? When will the `else` part of this be used?

lib/scope.js
```diff
@@ -1125,28 +1113,87 @@ module.exports = {
...
 
-    // Invalid value defaults to 'False' with a line added to the report
-    } else if ( isNaN( parseInt( value )) ) {  // Loop will end
-      // console.warn?
-      await scope.addToReport( scope, { key: `id`, value: `You must use a number or \`False\` in an \`.there_is_another\` row. The value was instead ${ row.value }. This test will default to \`False\` to avoid an infinite loop.` });
-      return `False`;
+    // Story table stuff
+    // Target: 3 (3 items. press 'yes' twice.)
+    // Times used: 0 -> `True`, 1 -> `True`, 2 -> `False`, 
+    } else if ( row.times_used >= int_val - 1 ) {  // Story table loop will end
```
niharikasingh, line 1193: Should this check `row.source` as well?

lib/scope.js
```diff
@@ -1125,28 +1113,87 @@ module.exports = {
...
+    if ( value.match( /(false|no|unchecked|uncheck|deselected|deselect)/i )) {
       return `False`;
 
-    } else if ( !from_table && ( value === `True` || value === `true` )) {
-      // This is not a table row, it's a step or something, so we should be safe from infinite loops
-      return `True`;
+    } else if ( value.match( /(true|yes|checked|check|selected|select)/i )) {
+      if ( from_table ) {
+        // Just in case, but this should never happen.
+        await scope.addToReport( scope, { key: `id`, value: `WARNING: The attribute \`.there_is_another\` is invalid in story table tests. Replace it with \`.target_number\` in your \`var\` and \`trigger\` columns. Set the \`value\` to the number of items in that list. This test will now set this row's \`value\` to \`False\`. See https://suffolklitlab.github.io/docassemble-AssemblyLine-documentation/docs/automated_integrated_testing#there_is_another. The row data is\n${ JSON.stringify( row.original )}` });
```
niharikasingh, line 1186: Is this the right error message here? I think someone could reach here even if they don't have `there_is_another` in their table, e.g., `| x.target_number | true | users.target_number |`

Also, the linked documentation seems to indicate that `there_is_another` is valid in the trigger column. This error message says otherwise.

lib/scope.js
```diff
@@ -1125,28 +1113,87 @@ module.exports = {
...
+        return `False`;
+      } else { return `True`; }
 
-    // Invalid value defaults to 'False' with a line added to the report
-    } else if ( isNaN( parseInt( value )) ) {  // Loop will end
```
niharikasingh, line 1144: Do we not worry about non-integer values anymore? I don't see any error handling for those.

lib/scope.js
```diff
@@ -1125,28 +1113,87 @@ module.exports = {
...
+
+      // Add a `.there_is_another` var if it might be needed.
+      // If it's not used in the end, that's fine.
+      // `.target_number` is sometimes a stand-in for `.there_is_another`.
+      // That seems the clearest way to communicate to developers about what they
+      // should write in the test. We will attempt to write more about
+      // the rationale in the documentation at some point.
+      if ( from_table && var_datum.var.match( /\.target_number$/ )) {
+        let special_row = {
+          artificial: true,
```
niharikasingh, line 1135: We aren't using this `artificial` value anywhere right now, that I can tell.

lib/scope.js
```diff
@@ -765,20 +754,20 @@ module.exports = {
...
+  setVariable: async function ( scope, { field, var_data }) {
     /* Returns true if the field was set.
     *
     * TODO: Will the selector ever change the second time around?
     *   If so, we need to store the handles before this.
     * 
     * @returns bool
    */
-    let was_set = false;
+    let row_used = null;
```
niharikasingh, line 765: Hmm... I'm not sure why these changes are showing up in this diff. I thought these were from another MR that's already merged.

niharikasingh:
Based on what I can glean from the Docassemble documentation, `there_is_another` questions are not the same as `target_number` questions. With this update, looks like we're forcing people to choose the latter instead of the former. 

If this change is motivated only by the fact that "people are confused by the number in the value column," then my preferred solution would be to accept both `there_is_another` and `target_number` and clarify the documentation to say that the values for both are the same (in other words, that people should treat both rows similarly when choosing what value to fill in). And if people still get the value wrong, the test should provide immediate feedback by failing or producing output that shows that some rows didn't get used. 

If there are other motivations for this change, then the changes in this MR work too.

lib/scope.js
plocket: This might be one of two questions, so I'll answer both:

If the question is about when `.times_used` will be used:
1. I plan to use it in reports to list the rows there were used vs. the ones that weren't
2. I already don't love the idea of only some rows having some props, like `source`. I'd like to keep them as consistent as possible.

If it's meant literally, this happens with all rows, so this will be used with all non `target_number` rows as well.

Thoughts?

lib/scope.js
plocket: Thanks for spotting this! I think this only gets run right now at the beginning of the `processVar()` loop, so nothing has `source`, but I can see that it would be wise to future-proof it in case things change. I'll do that before we merge this.

lib/scope.js
plocket: re `| x.target_number | true |` - good catch. That's invalid input too. I'm not sure how to handle that exactly. This might be 'good enough for now' + creating an issue for thinking about that. I'd love your thoughts. Made #296.

re the trigger column: Nice catch there too. That definitely can be changed right away.

lib/scope.js
plocket: That was partly me making an assumption that nothing that unsafe could be passed in. I debated about it for a while. You may be right that keeping it in would be a good safety feature. We'd want to give some kind of message to the developer, though. We don't want it to just fail silently. `A ${ row.var } row can't have a value of ${ row.value }`?

lib/scope.js
plocket: That's true. It makes me nervous to have a non-dev generated row around without explicitly saying it. It's a bit of premature future-proofing. Is it worth keeping around or is it confusing?

lib/scope.js
plocket: That's disconcerting. Not sure how to track that down...

plocket:
Thanks for that input. We should talk about it real time as I'm not sure I currently agree and I'd like to hear more and discuss with more nuance. I'll try to hit you up and see what's convenient.

plocket:
Thanks for such a thorough PR, btw!

lib/scope.js
niharikasingh: Ah I didn't realize we wanted to track times used for all rows, not just target_number rows. 👍

lib/scope.js
niharikasingh: In my head, `source` performs the same function, as it only exists for non-dev generated rows. Can we just check for the existence of `row.source`?

niharikasingh: I'm fine with keeping this around too, but maybe with a comment saying it is not currently used but is future-proofing to mark non-dev generated rows.

plocket:
Clarified that developers can still use `there_is_another` in their code and also clarified the documentation that will be pushed for it at https://suffolklitlab.github.io/docassemble-AssemblyLine-documentation/automated_integrated_testing#there_is_another-loop

Issue #141, opened 2021-01-22
Make one, or maybe two, utility function(s) for making safe names


plocket:
I say maybe two because I think most safe-name making functions would not include kanji and that would make it hard for developers to see which reports go with which language. So maybe:
1. An html-safe function
1. One that will still show kanji and other such characters. Maybe we'd say this is a file-name-safe function

Thoughts?

plocket:
- [ ] safe filename
- [ ] safe report name

plocket:
Safe filename is handled. I think safe report name too because I think the same variable is used for both, but I'm not sure. It all currently gets rid of numbers, though, and I'd rather it not do that. That might be another issues. I need to take myself off this for now, but the next thing to do:

- [ ]  check that report names have been made safe

plocket:
I think various names are currently handled appropriately.

PR #295, opened 2021-06-25
Addresses #141, safe names - sanitize filenames

plocket:
#141, but only the filenames part. This has been a while coming I think. Not all developers would know that their scenario's description needs to be safe for a filename. It's also a safe directory name, but I _think_ filenames are safe directory names, right?

Replace invalid chars with one '_' each. There isn't really a more refined way to do it without adding a lot of cruft if we want to respect any duplicate `_`s in the prefix that's passed in.

### DIFF
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -64,7 +64,7 @@ module.exports = {
     if ( lang ) { lang = `${ lang.replace(/_$/, '') }_`; }
     else { lang = ''; }
 
-    let safe_name = safe_filename(`${ prefix }${ lang }${ Date.now() }_${ scope.safe_id }`);
+    let safe_name = safe_filename(`${ prefix }${ lang }${ Date.now() }_${ scope.safe_id }`, { replacement: `_` });
     let safer_name = safe_name.substring(0, (255 - 10));  // Allow room for extensions
     return safer_name;
   },  // Ends scope.getSafeScenarioFilename()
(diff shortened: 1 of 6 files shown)


tests/unit_tests/names.fixtures.js
```diff
@@ -0,0 +1,34 @@
...
+names.undefined_output = ``;
+
+names.colon_input = `invalid:colon`;
+names.colon_output = `invalid_colon_`;
+
+names.slash_input = `invalid/slash`;
+names.slash_output = `invalid_slash_`;
+
+names.two_consecutive_invalid_input = `two_consecutive//invalid`;
+names.two_consecutive_invalid_output = `two_consecutive__invalid_`;
```
niharikasingh, line 31: I love these tests! Very thorough.

lib/scope.js
```diff
@@ -49,6 +50,25 @@ module.exports = {
     return report;
   },  // Ends scope.createFinalReport()
 
+  getSafeScenarioFilename: async function waitForShowIf( scope, { prefix='' }) {
```
niharikasingh, line 53: Does this function really need to wait for show ifs? I think this should say `getSafeScenarioFilename` instead. 

Also, I don't think this function needs to be async either. Nothing in it seems to require waiting for screens/requests.

niharikasingh (approved):

lib/scope.js
plocket: 1. `waitForShowIf` isn't even running - it's defining a function name and is a horrible place for that to be. Thank you for spotting it!
2. Many functions in here don't need to be async, but remembering which are and which aren't was causing me a headache, so I decided all `scope` functions would be async. I'm willing to discuss other perspectives, though.

PR #302, opened 2021-07-03
Allow filenames to end with .yml

### LINKED ISSUE #287: Allow devs to include `.yml` with their filename if they want to
Currently allowed: `Given I start the interview at "whatever"`

Want to also allow: `Given I start the interview at "whatever.yml"

This will be in `scope.load()`

niharikasingh:
Closes #287 

Not sure I wrote the test correctly or put it in the right place.


plocket (approved):
It does the things and passes the tests, so it looks good to me! We'll worry about minor improvements after.

PR #306, opened 2021-07-03
Remove deprecated reference to Buffer, unused function

### LINKED ISSUE #167: Adjust use of Buffer - deprecation warning
Got this warning:
```
(node:9846) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
```

Found buffer [where we're converting to base64 in scope.js](https://github.com/plocket/docassemble-cucumber/blob/main/lib/scope.js#L226).

niharikasingh:
Closes #167 

The ticket asked for the `Buffer` usage to be rewritten, but we have a nice new function that works properly so I think we can just use that instead. Let me know if that's incorrect.

### DIFF
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -318,7 +318,7 @@ module.exports = {
     if ( !encoded_trigger_var ) {
       // TODO: Add documentation and then link to it in this message.
       await scope.addToReport(scope, { key: 'ids', value: `Warning: You are missing an element containing the data for the page's trigger variable. Some fields, like loops with index variables, may not work correctly in these tests. Talk to the folks at the AssemblyLine project to learn more or see an example here: https://github.com/plocket/docassemble-ALAutomatedTestingTests/blob/ad1232049d7ca7a9b97b22f8ca3915dd3dae8114/docassemble/ALAutomatedTestingTests/data/questions/all_tests.yml#L6-L10.` });
-      encoded_trigger_var = encoded_trigger_var || await scope.toBase64( scope, { utf8_str: `None` });
+      encoded_trigger_var = await scope.toBase64( scope, { utf8_str: `None` });
     }
     let trigger = await scope.fromBase64( scope, { base64_str: encoded_trigger_var });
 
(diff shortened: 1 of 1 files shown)


lib/scope.js
```diff
@@ -608,9 +608,15 @@ module.exports = {
...
+  toBase64: async function ( scope, { utf8_str }, exclude_padding=false) {
+    // Sometimes, converting to base64 results in padding characters 
+    // being added to the end. Docassemble does not use these padding
+    // characters everywhere, so we allow for stripping them using 
+    // the exclude_padding argument. For more details, see
+    // https://github.com/nodejs/node/issues/6107#issuecomment-207177131
     let buffer = Buffer.from( utf8_str, 'utf-8' );
-    return buffer.toString( `base64` );
+    let base64str = buffer.toString( `base64` )
+    return exclude_padding ? base64str.replace(/[^A-Za-z0-9]/g, '') : base64str;
```
plocket, line 619: Does docassemble use these padding characters anywhere? Maybe the default here should be `true`?

niharikasingh: works for me. i wasn't sure if docassemble ever kept the padding characters around.

niharikasingh:
Line 321 deserves special attention. Is `encoded_trigger_var = encoded_trigger_var || await scope.toBase64( scope, { utf8_str: `None` });` correct? Or should padding me included there?

plocket:
> Is encoded_trigger_var = encoded_trigger_var || await scope.toBase64( scope, { utf8_str: None }); correct?

It honestly doesn't matter if it's included or not as it's a dummy value. Maybe I should make that more clear, so we can make that an issue.

Unrelatedly, I think I should have deleted `encoded_trigger_var ||` considering this can only happen `if ( !encoded_trigger_var )`, so that's good to know too.

lib/scope.js
plocket: Sorry I didn't catch this before. It occurred to me that I may have seen characters outside of those in base64 strings in the past: https://stackoverflow.com/a/13195218/14144258. I think the safest thing to do might be to just remove any `=` in there? That would probably be `/=/g`.

plocket (approved):
Thank you! It'll be nice not to get that deprecation warning on every single run lol

Issue #138, opened 2021-01-22
Reports: Improve & Discuss what else, if anything, would be useful to report about variables


plocket:
For example:
1. Variables that were set on each page
1. Variables that remain after the target id screen has been reached

Both might be useful to the developer in understanding what's going on with their interview. They could also be used to refine generated tests to make them faster and more clear.

plocket:
Note: With the new approach to solving #79 and #115, we have to make sure to get rid of the placeholder text before putting a variable row in the report.

plocket:
A doc describing various options: https://docs.google.com/document/d/1gvSPUoTbF6wxCgMdpNgVYzq9cgFjQ-X8gdUig5iEUuY/edit#

This option seems strong: https://docs.google.com/document/d/1gvSPUoTbF6wxCgMdpNgVYzq9cgFjQ-X8gdUig5iEUuY/edit#bookmark=id.d6tnivk6kfvr. The initial report would contain a chronological list of the pages visited and the variables used. And, now, probably warnings that came at those various times. Also, if the scenario was a var table, at the very end add a var table made from rows that were used. I think it may also be useful to add a list of the table rows that weren't used.

The proposal:
Chronological list of
1. Scenario name and data
2. page id + sought var
2. field set? row set? with value
3. warnings
2. For each story table the scenario had:
   1. a story table of all the vars that were used (chronological? alphabetical?)
   3. the rows that weren't used (alphabetical?)

plocket:
See #103 for questions about future use of `Background`

plocket:
Report Questions
1. [alphabetical] End of table report to be chronological? No, alphabetical is recommended, so if they will copy/paste, it should conform to that.
1. Original row or actual row? (dev will not understand actual row).
1. Fix? Page continuation symbol printed before error is thrown.
1. Each block of the report is pretty complicated. e.g. each story table has its own summary
1. No divisions between table sections? Change that?
1. With defined buttons, next page id is put in report before the row data of the page on which it was set
1. Fix double escaping Step string new lines.

Report notes (to document?)
1. Does show the page id of the last question before error thrown.

PR #310, opened 2021-07-06
Re: #138, Add variable information to reports

plocket:
Address #138 (may close it too, upon discussion). Notes and caveats also covered in #138:
- End of table report is chronological so that people can copy/
paste it into their files if desired - alphabetical listing of
vars is recommended. (most useful for used vars)
- Problem with double char escaping (e.g. \\n) with Steps, but
avoids accidental new lines from story table strings.
- While running tests, page continuation symbol is printed in
during-test output, even if an error is about to be thrown for that
page. (problem?)
- All vars are displayed as table rows, even if a Step was used.
- Report shows original row, not the rows actually used in the
table. The concern is that it will be confusing otherwise, even
though it then gives less insight into what may have happened wrong
so we'll have to try to count on the warnings we've set up.
- The report is pretty complicated to read when multiple story
tables are used in one scenario.

Added tests for reports, but for now they need a visual check by us - there's no automated snapshot.

### DIFF
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -69,7 +69,10 @@ module.exports = {
     /* Return a string genereated from test scenario data. */
     let [scenario_id, items] = scenario;
     let report = `\n---------------\nScenario: ${ scenario_id.replace(/_/g, ` `) }\n---------------\n`;
-    for ( let item of items ) { report += `${ item.value }\n`; }
+    for ( let item of items ) {
+      if ( item.type === `page_id` ) { report += `screen id: `; }
+      report += `${ item.value }\n`;
+    }
     return report;
   },  // Ends scope.getPrintableScenario()
 
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -27,12 +27,12 @@ module.exports = {
     * keys should correspond to a list containing the relevant data. */
     if ( env_vars.DEBUG ) { console.log(`Report: type: ${ type }, value: ${ value }`) }
 
-    if ( !scope.report ) { scope.report = {}; }
+    if ( !scope.report ) { scope.report = new Map(); }
     if ( !scope.scenario_id ) { scope.scenario_id = `_report_${ Date.now() }` }
-    if ( !scope.report[ scope.scenario_id ] ) {
-      scope.report[ scope.scenario_id ] = [];
+    if ( !scope.report.get( scope.scenario_id )) {
+      scope.report.set( scope.scenario_id, [] )
     }
-    scope.report[ scope.scenario_id ].push({ type, value });
+    scope.report.get( scope.scenario_id ).push({ type, value });
 
     return scope.report[ scope.scenario_id ];
   },  // Ends scope.addToReport()
@@ -40,13 +40,12 @@ module.exports = {
   getPrintableReport: async function( scope ) {
     /* Return a string genereated from test report data. Can/should this be extensible? */
     let report = `Assembly Line Kiln Automated Testing Report - ${ (new Date()).toUTCString() }\n\n`;
-    for ( let safe_id in scope.report ) {
+    for ( let [scenario_id, items] of scope.report ) {
 
-      report += `=== Scenario ${ safe_id } ===\nQuestion ids appeared in this order:\n`;
-      for ( let data of scope.report[ safe_id ] ) {
-        report += `- ${ data.value }\n`
-      }
+      report += `=== Scenario ${ scenario_id } ===\n`;
+      for ( let item of items ) { report += `- ${ item.value }\n` }
       report += `\n`;
+    
     }  // ends for every scenario
 
     return report;
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -37,19 +37,39 @@ module.exports = {
     return scope.report[ scope.scenario_id ];
   },  // Ends scope.addToReport()
 
+  convertToOriginalStoryTableRow: async function( scope, { row_data }) {
+    /* Returns the original data of row into a string formated as a story table row.
+    * Should this show the actual row data that was used instead? That could
+    * be pretty confusing to most devs. */
+    let source_row = row_data.source || row_data;
+    let original = source_row.original;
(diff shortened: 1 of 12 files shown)


lib/scope.js
```diff
@@ -20,40 +20,56 @@ let escape_regex = function( to_escape ) {
...
+    for ( let [scenario_id, items] of scope.report ) {
 
-      report += `=== Scenario ${ safe_id } ===\nQuestion ids appeared in this order:\n`;
-      for ( let q_id of scope.report[ safe_id ].ids ) {
-        report += `- ${ q_id }\n`
-      }
+      // Might convert to markdown in the future
+      report += `---------------\nScenario: ${ scenario_id.replace(/_/g, ` `) }\n---------------\n`;
+
+      for ( let item of items ) { report += `${ item.value }\n`; }
```
niharikasingh, line 67: Hmm... should this print something like `${ item.type.toUpperCase() }: ${ item.value }\n`?

That way, we don't have to type `{ type: `warning`, value: `WARNING: You are missing...` every time we want to specify this is a warning etc.

niharikasingh (approved):

lib/scope.js
plocket: It would be nice to avoid that repetition. I think the only one that would work with that right now is `warning`, though, so I'm not sure that abstraction works well for this. We might think about instead making functionality or a function that specifically adds a warning to the report?

tests/features/reports.feature
```diff
@@ -0,0 +1,54 @@
+@reports
+Feature: Reports show the right things
+
+Note: For now we'll have to check them visually
```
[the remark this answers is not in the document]
plocket: Hmm, I personally may find them a bit hard to describe. Let me check how hard it would be to create those tests so that we don't even have to think about it.

lib/steps.js
```diff
@@ -227,6 +227,45 @@ When(/^(?:the user gets)?(?:I get)? to ?(?:the)?(?: ?question)?(?: id)? "([^"]+)
 
   }  // ends while !id_matches
   // Can anything unsatisfactory get through here?
+
+  // Add report strings from supported table, not an altered one
+  for ( let row of supported_table ) {
+    row.report_str = `${ await scope.convertToOriginalStoryTableRow(scope, { row_data: row }) }`;
+  }
+  // Sort into aphabetical order, as is recommended for writing story tables
```
BryceStevenWilley, line 235: I understand that alphabetical is how the story tables are recommended to be written, but I don't think that's useful for looking at a test report: the rows should be chronological, so it matches the overall chronological structure of the report (and matches the previously shown screen ids).

BryceStevenWilley (commented):
(Most of this from the Backend code call we just had) From the user's perspective, all of the necessary information is there, it just needs a bit more formatting to make it more readable.

Some quick suggestions:
* adding `page id:` or `screen id`: before each id
* adding `set the following variables:` and then the table on the next row, so it's clear that the table is what happened, as opposed to what was _supposed_ to happen on that screen.
* Extra indention on the `Rows that were used` section.
```
page id:
  | randvar | x | | 
page id two:
  | randvar2 | y | | 

    Rows that were set...
       | | |...
```
IMO the indention makes it clear which page ids the rows belong to.

* you mentioned that if it fails, you print `Interview Stopped before it expected`. IMO you should just straight say "Scenario Failed", in bold red text (however much formatting you can do in the terminal), so it's clear that's where the issue happened. I'd probably have more feedback on the errors, but I'd need to see a fail state to see what it looks like. Maybe that'd be a good addition to the `tests/features/reports.feature` testing file :smile:

lib/steps.js
plocket: I hear that and thank you for the feedback!

Regarding order of rows, I didn't quite get to go into that during the discussion, so I'll try to do so here.

First, and least significant: the generator generates a bunch of extra rows because it doesn't know how to tell which variables are or aren't necessary. One hope for this was that it could help people who are using the generator so that they can simplify their table.

Second, and more compelling for me: When you're looking for a specific row to see where it appears, which is what I imagine these tables being useful for, it's hard to remember which row appeared where in chronological order, especially if you're changing the order of screens a lot as some of our developers do. If the rows are in alphabetical order, it would be easier to locate any one specific row. It's the same reason it's recommended for the tables in general. I've found it useful, at least. It probably depends on the type of problems one runs into the most.

I think I had another reason, but I can't recall right now.

plocket: Also, not exactly sure yet how to handle catching errors in this environment where steps are fragmented in such a way. Would love some eyes on that.

plocket:
Just as an fyi, many of our warnings currently contain mostly the data that these new reports are now providing (plus a little extra), but in a less friendly way. We can remove that confusing stuff after we implement these new reports. For some people, the extra info would be useful, but I'm not sure how to include that without confusing everyone else.

tests/features/reports.feature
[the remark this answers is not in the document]
plocket: I added a couple report tests for passing Scenarios. It just seemed too hard to describe the desired output.

plocket:
Would love folks to give the reports in the action another look with some adjustments, @niharikasingh , @BryceStevenWilley.

Things not yet handled/needing further discussion that may wait for a future issue:
- Order of rows in the 'Rows that were set' tables.
- Testing reports for failing scenarios.

lib/scope.js
plocket: #314. We've got more of these now, so it won't be so special.

lib/steps.js
plocket: Made a new PR for an experiment on testing failing tests. I _think_ it's what we're looking for.

tests/features/reports.feature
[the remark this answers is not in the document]
niharikasingh: Ah, yup, that's what I meant. Sorry that was unclear.

niharikasingh (approved):

PR #318, opened 2021-07-08
Close #317, Add custom timeout Step. Also shortens default timeout

### LINKED ISSUE #317: Step: Allow dev to set a custom global timeout
Let's start with just a global custom timeout step. I'll make an issue just for that.

_Originally posted by @plocket in https://github.com/plocket/docassemble-cucumber/issues/126#issuecomment-876396403_

plocket:
Shortens it because now devs who need more time can add it. It
makes an assumption about how long most interviews will need and
we'll have to see if that bears out. Default timeout is now 30secs

~It does lengthen our own tests by a chunk, unfortunately.~ [Actually, it shortens them by a bit in some ways because the default wait time is cut down!]

Closes #317.

### DIFF
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -157,7 +157,7 @@ module.exports = {
     *    the global permitted timeout */
 
     // Stop/resolve if too much time has passed
-    let expirationTime = 1.9 * 60 * 1000;  // m * s * ms
+    let expirationTime = scope.timeout - 200;  // ms
     if ( !endTime ) { endTime = new Date((new Date()).getTime() + expirationTime); }
     let diff = endTime.getTime() - Date.now();
     if ( diff <= 0 ) { return false; }
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -1256,6 +1256,8 @@ module.exports = {
     let was_found = false;
     for ( let error_key in possible_errors ) {
       if ( possible_errors[ error_key ][0] ) {
+        // If you're debugging and you close the window early, the process will still
+        // take this long to exit after the tests finish. Not sure how to account for that.
         if ( env_vars.DEBUG ) { await scope.page.waitFor( 60 * 1000 ); }
         was_found = true;
         error_handlers[ error_key ] = possible_errors[ error_key ][0];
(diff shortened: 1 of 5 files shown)


plocket:
I think for some reason this breaks custom timeouts or something. Maybe just when debugging. I keep getting a hang at the end where the test ends, but the process doesn't exit for a bit. Again, maybe just when debugging, which means something wrong with a debugging timeout.

lib/scope.js
```diff
@@ -178,7 +178,7 @@ module.exports = {
     *    the global permitted timeout */
 
     // Stop/resolve if too much time has passed
-    let expirationTime = 1.9 * 60 * 1000;  // m * s * ms
+    let expirationTime = scope.timeout - 200;  // ms
```
niharikasingh, line 181: Hmm... why subtract 200ms here? I can't figure it out.

lib/steps.js
```diff
@@ -43,15 +43,17 @@ to detect that.
 regex thoughts: https://stackoverflow.com/questions/171480/regex-grabbing-values-between-quotation-marks
 */
 
-setDefaultTimeout( 120 * 1000 );
+// Hoping that most load and download times will be under 30 seconds.
+// TODO: Check what average load and download times are in our current interviews.
+setDefaultTimeout( 30 * 1000 );
```
niharikasingh, line 48: At some point, it would be nice to encode the `30` into a variable (something like `DEFAULT_GLOBAL_TIMEOUT`) so we don't have to repeat it in multiple places.

niharikasingh (approved):

lib/scope.js
plocket: Good question. I'm actually not entirely sure. Weird. I remember thinking about this shortening again as I edited and deciding to it was needed, but I can't currently remember why. Maybe it was to make up for the time it took to get to this function? So that any error/warning message we put in here would be sure to get logged before the outer timeout triggered? If so, 200ms is a lot.

plocket: #321

lib/steps.js
plocket: #322. I thought I planned to do that at some point and then thought there was a barrier. Something about not being able to get external context values into cucumber functions. I now think I hallucinated that, so we should definitely try again.

plocket:
Decided the DEBUG timeout will stick around. If we're debugging, we can use ctrl+c. Added comment as such and will merge when I next have time after the tests have passed.

PR #329, opened 2021-07-13
Add in warning about one download only

### LINKED ISSUE #45: Add warning to download step - only one file can be handled currently
add back in?

_Originally posted by @niharikasingh in https://github.com/plocket/docassemble-cucumber/pull/33#discussion_r530051983_

If some link or button initiates downloading multiple separate files (I assume that's possible???) then the current code may not be able to account for that.

niharikasingh:
Closes #45


plocket (commented):
Is it one download per Scenario, or one download per Step? That is, one call to this particular Step and the click on this particular element.

niharikasingh:
It used to be one per scenario, but I can't figure out why we had that limitation, and if it still exists. Do you remember the reasoning for the original warning? If not, I can run some tests.

plocket:
I'm not sure. Maybe it's because we don't ensure that the filenames are unique in the downloads folder? That said, the name of each downloads folder is now unique per scenario, so it should only be a problem if they download a file of the same name in one scenario.

Also, I think the downloads functionality wouldn't handle a case where one click downloaded multiple files. That would be a once-per-step issue. Not sure if that's even possible, though.

niharikasingh:
Check what happens if we click download twice on the same page, resulting in the same file getting downloaded twice. Does the name conflict cause issues?

plocket:
Dang, the downloads folder only has one download in it. Also, it looks like there's only one downloads folder for everything. If multiple scenarios download files, will there still only be one downloads folder?

niharikasingh:
Ah darn, I thought multiple downloads in the same scenario worked locally, but I think I see what's happening now. Chrome downloads the file twice, but because it has the same name, it only saves once in the downloads folder. I think what we need to somehow be able to specify `I download docassemble-filename.pdf as local-filename.pdf`. That can be a later ticket. For now, I'll fix the warning to clarify.

niharikasingh:
Multiple scenarios with downloads work, see here: https://github.com/plocket/docassemble-cucumber/actions/runs/1063493418

niharikasingh:
😓 i had trim trailing whitespace turned on. oops. i can undo that if you want.

tests/features/interactive_steps.feature
```diff
@@ -75,10 +75,82 @@ Scenario: I set text-type values
...
+  Then the question id should be "first signature"
+  When I sign
+  And I tap to continue
+  # Next page
+  Then the question id should be "second signature"
+  When I sign
+  And I tap to continue
+  # Next page
+  Then the question id should be "simple doc"
+  When I download "simple-doc.pdf"
```
plocket: I think there are supposed to be two of these lines, right?

plocket: Oh! Is this supposed to be proof that downloading in two separate scenarios works ok?

plocket: Note to make future issue: We might want to add something like this to our tests so that this can be confirmed as well.

plocket (approved):
Looking good I think! No worries about whitespace. It's clear enough what the significant change is. Maybe we can add the details of our discoveries to an issue somewhere and note to add it to the documentation.

tests/features/interactive_steps.feature
niharikasingh: Yeah, this was to test whether downloading in two separate scenarios works okay (it does!).

I previously did have two downloads in a single scenario, but that doesn't work yet, so I took it out for now. I made an issue for fixing this later: #357

PR #334, opened 2021-07-15
Make sure only strings are base64ed

### LINKED ISSUE #44: Change base64 encoder fallback string to be more clear and descriptive
change string to something like "will never match existing dom node"

_Originally posted by @niharikasingh in https://github.com/plocket/docassemble-cucumber/pull/33#discussion_r530051120_

niharikasingh:
Closes #44

### DIFF
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -625,7 +625,7 @@ module.exports = {
 
   toBase64: async function ( scope, { utf8_str }, exclude_padding=true) {
     if ( typeof( utf8_str ) !== `string`  ) {  // Prevent an error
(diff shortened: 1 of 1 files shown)


plocket (commented):
A good start. Might want to make this a valid base64 string. I think if you just remove spaces or something like that it would keep it safer. [Not sure how to still keep that sentence clear, though.] What do you think?

niharikasingh:
I used the rules from https://regexland.com/base64/ to turn the error message into a valid (but hopefully still readable) base64 string.

plocket (approved):
Confirmed can be validly converted to base 64! Thank you!

Issue #343, opened 2021-07-22
Discuss: Warn about vars in the data that are on the page, but hidden?


plocket:
> [Warn dev] if a variable in the var data is present on the page, but remains hidden the whole time.

It would be a distinct change in behavior. A plus is that we could probably easily send a convenient error message for the circumstance where it doesn't show up. Not sure what the return value would be to ensure that the loop doesn't move on.

plocket:
This is if a variable in the var data is present on the page, but remains hidden the whole time.

plocket:
Maybe give a warning that it wasn't revealed.

plocket:
Question: Why would the current report, which shows which fields were not used, not be sufficient?

A: I just think it might be a little more clear/explicit as opposed to looking at a list that may be long if it was generated by the test generator.

I think this is a future/v5 goal possibly.

plocket:
Also have [a request to add hidden fields info to random answers tests](https://github.com/SuffolkLITLab/ALKiln/discussions/531#discussioncomment-2442622).

Issue #342, opened 2021-07-22
Docs: Add to docs that tests may fail because the server is too slow and hidden fields did not get shown fast enough


plocket:
The developer would need to run the tests again.

plocket:
It's not the da server that would be too slow, it would be the GitHub server running the client. Race conditions can pop up everywhere in end-to-end testing.

PR #345, opened 2021-07-22
Close #341, Add ability to upload multiple files

### LINKED ISSUE #341: Step: upload a file
In preparation for https://github.com/SuffolkLITLab/docassemble-AssemblyLine/issues/139, create a step that will let a developer upload a file. My current thinking is that those files will be stored in the Sources folder as well. There may be puppeteer complications. Relevant resources:

- Puppeteer api: https://github.com/puppeteer/puppeteer/blob/v2.0.0/docs/api.md#elementhandleuploadfilefilepaths
- Strangely, differing answers on Stackoverflow: https://stackoverflow.com/questions/59273294/how-to-upload-file-with-js-puppeteer. There's an answer there that's independent of puppeteer if we get desperate enough.

plocket:
Should pass after https://github.com/plocket/docassemble-ALAutomatedTestingTests/pull/152 is reviewed as sufficient and merged.
Also suppress error from duplicate of [class="file-caption-name"]
input and increase `show if` timeout because test skipped setting
hidden field and all I can assume is that it didn't have enough
time to show up. Made #342, #343 for it.

Close #341


lib/scope.js
```diff
@@ -991,13 +994,78 @@ module.exports = {
...
+        await scope.afterStep(scope, { waitForShowIf: true });
       }
     },  // Ends scope.setField.input()
   },
 
+  uploadFiles: async function ( scope, { handle, set_to }) {
+    /* Try to upload one or more files to a file-type input field. */
+
+    // Allow a typo where someone forgot to include a space after a comma?
+    let file_names = set_to.split(`,`);
```
niharikasingh, line 1012: In my head, this line belongs closer to the for loop below where we use the `file_names` variable for the first time.

lib/scope.js
```diff
@@ -991,13 +994,78 @@ module.exports = {
...
+    }
+
+    let paths = [];
+    // If a file exist, add it to the list of files to be uploaded. If it doesn't, add a warning.
+    for ( let name of file_names ) {
+
+      // If the developer correctly used a ", ", there will be a space at the
+      // start of the string. Remove it. If they forgot the space, that's fine too.
+      name = name.replace(/^\s/, '');
+      let file_path = path.join( dir, name );
```
niharikasingh, line 1048: Looks like we're enforcing that all files listed in one line must be in the same directory (i.e., you can't have `some_png_1` in sources and `some_png_2` in features. We should document this.

niharikasingh (approved):

lib/scope.js
plocket: Great idea, thanks! #347

plocket: Now see https://github.com/SuffolkLITLab/docassemble-AssemblyLine-documentation/issues/207

PR #349, opened 2021-07-23
Hotfix: Wait for files to finish uploading before finishing upload Step

plocket:
I assumed puppeteer would take care of waiting, but it turns out that it does not. Gave me a chance to use Promises differently, though. It's nice to see their asynchronicity be useful for a change.


plocket:
I'd say we can squash this when we merge.

niharikasingh (approved):

Issue #340, opened 2021-07-20
Step (discuss): How do we identify files to download on the page? Current method not sufficient


plocket:
The download step uses the given string to find an href that contains the given string in it. In the AL interviews, almost all current interviews have a table with a 'View' button, a 'Download' button, and a 'Download zip' button. Examples of `href`s:

- View: href="/uploadedfile/89149/pdf_doc_1.pdf"
- Download: href="/uploadedfiledownload/89149/pdf_doc_1.pdf"
- Download zip: href="/uploadedfile/89152/multi_bundle_1.zip"

The numbers in there are different each time you run the interview. For the current download step you'd only be able to use 'pdf_doc_1.pdf' to identify the link and both the 'View' and 'Download' buttons would look the same.

I believe the difference we see with with the download file (`uploadedfiledownload`) is the use of `attachment=True` when calling the `url_for()` method of a document to create those links in the buttons.

Also, when writing tests for the ALDocument classes, I ended up showing the same document on the page multiple times. Are there reasons a regular developer may do the same? A file with the same href ending on the same page? That doesn't mean the documents would be the same, just that the filenames would be the same. Would they have the same text surrounding them too?

Options I can think of in some combination:

- Require that when a button will download something, it has to use `attachment=True` and then search for something that contains `uploadedfiledownload` and the given string.
- Allow user to provide regex to identify the href contents.
- Allow the user to pick the ordinal of the link (first, second, etc.) (i.e. if four links with the same file name appear, which item should be selected)
- Not language agnostic and probably not useful - Allow the user to identify text that's in an ancestor of the link. That would work for multiple files of the same name in different AL download tables where the row contains the title of the document as long as the titles were different. I'm not really sure that would be sufficient at all considering other factors.

None of these really seem sufficient. The most flexible one is ordinals, but that's also pretty fragile and, when code did change and tests failed, it would be hard to figure out why.

- [ ] When making these choices, we also need to double check what a regular attachment page's links look like [e.g. the docassemble default display of attachments with, I think `attachments:`], especially when files of the same name are included in the list of attachments.

plocket:
Another idea to add to the mix: Allow downloading multiple docs in one step - loop through the docs one by one, setting the custom timeout for each one so that, like story table rows, they each get the developer-defined amount of time to download.

plocket:
[...]

PR #353, opened 2021-07-25
#313, #314, Insert exception error messages into the reports as well

### LINKED ISSUE #313: Report: Add current `expect` messages to our custom report as well. Also just plain console logs if they're left.
Instead of just showing up with the cucumber log, they should also be in our reports. I think we may want to do both, though.

### LINKED ISSUE #314: Add auto-prefixes to some report items
[The `warning` and `error` types. Doing it now with `page_id`s.]

Hmm... should this print something like `${ item.type.toUpperCase() }: ${ item.value }\n`?

That way, we don't have to type ```{ type: `warning`, value: `WARNING: You are missing...``` every time we want to specify this is a warning etc.

_Originally posted by @niharikasingh in https://github.com/plocket/docassemble-cucumber/pull/310#discussion_r664990931_

plocket:
Close #313, add exception messages to reports.
Close #314, add auto-prefixes to some types of report lines.

Also adds some newlines/formatting to some messages to make those bits a bit shorter and more distinct.

### DIFF
--- a/lib/steps.js
+++ b/lib/steps.js
@@ -140,8 +140,13 @@ Given(/I start the interview at "([^"]+)"(?: in lang "([^"]+)")?/, async (file_n
   // Tap the language button if given
   let lang_url = null;
   if ( language ) {
+
     let [lang_link] = await scope.page.$x(`//a[text()="${ language }"]`);
-    expect( lang_link, `Could not find the link for the language "${ language }"` ).to.exist;
+
+    let msg = `Could not find the link with the text "${ language }"`;
+    if ( !lang_link ) { await scope.addToReport(scope, { type: `error`, value: msg }); }
+    expect( lang_link, msg ).to.exist;  // Cause test failure
+
     await scope.tapElement( scope, lang_link );
   }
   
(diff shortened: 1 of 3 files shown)


lib/steps.js
```diff
@@ -419,30 +443,44 @@ Then('I arrive at the next page', async () => {  // √
...
+  if ( !link ) { await scope.addToReport(scope, { type: `error`, value: msg }); }
+  expect( link, msg ).to.exist;
+
   await scope.afterStep(scope);
 });
 
-Then('I should see the link {string}', async (linkText) => {  // √
-  let [link] = await scope.page.$x(`//a[contains(text(), "${linkText}")]`);
-  expect(link).to.exist;
+Then('I should see the link to {string}', async ( url ) => {  // √
```
niharikasingh, line 464: The difference between "see the link xyz" and "see the link to xyz" being loosely match vs exact match isn't immediately obvious to me. Is there a way to reword this? Alternatively, you could deprecate the above function and keep just "see the link xyz". Or, you could merge the two functions somehow (first check for exact match, then if that doesn't exist, for loose match).

niharikasingh (approved):

lib/steps.js
plocket: It's actually even weirder than that and I'm planning to get rid of the previous one. That one uses the visible text of the link and is not language agnostic. This one uses the href of the link and _is_ language agnostic. The concern I currently have is over links with dynamic values. It's the same concern as the one about the download links and how to handle those (#340)

PR #354, opened 2021-07-25
Show how fixtures could be in actual .feature file

plocket:
Interested in input on this one - is it easier to maintain/read when the expected report text is in the `.feature` file itself, or does it just make things messy?

Also change Steps so that all report steps get checked in `After()` so that final parts of the Scenario report get added. Not sure it's completely necessary and would be interested in feedback on that too.

### DIFF
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -1463,4 +1463,13 @@ module.exports = {
     let report = await scope.getPrintableScenario( scope, { scenario });
     _internal_tests.reports( stored_report_key, report );
   },  // Ends scope.testReport()
+
+  testReport2: async function ( scope, { expected_report_text }) {
+    /* Test whether the report is as is expected. A meta test for testing
+    * the behavior of this testing framework. */
+    let current_report_obj = scope.report.get( scope.scenario_id );
+    let scenario = [ scope.scenario_id, current_report_obj ];  // Mimic how loop works with `Map`
+    let report = await scope.getPrintableScenario( scope, { scenario });
+    expect( expected_report_text ).to.equal( report );
+  },  // Ends scope.testReport()
 };
--- a/lib/steps.js
+++ b/lib/steps.js
@@ -695,15 +695,26 @@ When(/I set the address of ?(?:the var(?:iable)?)? "([^"]+)" to "([^"]+)"/, asyn
 // Framework Development
 //#####################################
 //#####################################
-Given(/the scenario error report should match "(.+)"/, async ( stored_report_key ) => {
-  /* Test a report for failing scenarios */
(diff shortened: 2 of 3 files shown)


tests/features/reports.feature
```diff
@@ -1,11 +1,30 @@
 @reports
 Feature: Reports show the right things
 
-Note: For now we'll have to check them visually
-
 @fast @r1
-Scenario: Report still has page id when I tap to continue without setting any fields
-  Given I start the interview at "all_tests"
+Scenario: Report still shows page id when I tap to continue without setting any fields
+  Given
```
niharikasingh, line 6: I think this should say "Given the Scenario should pass with this report:"

lib/scope.js
```diff
@@ -1463,4 +1463,13 @@ module.exports = {
     let report = await scope.getPrintableScenario( scope, { scenario });
     _internal_tests.reports( stored_report_key, report );
   },  // Ends scope.testReport()
+
+  testReport2: async function ( scope, { expected_report_text }) {
+    /* Test whether the report is as is expected. A meta test for testing
+    * the behavior of this testing framework. */
```
niharikasingh, line 1469: it would be helpful for future development if this explanation included how `testReport` is different from `testReport2`

lib/steps.js
```diff
@@ -730,20 +741,49 @@ After(async function(scenario) {
...
+
+      if ( expected_report_text ) {
+        await scope.testReport2( scope, { expected_report_text });
+      }
+
+    } catch ( err ) {
+      // Reset report values no matter what so they don't mess up future scenarios
+      // Maybe this should be in `Before`, but it would be so out of context there...
+      scope.expected_fail_report = null;
+      scope.expected_pass_report = null;
```
niharikasingh, line 764: Shouldn't these values get reset even if no error got thrown?

tests/features/reports.feature
plocket: It definitely should! Wonder why it's passing. Will have to fix that for sure.

lib/scope.js
plocket: If we use this methodology, `testReport` will be replaced with `testReport2`.

lib/steps.js
plocket: Good eye. It'll definitely be fixed before this is merged.

plocket:
Any general thoughts about one methodology over another?

plocket:
I'm currently working on trying out replacing both of those things with a different Step:
```
the (failed|passing) Scenario report should include
```

Thing is that I think in this case it should be able to collect things to test about the Scenario report and I haven't pinned down a data structure for that yet.

Issue #355, opened 2021-07-25
Test: Test warnings and errors that should show up in reports


plocket:

plocket:
This issue is too broad and will continue to grow. We've continued to grow the report tests. We'll have to make issues specific to certain tests at this point.

PR #356, opened 2021-07-26
#355, implement some report tests, alternative to #354

plocket:
This is an alternative to #354.

Allows testing multiple strings separately for one scenario, but excludes testing failing or passing. I'm not sure about it, but it does narrow the scope of this Step to one thing.

Is there a danger a Scenario would include the desired text, would be expected to pass, but would actually fail, and this wouldn't catch that? Should there be a separate 'the status of this Scenario is passing/failed' Step? I'm not sure. It just seemed conflated to have them both in the same Step.

(I think I accidentally pushed something to the other branch that I hadn't quite intended to, so you'll see some deletion of commented out code from that one. If we want to take this branch instead of #354, we should switch this to a PR on main. That way we can then review the changes fully in context.)

Addresses #355, though more tests need to be added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,7 +38,10 @@ Format:
 ### Security
 - 
 -->
-<!-- ## [Unreleased] -->
+## [Unreleased]
+### Added
+- Internal testing Step to test contents of reports include certain content.
+- Internal testing Step to check passing tests should be passing and failing ones should be failing.
 
 ## [2.4.1] - 2021-07-24
 ### Fixed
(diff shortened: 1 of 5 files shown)


plocket:
No worries about closing this or the other one btw - I can move over all the significant changes and tests to whatever we decide to go with.

niharikasingh:
I like the idea of keeping both features: `Given the scenario is (passing|failing)` and `Given the scenario report includes...`. I don't really have an opinion on whether both should be in the same step or two different steps, so I'll defer to you on that (or if you want to talk through the pros/cons, we can do that during a study session). 

(I also like the fact that in this implementation, the user doesn't need to include the whole report in the test if they don't want to. They can just include the important portions of the report.)

plocket:
Pointed to main and ready for final review, @niharikasingh

niharikasingh (approved):

PR #364, opened 2021-07-30
Re: #355, report tests, fix loading error not being triggered properly

plocket:
I think I'll just make a PR whenever I run into unexpected behavior that needs fixing so we can publish those as versions. Seems weird, but I'm not sure what would be better.

Addresses #355


lib/scope.js
```diff
@@ -299,8 +301,10 @@ module.exports = {
   },
 
   load: async function ( scope, file_name, attempt_num ) {
-    /* Try to go to the interview url a few times with a pause in between each try. */
-    let max_attempts = 3;
+    /* Try to go to the interview url a few times */
+    let attempt_timeout = 10 * 1000;
+    // End failed attempts a little before full timeout so that we can get a useful error message
+    let max_attempts = scope.timeout/attempt_timeout - 100;
```
niharikasingh, line 307: What's the reason for the `- 100`? 

- `scope.timeout = 30 * 1000` (Is this right?)
- `attempt_timeout = 10 * 1000`
- `max_attempts = 3 - 100`

niharikasingh (approved):

lib/scope.js
plocket: Amazing catch!

> What's the reason for the `- 100`?

I was hoping the comment would make that clear, so once we figure out a good description, we should improve the comment. It's there because a (very) little time will have passed before the code got here. That means that if we use the exact `scope.timeout` in the async processes in the loops to follow and all loops get triggered, the Step itself will timeout before this whole process finishes. That'll cause a generic timeout error, not our custom error.

That comment doesn't help the code work, though, if the math in general is wrong.

> * `scope.timeout = 30 * 1000` (Is this right?)

That's the default timeout we've set, yes.

> * `attempt_timeout = 10 * 1000`

No note on this line other than to say I'm still kind of doubting my decision to only give the page 10 seconds to load. Would love an opinion on that, but maybe that's a question for the da crew.

> * `max_attempts = 3 - 100`

You are totally right. I forgot the parens around `attempt_timeout - 100`!

Also, proper math would result in 3.03 repeating, which does highlight the fact that `Math.floor()` is missing from here and so the whole thing would not have worked even if the parens would have been there. Which is why I wish I knew a way to test this properly gorrammit. It would require the server to be so slow that it doesn't load in time, though. I don't know how to do that, but if I did then the tests for this framework would slow the server way down whenever they were run.

Thanks for the catch!

niharikasingh: Aah, I think the math error was the issue for me. Now that I know it should have parens, everything makes sense. 

Re: being able to test sllllllllllllow servers, did you look into this already? https://stackoverflow.com/questions/27881485/network-throttling-with-chrome-and-selenium Selenium allows simulating slow internet speeds. It's not technically the same as a slow server, but it'll still make the page load more slowly.

niharikasingh (approved):

lib/scope.js
```diff
@@ -1554,11 +1558,9 @@ module.exports = {
...
     let report = await scope.getPrintableScenario( scope, { scenario });
     let all_are_included = true;
-    for ( let content of expected ) {
-      if ( !content.includes( expected )) {
-        all_are_included = false;
-      }
-      expect( report ).to.contain( content );
+    for ( let one_expectation of expected ) {
+      if ( !report.includes( one_expectation )) { all_are_included = false; }
+      expect( report ).to.contain( one_expectation );
```
niharikasingh, line 1544: Not super important, but this seems slightly inefficient. Like 1562 will loop through all items in report once, and then 1563 will do so again. Moving this `expect` statement inside the above `if` block will make this slightly more efficient, because then the second loop won't happen every time.

niharikasingh (approved):

PR #365, opened 2021-08-02
Hotfix. Move load check so it doesn't error. Change the way repeated loading is attempted.

plocket:
I think the math is more simple now. I hope. It again wasn't quite the right thing before. This only gives two attempts at loading, but it does use fully half the allotted time to do so, which might be a good enough compromise - gives more time for the server to load. I dunno.

### DIFF
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -301,9 +301,11 @@ module.exports = {
 
   load: async function ( scope, file_name, attempt_num ) {
     /* Try to go to the interview url a few times */
-    let attempt_timeout = 10 * 1000;
-    // End failed attempts a little before full timeout so that we can get a useful error message
-    let max_attempts = Math.floor(scope.timeout/(attempt_timeout - 100));
+    // Each attempt to load will take half the time of the developer's preferred
+    // max timeout at this point. Except subtract a few ms - make sure to end failed
+    // attempts a little before full timeout so that we can get a useful error message
+    let max_attempts = 2;
+    let attempt_timeout = Math.floor(scope.timeout/max_attempts) - 50;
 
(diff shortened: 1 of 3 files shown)


lib/scope.js
```diff
@@ -315,34 +317,37 @@ module.exports = {
...
-        expect( false, msg ).to.be.true;
+        expect( end_status ).to.equal( `page loaded` );  // Can an error really be thrown in a `catch`?
       }
     }
 
-    if ( load_result.error ) {
-      await scope.addToReport(scope, { type: `error`, value: load_result.error });
-      expect( load_result.error ).to.equal( '' );
-    }
+    expect( end_status ).to.equal( `page loaded` );
```
niharikasingh, line 350: Instead of repeating this statement twice (once inside catch block and once outside), this might be a good place to add in a `finally` block.

niharikasingh (approved):

lib/scope.js
```diff
@@ -301,9 +301,11 @@ module.exports = {
 
   load: async function ( scope, file_name, attempt_num ) {
     /* Try to go to the interview url a few times */
-    let attempt_timeout = 10 * 1000;
-    // End failed attempts a little before full timeout so that we can get a useful error message
-    let max_attempts = Math.floor(scope.timeout/(attempt_timeout - 100));
```
niharikasingh, line 306: I think I see what you're talking about here! I think this would be more correct if it said `Math.floor(scope.timeout/(attempt_timeout + 100));` (changing the minus to a plus). We want each attempt to be given more time than the `attempt_timeout` value, because we want to account for the time needed to give a useful error message. In any case, your current code also works.

PR #368, opened 2021-08-04
Close #367, fix erroring too soon on no module error which

### LINKED ISSUE #367: Step: load page: When a certain error appears, keep trying to reload the page until the full timeout is up.
Not sure quite how to do this yet. Details:

More tests are failing more often now with a particular system error. It's quick and the error is something like `ModuleNotFoundError: No module named 'docassemble.playground123testingsomebranchname'`.

When working on loading the page, I had asked in docassemble about a da system page error that occurred when things were still reloading. I think I said something about the server reloading. The conversation seemed to indicate it would just cause a timeout. That info prompted the current code that ends quickly when detecting a system error.

With this new error message in hand, we're able to see that there might indeed be a system error when a project (module?) is not finished reloading:

> When my selenium-like code creates a project and pulls github code, it waits till `test.yml` can load for that project. When it immediately thereafter tries to load an actual interview file (that does really exist), it sometimes gets the `ModuleNotFoundError` . E.g. `ModuleNotFoundError: No module named 'docassemble.playground***testing155askquestion'`. What might be going on?

> Maybe the server is still in the middle of restarting and some threads have the updated module and others don't.

That means we need to change how this is working, but still retain the ability to have useful error messages.

Something like: If this particular error message comes up, if the timeout isn't up, keep trying to reload.

One thing that would help with this is: change the test run so that things fail when the setup fails [#55]. That way, we'll know if the code didn't get there at all or whether it's just not done reloading.

[Edit: That alone won't be enough to tell what kind of situation we're running into. The `test.yml` could load and then the error could turn up and it may not mean the package still needs to load and just needs extra time. It may be because the code was never able to get pulled. Just saying that failing on setup won't solve the main problem. `test.yml` just lets us know that the server itself is working.]

plocket:
needs to give a chance for the module to reload. Also, added
another check to the setup pull as per instructions. See issue
comment for details. That should supposedly help this not happen,
but I did add a line in the report if it still happens and we'll
hopefully be able to see if that's not working and then report on
it.

Hope these are enough to fix this problem finally, but I still don't know
of a way to test this reliably.

Close #367

### DIFF
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -299,11 +299,12 @@ module.exports = {
     await scope.tapElement( scope, elem );
   },
 
-  load: async function ( scope, file_name, attempt_num ) {
+  load: async function ( scope, file_name, attempt_num, start_time ) {
     /* Try to go to the interview url a few times */
     // Each attempt to load will take half the time of the developer's preferred
     // max timeout at this point. Except subtract a few ms - make sure to end failed
     // attempts a little before full timeout so that we can get a useful error message
+    start_time = start_time || Date.now();
     let max_attempts = 2;
     let attempt_timeout = Math.floor(scope.timeout/max_attempts) - 50;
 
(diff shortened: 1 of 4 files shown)


lib/scope.js
```diff
@@ -331,13 +332,26 @@ module.exports = {
...
-        await scope.addToReport(scope, { type: `error`, value: load_result.error });
-        expect( load_result.error ).to.equal( '' );
-      }
+
+        // `ModuleNotFoundError` could just be the server trying to reload, so give
+        // it more time until the timeout is almost over. End early to give good message.
+        // See https://github.com/plocket/docassemble-cucumber/issues/367
+        let global_timeout_is_almost_done = (Date.now() - start_time) >= (scope.timeout - 100);
+        if ( load_result.error.includes( `ModuleNotFoundError` ) && !global_timeout_is_almost_done ) {
+          await scope.addToReport(scope, { type: `warning`, value: `Attempted to load module. Got "${ load_result.error }"` });
```
niharikasingh, line 341: Maybe add to the error message "trying again"?

lib/scope.js
```diff
@@ -331,13 +332,26 @@ module.exports = {
...
+          await scope.addToReport(scope, { type: `error`, value: load_result.error });
+          expect( load_result.error ).to.equal( '' );
+        }
+
+      }  // ends if load_result.error
 
     } catch ( err ) {
       // If it doesn't work, maybe try again
-      if ( attempt_num <= max_attempts ) { await scope.load( scope, file_name, attempt_num ); }
+      if ( attempt_num <= max_attempts ) { await scope.load( scope, file_name, attempt_num, start_time ); }
```
niharikasingh, line 354: Changing this to `if ( attempt_num <= max_attempts ) || global_timeout_is_almost_done` would clarify the logic a little bit. That way, all the logging could be in the `if` block above, and the retrying in the `catch` block here. (Doing this will require changing `global_timeout_is_almost_done` to also take into account `load_result.error.includes( 'ModuleNotFoundError' )`.

niharikasingh (approved):

lib/scope.js
plocket: My concern is that we'll trigger an error before the check and that we still want to give it another try or two. That means that this logic needs to be in here as well. We do need to check for almost done here too, though, so I might move that up a level and use it in both places. How does that sound?

niharikasingh: I see the new issue. That proposed resolution sounds good to me.

PR #375, opened 2021-08-20
Unit testing done, still need to test in playground

### LINKED ISSUE #154: Allow to use the value 'today' for a date
[Maybe also today and then some math. For example `today - 1` or `today + 7`]

niharikasingh:
Closes #154


lib/scope.js
```diff
@@ -1003,6 +1003,27 @@ module.exports = {
         await scope.uploadFiles( scope, { handle, set_to });
         await scope.afterStep(scope, { waitForShowIf: true });
 
+      } else if (type == `date`) {
+        let set_to_date = set_to;
+        // If the date is specified in the format "today" or "today+2" or "today-1"
+        // it will need some pre-processing. 
+        if (set_to.indexOf(`today`) !== -1) {
+          // Find out how many days to add/subtract
+          let date_delta = set_to.match(/today([+-]\d*)?/)[1];
```
plocket, line 1012: The below may be more appropriate as the current regex allows just '-' or just '+' alone. It would then be defined and yet wouldn't be a valid int.

All to explain why requiring at least one digit is probably useful.

```suggestion
let date_delta = set_to.match(/today([+-]\d+)?/)[1];
```

plocket (commented):
Awesome work! A great feature that's been needing some love for a while. I think the report tests aren't the main place to test these changes - this is not some unique new report output. Probably the interactive tests and the table tests are the appropriate places for this kind of test. Does that make sense or am I off base here?

niharikasingh:
🎉 tests are passing

lib/scope.js
```diff
@@ -1065,6 +1065,27 @@ module.exports = {
         await scope.uploadFiles( scope, { handle, set_to });
         await scope.afterStep(scope, { waitForShowIf: true });
 
+      } else if (type == `date`) {
+        let set_to_date = set_to;
+        // If the date is specified in the format "today" or "today+2" or "today-1"
+        // it will need some pre-processing. 
+        if (set_to.indexOf(`today`) !== -1) {
```
plocket, line 1072: Just realized the developer may use spaces between some characters.
```suggestion
        if (set_to.indexOf(`today`) !== -1) {
          set_to = set_to.replace(/\s/g, '');
```

plocket (approved):
Awesome! I added one more suggestion that I just noticed, but it's always something we can take care of later. Feel free to merge as you wish.

plocket (approved):
Awesome! Thanks!

PR #381, opened 2021-09-29
Set default timeout before each scenario starts

### LINKED ISSUE #348: Fix scope.timeout is not reset at the beginning of a scenario
I think the last custom timeout that was set carries through to the other scenarios. We need to set our hardcoded default scope and cucumber timeout in the `Before` step.

niharikasingh:
Closes #348


plocket (approved):
This looks like exactly what's needed!

PR #382, opened 2021-09-29
Separate Scenario filenames, ids, and report headers for v3

### LINKED ISSUE #371: Allow numbers in safe filenames and report text
Right now, scenario names get stripped of numbers when converting them to either of the above. Avoid stripping those out as they can be quite useful.

niharikasingh:
Closes #371? 

I'm not sure what is required for this ticket. The function mentioned in the issue is already working as it should be. I added a test to confirm. 

[This has already been reviewed when #463 was merged.]

### DIFF
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -75,6 +75,36 @@ module.exports = {
     return trimmed;
   },  // Ends scope.toTrimmedJSON()
 
+  setReportHeading: async function (scope, { scenario }) {
+    if ( env_vars.DEBUG ) { console.log( scenario.pickle.name ); }
+    // Names for files and reports
+    // TODO: Allow scenario ids to have digits in their name
+
+    // With tag names automatically added to langauges, this should work to make
+    // each language unique.
+    let tags = '';
+    for ( let tag of scenario.pickle.tags ) {
+      tags += tag.name + '_';
+    }
+    tags = tags.replace(/_$/, '');
+    let new_id = `${ scenario.pickle.name }_${ tags }`;
+
+    // Should we add a counter? It would have to be unique to each scenario name.
+    // Not sure of another good way to indicate the _name_ is a duplicate
+    while ( scope.report.get( new_id ) !== undefined  ) {
+      new_id += '_';
+    }
+
+    scope.scenario_id = new_id;
+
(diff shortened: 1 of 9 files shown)


niharikasingh:
This is what scenario headings look like now: 

```
Scenario: 1633036242547_Fail with missng language link_@reports_@fast_@rf1_@error
---------------

ERROR: Could not find the link with the text "Latin"
**-- Scenario Failed --**

---------------
Scenario: Latin_1633036243584_Fail with wrong page id_@reports_@fast_@rf3_@error
---------------

ERROR: The question id was supposed to be "wrong question id", but it's actually "upload-files".
**-- Scenario Failed --**

---------------
Scenario: 1633036244829_Fail with a missing phrase_@reports_@fast_@rf4_@error
```

plocket (commented):
I haven't checked everything, but I'm seeing something in the error report that I didn't even think about before:

```
expected '\n---------------\nScenario: 1633057348354_Report lists
```

then more stuff, then

```
'\n---------------\nScenario: Report lists
```

The one shown above includes its timestamp. That will always be unique, but we're now not stripping out numbers. I haven't looked at the code, but there must be some processing we can do to strip it out of the report name before including other numbers... I have to believe that's true.

Take a shot at it if you can. If not, maybe we can meet about it.

plocket:
For some reason, the merge seems to have gotten rid of this functionality. I'm trying to track down what's going on.

lib/scope.js
```diff
@@ -75,6 +75,34 @@ module.exports = {
...
+    // TODO: Allow scenario ids to have digits in their name
+
+    // With tag names automatically added to langauges, this should work to make
+    // each language unique.
+    let tags = '';
+    for ( let tag of scenario.pickle.tags ) {
+      tags += tag.name.replace(/[^A-Za-z0-9]+/g, '_').replace(/(\P{L})+/gu, '_');
+    }
+    tags = tags.replace(/_$/, '');
+    let new_id = `${ scenario.pickle.name }${ tags }`;
```
plocket, line 90: This line used to be `scenario_name` in the PR previously. I think that's what may be different. I think we also may need to write separate functions to make the scenario heading than for making filenames. Would appreciate thoughts if people have them!

plocket: We have separated those concerns. See issue named in initial PR comment.

PR #400, opened 2021-12-07
Close #396 for v3, move cucumber to plain dependencies (not peer)

### LINKED ISSUE #396: Move cucumber to a dependency instead of a peer dependency
Also, probably remove devDependency of cucumber.

plocket:
Closes #396. You can see an explanation in the issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,10 @@ Format:
 - 
 -->
 <!-- ## [Unreleased] -->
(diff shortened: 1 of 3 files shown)


BryceStevenWilley (commented):
Not the right person to understand peer dependencies, but from a quick skim of https://nodejs.org/en/blog/npm/peer-dependencies/, LGTM.

Issue #438, opened 2021-12-17
For env_vars.js, create a function that gets all the environment variables


plocket:
I swear there was an issue for this, but I can't find it right now.

Instead of having `env_vars.js` create the objects `env_vars` once, make a function that processes and gets environment-variable-based values fresh each time, then use that function everywhere we currently need to get .

We still need to find a different name for this. Any ideas?

plocket:
Should it be a class with custom getters instead?

plocket:
Working on this along with #392 since it's going to use a bunch of those values.

rpigneri-vol:
Let's call this the "session".

plocket:
I like the addition of `session`. I think we may go with `session_vars` because of the deep association people have with `session` and the browser.

plocket:
In the current API PRs, I'm working on transferring everything to a "session_vars.js" file.

plocket:
Blocked by need to merge #442 before continuing progress.

plocket:
We need to do this for v3 as well as v4.

plocket:
Not blocked for v3 branch. Work off of release/v3 for now and use #442 session_vars.js as a model.

plocket:
Not implementing this for v3. There's a PR to close this linked above for v4.

plocket:
Closed by #513

Issue #432, opened 2021-12-13
Make da Project title more descriptive


plocket:
Right now we have `testing${branch_name}`. I think we should move to `$testing${repo_name}${branch_name}`.

Remember only alphanumeric that doesn't start with a number. No spaces or underscores.

I believe this is handled in `lib/utils/env_vars.js`.  Just before `env_vars.setProjectName`.

Some possible code:
```js
let repo_url_without_trailing_slash = env_vars.REPO_URL.replace(/\/$/, '');
let repo_name_parts = repo_url_without_trailing_slash.split('/');

let repo_name = repo_name_parts[ repo_name_parts.length - 1 ];
let no_starting_digit = ( `testing${repo_name}${env_vars.BRANCH_NAME}` ).replace( /^[0-9]+/i, '' );
env_vars.PROJECT_BASE_NAME = no_starting_digit.replace( /[^A-Za-z0-9]+/gi, '' );
```

plocket:
Because of development testing concerns I have, I'm doing this now in API development.

plocket:
This will be solved for v4 by the PR referenced in here. It still needs to be solved for v3, though. Can be seen in this line of code: https://github.com/SuffolkLITLab/ALKiln/blob/5782821f6de94335b6cf715e91986cccce4639a1/lib/utils/session_vars.js#L97.

Connected to #438, session_vars.

plocket:
We're leaving v3 behind for now. Has been solved for v4 at some point. Probably in the linked PR.

Issue #392, opened 2021-12-03
Integrate API key to remove/reduce errors caused by server reload


plocket:
Problem: When someone on the server edits a module, pulls in a repo with a module, or does various other things, the server restarts. That takes a while and causes test timeout errors - false failures. A test running, in fact, can cause a server reload timeout error.

Solution: There is a way to ping the API to see if the server is reloading/busy. Use that to detect when retrying is more appropriate than failing. This needs to be done wherever the test is going to have to wait for a page to load.

This will bump us to v4. We could take extra care to avoid breaking v3, but the complication introduced by that doesn't seem worth it.

We expect this change to take a while because of all the associated tasks and limitations. For example, currently, we don't know how to use the API at all, so this is going to involve that learning as well.

Things to implement before this will be ready for release:
1. Adjust the setup interview to take people through getting an API key.
2. Adjust the setup code to create the correct GitHub secret.
3. Code the API for the setup [and takedown] steps.
4. Code the API for running the tests.
5. [Remove use of username and password (not required, but nice). e.g. Use the API to set up tests as well? e.g. getting environment variables that we'd otherwise need to ask the user for? Though now it's seeming more and more likely that we won't need the ones that we can get through the API at runtime anyway.]

Things to do after implementation:
1. Find all the interviews using this testing framework.
2. Give them the appropriate GitHub secret.
3. Update their versions.

I think, though, that folks using this framework all use organizations (it's probably just us) and we will only need to adjust the organization secrets to make this work, so we can do it all in one go. [The packages that have repo secrets that override the GitHub secrets may have a problem. There are some of those for suffolk that are using the dev server instead of the test server. That said, when we update their script versions, admins can also check their secrets.]

See https://github.com/SuffolkLITLab/ALKiln/issues/389#issuecomment-984844240, "Option 2".

plocket:
## Server restart

Detecting that there isn't a problem with the  server:
> ...the API for installing a package returns a code that can be passed to the package_update_status endpoint, which will return completed when the server has finished restarting. See https://docassemble.org/docs/api.html#package_update_status
>
> Also, when polling that endpoint, I recommend using a short timeout and retry the API call in a loop. Depending on how busy the server is, a call to the package_update_status API may time out, so it is best to poll it in a loop, with the assumption that the server may not be able to respond to the status request while it is in the middle of restarting.

## Setup
Create a project: https://docassemble.org/docs/api.html#playground_post_project

[...]

PR #442, opened 2021-12-21
Address #392, use API to help with server functionality

plocket:
[~Still needs review, but merging shouldn't happen before merging the action and updating v3 with main.~ Merging can happen now.]

There's a lot here, both as notes in the issue and as changes in the files. Some of it is just changing a folder name, but this might take a real-time discussion depending on how much folks know about these API requests.

This addresses #392. Specifically the setup and takedown steps. The next part, helping manage the tests themselves, will probably be a lot smaller, if that's any comfort.

I hope I haven't missed anything important, but this is a big move. I can try to redo this in a more step-by-step fashion, making a PR for each step. I have some ideas how, so do let me know if that would help. I just needed to see how it would all fit together.

This also addresses a couple of other issues, and I'll try to track them all down and add them here. (Addresses #438, session_vars). (Addresses #432, better Project name, for v4, I think - title of docassemble Project. See [this line of code](https://github.com/SuffolkLITLab/ALKiln/blob/5782821f6de94335b6cf715e91986cccce4639a1/lib/utils/session_vars.js#L97).) It made sense considering the other changes I was making. I'm very unsure about how I'm handling `session_vars` (which are meant to eventually replace `env_vars`). ~It's a lot of getting things for stuff that's basically static, but there are some in there that need to be gotten, and I thought it would be more consistent.~ See comment below about at least part of why `session_vars` is all `.get...()`. Maybe it needs to be broken up further somehow, though. Open to suggestions.

In the future, `da_i.get_dev_id()` will be needed to get the interview address. Right now its use is a bit questionable. That discussion has started in #392.

`da_i.throw_an_error_if_server_is_not_responding()` is still more problematic, but it's not currently being used by anything. It was before and it will be again, but it'll have time to get worked out later I think.


plocket:
`session_var.get_foo()` functions: I just remembered part of why everything is inside a `.get_` - it's partly to make it easier to test. Right now, when you `require` `env_vars`, that creates the values once and we can't vary their values to run multiple tests in one file. This way, we'll be able to change `process.env` vars and then get the `session_var`s new values with `session_var` getting functions.

plocket:
I think I should also add `DOCASSEMBLE_DEVELOPER_API_KEY` to the action file now that it got merged into master. Also, a couple API keys disappeared and I'm trying to find out if those have an expiration date.

plocket:
Also, after approval, let me make sure other things are all set before we merge.

plocket:
This may require the setup interview to make a different `example.feature` - I'm not sure this ends up creating a test.yml file in the playground. We should test that somehow.

plocket:
Welp, did not expect that to happen. Continue to review and I'll work this out.

plocket:
It was a secret issue, we're good.

rpigneri-vol:
For later: Let's double-check how we handle DA server reboots caused by other end users uploading packages with modules.

rpigneri-vol:
Change to "Starting to upload your package to a new Project in the server's testing account."
```
ALKiln setup: Trying to create a docassemble Project in the server testing account and pull this repository's code into that Project.
```

Change to "Checking the testing account's API key."
```
ALKiln : Checking that the API key of the docassemble server's testing account is valid.
```

This can remain the same:
```
ALKiln : Your testing account's API key is perfectly valid.
```

Remove:
```  
ALKiln : Trying to create a docassemble project.
```

Change to "Trying to create a docassemble Project with the name . . ."
``` 
ALKiln : Trying to create a Project with the name testingdocassembleALAutomatedTestingTestsmain1
```

Change "was" to "is".
Remove the first sentence.  Put the Project name "testingdocassembleALAutomatedTestingTestsmain1".
```
ALKiln : The docassemble server was ready. Created a Project named testingdocassembleALAutomatedTestingTestsmain1  
```

* Change to "Stored name of Project locally".  Make this a debug log.
```
ALKiln: Saved new project name: testingdocassembleALAutomatedTestingTestsmain1
```

This does not need to change:
```
ALKiln : Trying to pull code from the "main" branch.
```

If we get a task_id, we will go down the following path:
Change to "Docassemble is pulling code with task_id "VeGlOAtgwSVjKxTNOBMBuKQU".  This usually happens if the uploaded package has a module.  Oftentimes, uploading these packages takes a few minutes."
```
ALKiln : Waiting for The docassmble server to restart after pulling code with task_id "VeGlOAtgwSVjKxTNOBMBuKQU".
```

Change "The docassmble" to "the docassemble".

~Change below to "The docassemble server is getting ready to restart."~
Change to "The docassemble server is uploading the package."
```
ALKiln : The docassmble server is still working on it.  
ALKiln : Attempt 1 of 15 did not succeed. Will try again in 5 seconds.  
ALKiln : The docassmble server is still working on it.  
ALKiln : Attempt 2 of 15 did not succeed. Will try again in 5 seconds.
```

~Change below to "The docassemble sever is starting up."~
~Change to "DA is in process of uploading the code."~
~Change to "DA is busy uploading your Project."~
~Change to "The docassemble server is busy uploading the module in your package."~
Change to "The docassemble server is still uploading the package." (This language intentionally differentiates for Kiln developers a working server from a busy server.)
```  
ALKiln : The docassmble server is busy.  
ALKiln : Attempt 3 of 15 did not succeed. Will try again in 5 seconds.  
ALKiln : The docassmble server is busy.  
ALKiln : Attempt 4 of 15 did not succeed. Will try again in 5 seconds.  
ALKiln : The docassmble server is busy.  
ALKiln : Attempt 5 of 15 did not succeed. Will try again in 5 seconds.  
ALKiln : The docassmble server is busy.  
ALKiln : Attempt 6 of 15 did not succeed. Will try again in 5 seconds.  
```

Change to "The docassmble server has uploaded your package."
```
ALKiln : The docassmble server restarted successfully.  
ALKiln : Package has been pulled and the docassmble server has finished restarting.  
```

This can remain the same:
```
ALKiln setup: Success! Test setup has finished successfully.  
```

Other Thoughts

As an alternative use case, if we do not get a task_id, we could print a comfort message via setInterval to let the end user know that Kiln is still processing.
* Could we maybe have the first five iterations say "Working on it" and the rest say, "Still working on it"?

Takedown

Move the "removing file" message to a Debug log level.

plocket:
To follow up on the conversations we had to build those comments:

> For later: Let's double-check how we handle DA server reboots caused by other end users uploading packages with modules.

I found that we're just using long timeouts, which seems to work fine, but gives the user no feedback about what's going on.

> As an alternative use case, if we do not get a task_id, we could print a comfort message via setInterval to let the end user know that Kiln is still processing.
> * Could we maybe have the first five iterations say "Working on it" and the rest say, "Still working on it"?

I still like this idea. I don't know if I specified this when we wrote the above comment, but `setInterval()` is an asynchronous function, so I think it has the best chance of working under these circumstances. I haven't yet tried using such a thing around an `await` and `await` is blocking specifically asynchronous functionality, so I'm not yet sure how it'll behave.

This was a great conversation and I think it's come up with messages that will communicate with the user much better than what's in there now! Thanks!

lib/docassemble/docassemble_api_interface.js
```diff
@@ -0,0 +1,377 @@
...
+da_i.get_dev_id = async function ( dev_id_options ) {
+  /** For consistency, wrap _da.get_dev_id().
+  *    See https://docassemble.org/docs/api.html#user.
+  *    Should we throw the 'Access Denied' error in here instead? */
+
+  // Default timeouts
+  let { timeout } = dev_id_options ||  { timeout: 75 * 1000 };
+  return await _da.get_dev_id( timeout );
+};  // Ends da_i.get_dev_id();
+
```
rpigneri-vol, line 94: Something to think about:
* We could create two functions: one to check if the server is active, and another to actually return the developer ID.  The first one will just validate that a develop ID is true and return true if the value is returned correctly.  The other will actually return the developer ID.

lib/docassemble/docassemble_api_interface.js
```diff
@@ -0,0 +1,377 @@
...
+  } catch ( error ) {
+    // There's only one possible sever response error
+    if ( error.response && error.response.status === 403 ) {
+      // Throw in here, or at the caller
+      throw( `Error: Request failed with status code 403, "Access `
+        + `Denied". This testing account's `
+        + `API key does not exist on this server. Are you sure your `
+        + `SERVER_URL GitHub secret is set to the server address on `
+        + `which this testing account exists? Are you sure your `
+        + `DOCASSEMBLE_DEVELOPER_API_KEY GitHub secret is correct? `
```
rpigneri-vol, line 123: DOCASSEMBLE_DEVELOPER_API_KEY GitHub secret is a key associated with this user?

lib/docassemble/docassemble_api_interface.js
```diff
@@ -0,0 +1,377 @@
...
+    // As long as this succeeds, we're good
+    await da_i.get_dev_id( api_key_options );
+    log.info({ pre: `Your testing account's API key is perfectly valid on this server.` });
+    return true;
+
+  } catch ( error ) {
+    // There's only one possible sever response error
+    if ( error.response && error.response.status === 403 ) {
+      // Throw in here, or at the caller
+      throw( `Error: Request failed with status code 403, "Access `
```
rpigneri-vol, line 118: Is there a more specific exception type that we could throw here, or is that not how node works?  In Python, for example, you might want to throw a requests.AccessDeniedException during an AccessDenied situation.

lib/docassemble/docassemble_api_interface.js
```diff
@@ -0,0 +1,377 @@
...
+    } else {
+      // Any other error should be thrown
+      throw error;
+    }  // ends throw different errors
+  }  // ends try to get dev_id
+};  // Ends da_i.try_api_key()
+
+
+// node -e 'require("./lib/github_integration/docassemble_api_interface.js").loop_to_create_project()'
+da_i.loop_to_create_project = async function( create_project_options ) {
```
rpigneri-vol, line 138: Let's consider renaming "create_project_options" to "creation_options".

lib/docassemble/docassemble_api_interface.js
```diff
@@ -0,0 +1,377 @@
...
+        } else {
+          // Other response errors should be thrown with link to docs
+          throw `${ error.response.status }: ${ error.response.data } See https://docassemble.org/docs/api.html#playground_post_project for more details.`;
+        }
+
+      } else {
+
+        // If it's a timeout error, server is busy. Try again after waiting.
+        if ( is_timeout_error( error ) && name_incrementor < max_tries ) {
+          log.info({ pre: `The docassemble server was busy when trying to create project. `
```
rpigneri-vol, line 181: Capitalize Project.

lib/docassemble/docassemble_api_interface.js
```diff
@@ -0,0 +1,377 @@
...
+            + `Attempt ${ name_incrementor } of ${ max_tries }. Will try again `
+            + `after ${ wait_seconds } seconds.` });
+          await time.waitForTimeout( wait_seconds * 1000 );
+          continue;
+
+        // A non-timeout, non-server error should just be thrown
+        } else {
+          log.warn({ pre: `Ran into unexpected error when trying to create project.` });
```
rpigneri-vol, line 189: Capitalize Project.

lib/docassemble/docassemble_api_interface.js
```diff
@@ -0,0 +1,377 @@
...
+          throw error;
+
+        }  // ends type of non-response error
+      }  // ends type of reponse error
+    }  // ends try
+  }  // ends while
+
+  // Failed because there were too many project names already taken
+  if ( response.data && response.data.includes( `Invalid project` ) && name_incrementor > max_tries) {
```
rpigneri-vol, line 198: Project, not project.

lib/docassemble/docassemble_api_interface.js
```diff
@@ -0,0 +1,377 @@
...
+
+  try {
+
+    let response = await _da.pull( timeout );
+
+    // If server didn't need to restart
+    if ( response && response.status === 204 ) {
+      log.debug({ pre: `Status 204 on pull. Did not need the docassemble server to restart.` });
+      return null;
+    }
```
rpigneri-vol, line 224: We decided to add a customer-facing message saying that the code got pulled.

lib/docassemble/docassemble_api_interface.js
```diff
@@ -0,0 +1,377 @@
...
+      log.info({
+        pre: `\`task_id\` "${ task_id }" has expired. If it's been `
+          + `less than 1 hour since the task started, the task is `
+          + `complete. See https://docassemble.org/docs/api.html#restart_status.`,
+        data: response,
+      });
+
+      // Server is working, but status unexpected? What to return? Throw?
+      // We shouldn't even get here according to the API docs, but it feels
+      // weird to have no default behavior
```
rpigneri-vol, line 273: We have decided that the meaning of this comment is clearer if it is moved above the log.info(...) line.

lib/docassemble/docassemble_api_interface.js
```diff
@@ -0,0 +1,377 @@
...
+    log.info({ pre: `Ran into an error when deleting Project ${ project_name }` });
+    throw error;
+  }
+
+};  // Ends da_i.delete_project()
+
+
+// node -e 'require("./lib/github_integration/docassemble_api_interface.js").__delete_projects_starting_with( "testingdocassembleALAutomatedTestingTestsmain", 1, 1 )'
+da_i.__delete_projects_starting_with = async function ( base_name, starting_num=1, ending_num=50 ) {
+  /** Useful for development with accidental infinite loops. */
```
rpigneri-vol, line 364: We have decided to change this to "***FOR DEVELOPMENT ONLY*** Useful for manually removing a set of projects that start with a shared prefix.  This happens when an infinite loop is created."

lib/docassemble/docassemble_api_interface.js
```diff
@@ -0,0 +1,377 @@
...
+    }  // ends if done_restarting
+  }  // ends while not done restarting
+
+  log.info({ pre: `Package has been pulled and the docassmble server has finished restarting.` });
+};  // Ends da_i.wait_for_server_to_restart()
+
+
+// Delete the Project
+// node -e 'require("./lib/github_integration/docassemble_api_interface.js").delete_project( "create1" )'
+da_i.delete_project = async function ( delete_options, server_wait_options ) {
```
rpigneri-vol, line 337: We decided to remove the server_wait_options argument.

lib/docassemble/docassemble_api_interface.js
```diff
@@ -0,0 +1,377 @@
...
+  let { timeout } = delete_options ||  { timeout: 75 * 1000 };
+
+  let project_name = session_vars.get_project_name()
+  log.info({ pre: `Trying to delete Project ${ project_name }` });
+
+  try {
+
+    let response = await _da.delete_project( timeout );
+    log.info({ pre: `Deleted Project "${ project_name }"` });
+    return response.data;
```
rpigneri-vol, line 350: We do not need to return data since the DELETE request type does not return a body.

lib/docassemble/docassemble_api_interface.js
plocket: Errors (exceptions) in js do have type names, including server response errors. I don't think any "local" error types fit here. I think it's most appropriate as a server response error. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error#error_types.

I also think that sometimes those error type names aren't helpful. E.g. 403's type is "forbidden". I think that would be more alarming than helpful. My take for this situation - the message here gives much more information and if they want to find out more about the error type, looking up "403" would be enough. Let me know what you think! See https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403.

Regardless, we should definitely watch out for other places where these type names would be helpful.

plocket: If you want to include a type, you can do it this way:
```js
throw new ReferenceError( `Text of error` );
```

lib/docassemble/docassemble_api_interface.js
plocket: Ah. Actually, this is not a message to the user. This is actually the text that docassemble sends back in the error and we do have to check for that exact text unless we use regex with the `i` flag.

plocket: Sorry that I missed that when we were discussing it.

rpigneri-vol (changes requested):

rpigneri-vol (approved):
These changes look good.  It looks like everything has made it in that we need!

PR #446, opened 2021-12-24
Create/overwrite package.json with our action.yml, closes #441

### LINKED ISSUE #441: Further simplify the developer's process using our composite action
Further simplify the developer's process. Edit our action so it creates the files the dev's repo needs to run the tests. Specifically package.json. Devs then won't need any file other than their workflow file.

That's it.

In sync with that:
- Create instructions in documentation on how to set up the development environment locally
- Create security instructions in documentation. See https://github.com/SuffolkLITLab/ALKiln/issues/425
(https://github.com/SuffolkLITLab/docassemble-AssemblyLine-documentation)

_Originally posted by @plocket in https://github.com/SuffolkLITLab/ALKiln/issues/420#issuecomment-998995228_

[We will overwrite their package.json if they have it. It doesn't commit the files it makes, so if it's even able to affect other jobs in the workflow, it can only affect them in the workflow that uses our action.

If that becomes a problem, we'll probably recommend they keep the conflicting jobs in separate workflows. If there are reasons why they can't do that, we can recommend they set up their own action instead of using our action, and that they control their own package.json. That's how the dependents do it now. It'll take more upkeep on their part, but if they're doing that kind of development, it's likely they will need to take control of those things anyway.]

plocket:
Closes #441. Merging it into `main` as well. Not yet sure how to develop multiple stable releases, but giving it a shot. Will update releases/v3 when this is merged to catch it up with main before the big divergence.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@ Format:
 <!-- ## [Unreleased] -->
 ### Added
 - Add action.yml that runs most of what users' workflows run now, along with notes for a new user workflow that will take less maintainance. See #420. We need to add documentation on how to write a workflow file as it currently is if they want to take back control. Setup interview has not yet been updated with this workflow.
+- Add package.json creation/overwriting to action.yml. Simplify package.json.
 
 <!-- ## [3.0.1-peer-deps.1] - 2021-12-07 -->
 ### Removed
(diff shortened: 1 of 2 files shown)


BryceStevenWilley (approved):
:rocket:

PR #463, opened 2021-12-30
Separate Scenario report heading vs. filename vs scenario id

plocket:
Basically, I looked at this and it seemed like maybe it was time for a bigger change. Instead of putting numbers in scenario headers and/or adjusting tests to allow those to pass, I'm trying to make a more extensive change to how we're managing that whole system.

Adds base-scenario-file-name creation and adds tests for that too.
Uses "-" as separator for major items as long as it doesn't break our .gitignore and artifact-getting name matching in the workflows. A change in that will have to wait for a breaking version.

For #371.

### DIFF
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -75,33 +75,32 @@ module.exports = {
     return trimmed;
   },  // Ends scope.toTrimmedJSON()
 
-  setReportHeading: async function (scope, { scenario }) {
+  addReportHeading: async function (scope, { scenario }) {
+    /** Return printable heading based on Scenario name and tags.
+    *    Allow characters from other languages.
+    */
     if ( env_vars.DEBUG ) { console.log( scenario.pickle.name ); }
-    // Names for files and reports
-    // TODO: Allow scenario ids to have digits in their name
 
-    // With tag names automatically added to langauges, this should work to make
-    // each language unique.
-    let tags = '';
+    // Collect all tag names
+    let tag_names = [];
     for ( let tag of scenario.pickle.tags ) {
-      tags += tag.name.replace(/[^A-Za-z0-9]+/g, '_').replace(/(\P{L})+/gu, '_');
-    }
-    tags = tags.replace(/_$/, '');
-    let new_id = `${ scenario.pickle.name }${ tags }`;
-
-    // Should we add a counter? It would have to be unique to each scenario name.
-    // Not sure of another good way to indicate the _name_ is a duplicate
-    while ( scope.report.get( new_id ) !== undefined  ) {
-      new_id += '_';
+      tag_names.push( tag.name );
     }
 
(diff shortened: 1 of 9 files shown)


lib/scope.js
```diff
@@ -75,33 +75,32 @@ module.exports = {
...
-    // Should we add a counter? It would have to be unique to each scenario name.
-    // Not sure of another good way to indicate the _name_ is a duplicate
-    while ( scope.report.get( new_id ) !== undefined  ) {
-      new_id += '_';
+    let heading = `\n---------------\n`
+      + `Scenario: ${ scenario.pickle.name }`;
+
+    if ( tag_names.length > 0 ) {
+      heading += `\nTags: ${ tag_names.join(' ') }`;
     }
```
k3KAW8Pnf7mkmdSMPHz27, line 95:
```suggestion
    if ( tag_names ) {
      heading += `\nTags: ${ tag_names.join(' ') }`;
    }
```

k3KAW8Pnf7mkmdSMPHz27 (approved):
LGTM! :partying_face:

lib/scope.js
plocket: I think you said this is a convention in some linters. You also said you prefer the way the code currently is. Thanks for bringing up the discussion. When we talk about code style, we can include this as one of the items. For now, I, like you, prefer to be explicit.

tests/unit_tests/getSafeScenarioFilename.test.js
```diff
@@ -1,100 +1,83 @@
...
+ 
+   it(`replaces two consecutive invalid chars with one '_' each`, async function() {
+     let result = await getSafeScenarioFilename( scope, { prefix: names.two_consecutive_invalid_input });
+    test_filename( names.two_consecutive_invalid_output, result );
+   });
+ 
+   it(`does not replace numbers`, async function() {
+     let result = await getSafeScenarioFilename( scope, { prefix: names.numerical_input });
+    test_filename( names.numerical_output, result );
+   });
```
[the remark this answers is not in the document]
plocket: Yeah, I can't see a reason to put `_` on the end of the filenames. Maybe we were appending something there previously. I'll change that before merging.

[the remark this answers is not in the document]
plocket: Ohhh, this is the prefix, not the whole name. My bad. Original point taken - we should not de-duplicate underscores.

Also, this currently isn't effectively deduplicating anyway. That regex would be different.

lib/scope.js
k3KAW8Pnf7mkmdSMPHz27: I double-checked, apparently, my suggestion is incorrect for JavaScript. 
```
if ( tag_names.length ) {
``` should be possible but it doesn't change much X)

Issue #470, opened 2022-01-07
Step: compare YAML files


plocket:
### Feedback needed for behavior

1. Where should the comparison YAML files be stored? React with ❤️ for the Sources folder. React with 👍 for a separate folder on the repo itself. For other ideas, add a comment. The developer will use the name of the comparison file in the Step.
2. Should the downloaded YAML files be made into artifacts only if they error or always? React with 😄 for only if they error, react with 🎉 for always.

Requested for ALWeaver refactoring regression tests.

**Step text:**

```
And "file_url_name.yml" matches "original_filename.yml"
When I download "file_url_name.yml" it matches "original_filename.yml"
```
~[I'm planning on option 2 at the moment]~

~Should it be `matches` or `is the same as`? Something else?~

Wanted to highlight that it needs to be a text value, so went with
```
Then the text in the JSON variable "variable_with_text_value" should be
```

### Other logistics

**YAML comparison:**

I've looked around and https://www.npmjs.com/package/js-yaml seems like the hottest kid on the block.

Theoretically, we might be able to just compare the YAML files as text, but then superficial changes like whitespace fixes would require a new file uploaded. Comparing JSON will be more flexible. We can use [chai's `.deep.equal()`](https://www.chaijs.com/api/bdd/#method_deep) to check for equality.

Someday we can think of adding a pretty diff, but I had trouble finding a good library. For now we might leave instructions to use a diff checker.

**Renaming files:**

We can only do it in a kind of annoying and gnarly way - download the file, detect the completion, rename that specific file after it's downloaded
- Get current filename so we can change it after: https://stackoverflow.com/a/57419156/14144258. Not sure what we'll do if the filename the user gives to rename to is the same as another file that's currently in the folder.
- Rename the file: https://stackoverflow.com/a/22504722/14144258

plocket:
React to this with 👍 if a filename should automatically append its page id regardless of the name the user gives to it.

plocket:
For YAML and this situation specifically, we might make a totally different step. Maybe:

```
Then the value of the JSON variable "var_name" should match "file_name.txt/json/something requirable"
```

> It looks like the interview source is in a template, not directly a variable you could grab. https://github.com/SuffolkLITLab/docassemble-ALWeaver/blob/e7a83717e0079f16a2fae5b18a54b079b2f255f1/docassemble/ALWeaver/data/questions/assembly_line.yml#L1420-L1426. It would be a 1 line fix to save the string as a variable though:
[...]

PR #479, opened 2022-01-08
Add v3 Step to compare JSON page var string value to given str value 

plocket:
Also:
- Downloads page's JSON vars to .json file in "downloads" artifacts.
- Adds tests for the step and its error report.

Addresses #470 for v3. Will have to do the same for v4 before closing.

### DIFF
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -179,6 +179,30 @@ module.exports = {
     return safer_name;
   },  // Ends scope.getSafeScenarioFilename()
 
+  getJSONFilename: async function( scope ) {
+    let filename = await scope.getSafeScenarioFilename( scope, {
+      prefix: `downloads_${ scope.page_id }_JSON`
+    });
+    filename += `.json`;
+    return filename;
+  },
+
+  savePageJSONData: async function( scope ) {
+    /** Save the JSON variables of the current page to a json file.
+    *
+    * @returns {Object} info Info about the result.
+    * @returns {string} info.filename The name of the file that was saved.
+    * @returns {Object} info.json The variables object.
+    */
+    // Get a unique json file filename
+    let filename = await scope.getJSONFilename( scope );
+    // Get the data
+    let json = await scope.getPageJSON( scope );
+    // Save the data
+    fs.writeFileSync( filename, JSON.stringify( json, null, 2 ));
+    return { filename, json };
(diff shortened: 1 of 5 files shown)


lib/steps.js
```diff
@@ -573,6 +573,44 @@ Then(/the "([^"]+)" link opens in (a new window|the same window)/, async (linkTe
...
+  if ( actual_value !== expected_value ) {
+    await scope.addToReport( scope, {
+      type: `error`,
+      value: `The variable "${ var_name }" was not equal to the expected `
+        + `value on the "${ scope.page_id }" screen. Check the downloads `
+        + `file ${ filename } to see the page's JSON variables.`
+    });
+  }
+
+  expect( actual_value ).to.equal( expected_value );
```
BryceStevenWilley, line 611: Does the `expect` here print the actual value and the expected value of the variable? In my experience, those values are much more helpful than just the name of the variable, and it currently looks like the tester has to go to the saved JSON report file and to the original test that failed to see what those values are.

lib/scope.js
```diff
@@ -179,6 +179,30 @@ module.exports = {
     return safer_name;
   },  // Ends scope.getSafeScenarioFilename()
 
+  getJSONFilename: async function( scope ) {
+    let filename = await scope.getSafeScenarioFilename( scope, {
+      prefix: `downloads_${ scope.page_id }_JSON`
+    });
+    filename += `.json`;
+    return filename;
+  },
```
BryceStevenWilley, line 188: From my JS ignorance: why does this need to be `await`ed? Or maybe the question should be why is `getSafeScenarioFilename` async? Nothing in it seems to be dependent on IO, it's just string concatenation.

BryceStevenWilley (approved):
LGTM! just a few clarifying questions.

lib/steps.js
plocket: It does print the values.

lib/scope.js
plocket: You're right about async - because it's async it needs `await`. It's async because everything's in scope right now and I made scope consistent. When I call something in scope, I wasn't always sure what needed to be `await`ed and what didn't, so I made it all async. There's an issue out there somewhere about refactoring name/string manipulation into its own file, and other such refactoring, so that not everything is in scope and not everything needs to be async.

PR #488, opened 2022-01-10
Security: Pinned colors library to 1.4.0

plocket:
Because it's a security issue, I already published it. ~I think. I'd love a double check on this because I published before I committed and I don't recall if that flow is sufficient.~ I'm just going to check it myself. Faster that way.

See https://www.bleepingcomputer.com/news/security/dev-corrupts-npm-libs-colors-and-faker-breaking-thousands-of-apps/

We don't use other libraries mentioned in there, so we're good on those.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,11 +39,15 @@ Format:
 - 
 -->
 <!-- ## [Unreleased] -->
-## [3.0.7] - 2022
+## [3.0.7] - 2022-01-10
+### Security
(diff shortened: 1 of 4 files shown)


plocket:
Draft until we can test the action. See #489.

BryceStevenWilley (approved):
Changes here LGTM.

plocket:
Confirmed as working at https://github.com/plocket/docassemble-CompositeActionsTest/actions/runs/1678835970. This is for v3, so I'm going to make the same change in v4. I might do special admin magic to be able to merge it myself.

Issue #40, opened 2020-11-25
Add explicit error handling for a name question argument that has too many parts.


plocket:
Add explicit error handling for names with too many parts.

_Originally posted by @niharikasingh in https://github.com/plocket/docassemble-cucumber/pull/36#discussion_r530060503_

Same for addresses that don't have the correct format.

plocket:
This is the function that needs help: https://github.com/SuffolkLITLab/ALKiln/blob/releases/v3/lib/steps.js#L755-L790. If the developer used more than the max number of space separated strings, they should be told only the max number will be used.

plocket:
Also, if needed, remove any excess strings from the end of the list if necessary.

plocket:
Closed by #501

plocket:
Closed by #504

PR #501, opened 2022-01-28
[WIP] Max spaces, user prompt, excess string

mcdonaldcarolyn:
Work in progress

This is pr when I am working on it, will let you know if I make progress.


lib/steps.js
```diff
@@ -777,6 +777,12 @@ When(/I set the name of ?(?:the var(?:iable)?)? "([^"]+)" to "([^"]+)"/, async (
   ];
 
   let name_parts = name_str.split(' ');
+
+  if(name_parts.length > 4) {
+    await scope.addToReport(scope, { type: `warning`, value: `There were too many inputs in the name field` });
+    name_parts = name_parts.slice(0, 4);
+  }
+
```
mcdonaldcarolyn, line 785: Unsure of the warning language and style, let me know what you prefer there.

tests/features/assembly_line.feature
```diff
@@ -41,3 +41,15 @@ Scenario: I have four name parts
...
+Scenario: I have five name parts
+  Given I start the interview at "AL_tests"
+  And I set the name of "users[0]" to "Uli Udo User Sr Jr"
+  And I tap to continue
+  And I set the address of "users[0]" to "120 Tremont Street, Unit 2, Boston, MA 02108"
+  And I tap to continue
+  And I set the name of "users[1]" to "Uli2 User2"
+  And I set the var "users[i].proxy_var" to "Mixed proxy var step test"
+  And I tap to continue
+  Then the question id should be "end"
```
mcdonaldcarolyn, line 55: copied and pasted the test above, to create test for new code. Does this meet your standards for test cases, happy to change.

mcdonaldcarolyn (commented):
Created if statement to ensure there are only 4 name parts and added test feature to test the code.

tests/features/assembly_line.feature
plocket: The test looks great! There's another step to take and I'll walk you through that. We'll be moving it to the report test file and adding a test to make sure the warning appears.

mcdonaldcarolyn: Oh great!!!

lib/steps.js
```diff
@@ -777,6 +777,12 @@ When(/I set the name of ?(?:the var(?:iable)?)? "([^"]+)" to "([^"]+)"/, async (
   ];
 
   let name_parts = name_str.split(' ');
+
+  if(name_parts.length > 4) {
+    await scope.addToReport(scope, { type: `warning`, value: `There were too many inputs in the name field` });
```
plocket, line 782:
```suggestion
    await scope.addToReport(scope, { type: `warning`, value: `The name "${ name_str }" has more than 4 parts, but 4 is the maximum allowed. The test is using the first 4 parts.` });
```

lib/steps.js
plocket: This was great! It used our format to great effect. I made a "suggestion" for the warning language that I think gives more information. You can give me feedback on it or you can just commit it, whichever you prefer.

mcdonaldcarolyn: hmm either works for me. Does it need to consider how people can make mistakes and autofill? Or am I REALLY over thinking it?

plocket (changes requested):
Great start and it'll be a great improvement for feedback to developers. There are some comments I've made about changes and after those are resolved, we'll be ready to merge 👍

lib/steps.js
plocket: I had never considered autofill. Dang. We'll have to think about that, but maybe as a separate exploratory issue.

Regarding other mistakes, [anything should be permissible for names](https://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/), so it can't give wrong input for the text fields. If you mean the dropdown, that would be a developer mistake and they'd need to fix that.

To explain more and [increase the bus factor](https://developerexperience.io/practices/bus-factor), I figure that a developer is much more likely to add an extra name in the middle than they are to get the wrong suffix (final string). We sort of see it in your test example - the last word in the string is "III".

If we do this, we might tweak the test - instead of having the last two strings be suffixes, only the last one should be a valid suffix.

plocket: Autofill: Do we press "enter" after filling in a field? If not, we should be good.

plocket: We don't! We just get into the next field from what I can tell. Unless puppeteer is doing something weird and pressing `enter` somewhere, we don't do it ourselves, so we should be good. Thanks for bringing it up, it was a great thing to check on!

plocket (approved):
This looks great and gives a lot better feedback to our users! 🚀 

Thanks for this contribution. We'll merge it, add to the changelog, and then publish it 👍

plocket:
Closes #40

PR #506, opened 2022-02-08
Allow a developer to wait as a first Step v4. #387. Add test.

### LINKED ISSUE #387: Fix "I wait _ seconds": Allow that step to work before an interview has been loaded
That step (correctly) uses our `afterStep` code to complete what it's doing. We need to configure `afterStep` to handle situations where no page has yet been loaded so that it can be used even when interviews have not yet been loaded so that someone can wait before an interview loads.

This is actually a hack for dealing race conditions for the first test of the test suite. Right now, we're having trouble with server load times after the repo has been pulled into a project - the server often still hasn't completely loaded and our load tests aren't sufficient for that. That's a whole other complicated discussion. This fix might also be useful under other circumstances, though I don't know what those would be.

plocket:
Closes #387.

Also, generally adds safety measures for when page does not exist.

Very similar to PR #459, but moving the responsibility down to a spot that most functions make use of, meaning that it'll be applied to a lot more cases. They mostly won't need it, but it might still be worth being more comprehensive.


lib/scope.js
```diff
@@ -207,45 +209,47 @@ module.exports = {
...
-      await scope.addToReport(scope, { type: `error`, value: system_error_text });
-      throw system_error_text;
-    };
+        await scope.addToReport(scope, { type: `error`, value: system_error_text });
+        throw system_error_text;
+      };
+
+      // Wait for elements that might get revealed
+      if ( options.waitForShowIf ) { await scope.waitForShowIf( scope ); }
+    }  // ends if scope.page
```
plocket, line 245: Most of `afterStep` should only be run if `scope.page` exists.

tests/features/establishing_steps.feature
```diff
@@ -20,3 +20,7 @@ Scenario: I am able to set a custom wait time before an interview has been loade
 Scenario: Interview name includes url args
   Given I start the interview at "url_args.yml&from=theinternets&random=zoo"
   Then I should see the phrase "zoo"
+
+@slow @e4 @wait_first
+Scenario: I can wait as a first step in a test
+  Given I wait 1 second
```
BryceStevenWilley, line 26: Should also add another test that uses `I wait 1 second` in a real test. Copying the previous test as an example:
```suggestion
  Given I wait 1 second
  
@fast @e3 @urlargs
Scenario: Interview name includes url args
  Given I wait 1 second
  Then I start the interview at "url_args.yml&from=theinternets&random=zoo"
  Then I should see the phrase "zoo"
```

BryceStevenWilley (approved):
LGTM, but should add one more test that uses the new `Wait` at the beginning of an interview.

PR #508, opened 2022-02-08
Remove failing `description` in copy/paste example in action.yml

plocket:
The comments there are supposed to contain code that our users (developers) can copy/paste into their file. Right now if they do that their file will fail. This will fix that.


mcdonaldcarolyn (approved):

mcdonaldcarolyn:
Looks good to me.

PR #509, opened 2022-02-09
Add "log in" to server Step for v4, closes #499

### LINKED ISSUE #499: Add Step to log into server account (user request)
Requested feature. Add Step to allow the dev to use a username and a password GitHub SECRETs to sign in to the developer's server before going to the interview.

For now, this will be done in a Step, not in the table because:
1. The DOM is different on the login page.
2. The interview needs to go to the server address, not the interview address.
3. We'd have to create magical syntax to let the developer specify which items in the values column were actually the names of GitHub SECRETs as opposed to just regular things to input into a field. They have to be secrets so that a casual observer can't see them and use them to log in to the server nefariously.

[When we write the docs, remember to instruct the developer to add the secrets to their `with` block in their workflow so that kiln can see those secrets too.]

plocket:
This should look pretty familiar to some folks, with a few tweaks here and there owing to the differences between v3 and v4.

Includes tests. Usually, the test goes to the interview as an unregistered user. Now developers can choose to have their 'user' sign in before going to their interview.

Closes #499

Squashed commits:
- Add "log in" to server Step with test. No report tests yet.
- Add more comprehensive tests for "log in" Step

[Will have to adjust once versions of dependencies are updated in previous PR]


lib/scope.js
```diff
@@ -452,6 +452,49 @@ module.exports = {
...
+
+    // Try to log in
+    await scope.page.type( `#email`, email );
+    await scope.page.type( `#password`, password );
+    await scope.page.$eval( `button[name="submit"]`, (elem) => { return elem.click(); });
+
+    // Wait to see the success message.
+    // This is a bit of an assumption and needs success and failure testing
+    // TODO: use a Promise.race to fail faster.
+    await scope.page.waitForSelector( `.datopcenter .alert-success` );
```
mcdonaldcarolyn: Would it be helpful to add message about login success to reports?

lib/steps.js
```diff
@@ -220,6 +220,22 @@ Given(/I start the interview at "([^"]+)"(?: in lang "([^"]+)")?/, {timeout: -1}
   await custom_aferStep_timeout;
 });
 
+Given(/I log ?(?:in)?(?:on)?(?:to the server)? with(?: the email)? "([^"]+)" and(?: the password)? "([^"]+)"(?: SECRETs)?/i, async ( email, password ) => {
```
mcdonaldcarolyn: Would it be useful to allow  "sign in" as well as log in.

tests/features/reports.feature
```diff
@@ -345,6 +345,37 @@ Scenario: I can't match JSON page var to str
   wrong_value
   """
 
+@fast @rf22 @login
+Scenario: Fail with wrong email secret
```
mcdonaldcarolyn, line 349: Could add specific failed message about email address

tests/features/reports.feature
```diff
@@ -345,6 +345,37 @@ Scenario: I can't match JSON page var to str
...
+  Given the final Scenario status should be "failed"
+  Given the Scenario report should include:
+  """
+  email address GitHub SECRET
+  """
+  Given I log in with the email "WRONG_EMAIL_NAME" and the password "USER1_PASSWORD"
+
+@fast @rf23 @login
+Scenario: Fail with wrong password secret
```
mcdonaldcarolyn, line 358: Could add specific failed message about password.

lib/steps.js
mcdonaldcarolyn: Also could be helpful to add a test for that.

mcdonaldcarolyn (approved):
This looks great, please see comments for future changes.

tests/features/reports.feature
plocket: I know what I said, but I've actually realized that since this is in "reports.feature" file, we can assume these are about messages/lines in the report. Sorry about that.

tests/features/reports.feature
plocket: Same as previous.

Issue #164, opened 2021-03-11
Update to new cucumber package location for v7?


plocket:
Will involve a change in cucumber package name among other things. See message at https://www.npmjs.com/package/cucumber.

plocket:
When we do this, we will need to change the code in the setup interview and update all the action files of the interviews that currently use our package. In which case, it might be good to first discuss how we're handling these kinds of very awkward updates so in future we don't have to update everyone's action scripts. See #128 for that discussion.

plocket:
Migration guidance: https://github.com/cucumber/cucumber-js/blob/main/docs/migration.md

plocket:
Closed by #503

PR #519, opened 2022-02-19
Update report output to use new cucumber statuses

plocket:
We were using the old cucumber lower-case statuses - "passed", "failed" - so reports then didn't put them in the correct sentences. Related to #164, updating to newest version of cucumber.


mcdonaldcarolyn (approved):
Looks great!

PR #521, opened 2022-03-02
483 throw error for uploading nonexistent files v4

### LINKED ISSUE #483: Remove the "none of those files exist in the upload directory" warning
The current error is confusing and isn't needed:
https://github.com/SuffolkLITLab/ALKiln/blob/87068a28c44bc569f216cb2453986e4643144227/lib/scope.js#L1143

A previous warning is already shown when a particular file can't be found at the location and that is already more clear:
https://github.com/SuffolkLITLab/ALKiln/blob/87068a28c44bc569f216cb2453986e4643144227/lib/scope.js#L1137

Change the `if` statement:
```js
if ( paths.length > 0 ) { ... }
```

We don't need an `else`. We also don't need the comment about possibly throwing an error - we don't want to error here. Our general goal, at the team's request, is to be pretty permissive with what we let slide.

mcdonaldcarolyn:
also change warning to error in report
closes #483


plocket (changes requested):
Dang, you need to delete log.txt in this branch. At some point you committed stuff when you didn't yet have the new .gitignore. Happens to me all the time :/

plocket (approved):
All set, thank you!

PR #527, opened 2022-03-23
Switch to suffolklitlab org, allow testing pre-release, fix

### LINKED ISSUE #393: Create npm package with the correct name under the correct org
Right now, it's called `docassemble-cucumber`. We need `alkiln`, I think. If Suffolk is interested, we also need to find out if Suffolk has a npm org and put these releases under that.

plocket:
typo in regex branch name parsing.

Close #393 for v4. Will not be addressing v3.

See it working at https://github.com/plocket/docassemble-CompositeActionsTest/actions/runs/2029985148 ~https://github.com/plocket/docassemble-CompositeActionsTest/actions/runs/2029915837~

[This is supposed to be derived from https://github.com/SuffolkLITLab/ALKiln/compare/releases/v4...393_test_action_npm, which is a more tested version. I just didn't want all those experimental commits.]

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -53,6 +53,7 @@ runs:
         echo "PLAYGROUND_PASSWORD=${{ inputs.PLAYGROUND_PASSWORD }}" >> $GITHUB_ENV
         echo "PLAYGROUND_ID=${{ inputs.PLAYGROUND_ID }}" >> $GITHUB_ENV
         echo "DOCASSEMBLE_DEVELOPER_API_KEY=${{ inputs.DOCASSEMBLE_DEVELOPER_API_KEY }}" >> $GITHUB_ENV
+        echo "ALKILN_VERSION=${{ inputs.ALKILN_VERSION }}" >> $GITHUB_ENV
       shell: bash
     - name: "ALKiln: confirm info"
       run: echo -e "\nRepo is $REPO_URL\nBranch ref is $BRANCH_PATH\nManually added languages are $EXTRA_LANGUAGES\nAll other data is secret\n"
--- a/action.yml
+++ b/action.yml
@@ -21,41 +21,18 @@ inputs:
     description: 'Other languages you want to test. This should be in your GitHub SECRETS or the SECRETS of your org.'
     required: false
     default: ''
+  # This is necessary to test pre-releases at the very least. It does seem
+  # strange considering we're controlling everything through the action, but
+  # that actually doesn't touch the npm version that's pulled in.
+  ALKILN_VERSION:
+    description: 'Version of ALKiln to get from npm.'
+    required: false
+    default: '4.0.0'
 
 runs:
   using: "composite"
   steps:
-    # This doesn't use an already established package-lock.json and that
-    # comes with some types of test breaking risks. Same problem as before.
-    - name: Overwrite or create package.json
-      run: |
-        cat << EOF > "package.json"
-        {
-          "scripts": {
-            "test": "npm run langs_setup && npm run cucumber -- $@",
-            "cucumber_base": "./node_modules/.bin/cucumber-js --require ./node_modules/docassemble-cucumber/lib/index.js",
-            "cucumber": "run_cucumber(){ npm run cucumber_base -- \"$@\" docassemble/*/data/sources/*.feature; }; run_cucumber",
-            "langs": "npm run langs_setup && ./node_modules/.bin/cucumber-js --require ./node_modules/docassemble-cucumber/lib/index.js docassemble/*/data/sources/*_lang*.feature",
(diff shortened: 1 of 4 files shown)


action.yml
```diff
@@ -76,11 +53,43 @@ runs:
...
         echo "DOCASSEMBLE_DEVELOPER_API_KEY=${{ inputs.DOCASSEMBLE_DEVELOPER_API_KEY }}" >> $GITHUB_ENV
+        echo "ALKILN_VERSION=${{ inputs.ALKILN_VERSION }}" >> $GITHUB_ENV
       shell: bash
     - name: "ALKiln: confirm info"
       run: echo -e "\nRepo is $REPO_URL\nBranch ref is $BRANCH_PATH\nManually added languages are $EXTRA_LANGUAGES\nAll other data is secret\n"
       shell: bash
 
+    # This doesn't use an already established package-lock.json and that
+    # comes with some types of test breaking risks. Same problem as before.
+    - name: Overwrite or create package.json
```
k3KAW8Pnf7mkmdSMPHz27:
```suggestion
    - name: Temporarily overwrite or create package.json
```

action.yml
```diff
@@ -76,11 +53,43 @@ runs:
...
+        }
+      shell: bash
+
+    # Install node
+    - name: "ALKiln: Install node packages"
+      uses: actions/setup-node@v1
+      with:
+        node-version: "17"
+    - run: npm install
+      shell: bash
```
k3KAW8Pnf7mkmdSMPHz27, line 91: illegitimate nitpick: it is out-of-scope of this PR, but I believe https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#defaultsrun will remove the need of repeating the `shell: bash`.

k3KAW8Pnf7mkmdSMPHz27 (approved):
LGTM! My GitHub Actions-fu isn't great, but from what I know this looks good :+1:

action.yml
plocket: #528

action.yml
```diff
@@ -76,11 +53,43 @@ runs:
...
+            "colors": "1.4.0",
+            "@cucumber/cucumber": "^7.3.2",
+            "@suffolklitlab/alkiln": "$ALKILN_VERSION"
+          }
+        }
+      shell: bash
+
+    # Install node
+    - name: "ALKiln: Install node packages"
+      uses: actions/setup-node@v1
```
k3KAW8Pnf7mkmdSMPHz27, line 87: The latest [setup-node](https://github.com/actions/setup-node#usage) is version 2

It makes sense that Dependabot does this by default, but if it doesn't it can be configured to do so as well (e.g., https://github.com/codeforamerica/brigade-project-index-statusboard/blob/a288e657777479ea82ce0fa09a056ef956a83ce3/.github/dependabot.yml#L3-L6)

plocket: #529

lib/utils/session_vars.js
```diff
@@ -37,7 +37,7 @@ session_vars.get_branch_name = function () {
   /* Get the name of the repository's branch that is being tested */
   let branch_name = null;
   if ( process.env.BRANCH_PATH ) {
-      branch_name = process.env.BRANCH_PATH.replace( `refs/head/`, '' );
+      branch_name = process.env.BRANCH_PATH.replace( `refs/heads/`, '' );
```
BryceStevenWilley, line 40: Also out of scope of this PR, but the `github.ref` variable in the action can be from [tags as well](https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables). DA doesn't let you pull from tags, just a note that should make it into docs or comments of the action.yml at some point.

package.json
```diff
@@ -30,9 +30,9 @@
   "author": "",
   "license": "MIT",
   "bugs": {
-    "url": "https://github.com/plocket/docassemble-cucumber/issues"
+    "url": "https://github.com/SuffolkLITLab/ALKiln/issues"
   },
-  "homepage": "https://github.com/plocket/docassemble-cucumber#readme",
+  "homepage": "https://suffolklitlab.github.io/docassemble-AssemblyLine-documentation/docs/automated_integrated_testing",
```
BryceStevenWilley:
```suggestion
  "homepage": "https://suffolklitlab.org/docassemble-AssemblyLine-documentation/docs/automated_integrated_testing/",
```

BryceStevenWilley: the `*.github.io` URL redirects to the `*.org` address anyway.

lib/utils/session_vars.js
plocket: #530

PR #535, opened 2022-03-29
Misc Quality of Life fixes

BryceStevenWilley:
Majority of correcting misspellings, which I only noticed because I have a spell checker in my IDE.
Only non-spell check fixes I made were:
* removed one unused variable `classname`
* Shows the id of the page that a table run got stuck on / experienced an error on.
* lets you say "I see the phrase ..." instead of "I should see the phrase..." It remove a word, and when reading Gherkin, I get the impression that "should" is implied, and doesn't really need to be stated.
* Adds a `cucumber_interview` command to let you run `*.feature` files from a different local dir. I needed this to actually run the features from the interview I was testing with this. If there's a way to run those (without having to install the npm package, which seems like it could make it difficult to develop or it might break npm in weird ways) with normal npm commands, would defer to that.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
   "scripts": {
     "test": "npm run unit && npm run langs_setup && npm run cucumber -- $@",
     "cucumber_base": "./node_modules/.bin/cucumber-js --require ./lib/index.js",
+    "cucumber_interview": "run_cucumber(){ npm run cucumber_base -- \"$@\"/docassemble/*/data/sources/*.feature; }; run_cucumber",
     "cucumber": "run_cucumber(){ npm run cucumber_base -- \"$@\" tests/features/*.feature; }; run_cucumber",
     "langs": "npm run langs_setup && ./node_modules/.bin/cucumber-js --require ./lib/index.js tests/features/*_lang*.feature",
     "langs_setup": "./lib/utils/langs.js 'tests/features/*.feature'",
--- a/package.json
+++ b/package.json
@@ -6,7 +6,6 @@
   "scripts": {
     "test": "npm run unit && npm run langs_setup && npm run cucumber -- $@",
     "cucumber_base": "./node_modules/.bin/cucumber-js --require ./lib/index.js",
-    "cucumber_interview": "run_cucumber(){ npm run cucumber_base -- \"$@\"/docassemble/*/data/sources/*.feature; }; run_cucumber",
     "cucumber": "run_cucumber(){ npm run cucumber_base -- \"$@\" tests/features/*.feature; }; run_cucumber",
     "langs": "npm run langs_setup && ./node_modules/.bin/cucumber-js --require ./lib/index.js tests/features/*_lang*.feature",
     "langs_setup": "./lib/utils/langs.js 'tests/features/*.feature'",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -6,6 +6,7 @@
   "scripts": {
     "test": "npm run unit && npm run langs_setup && npm run cucumber -- $@",
     "cucumber_base": "./node_modules/.bin/cucumber-js --require ./lib/index.js",
+    "cucumber_interview": "run_cucumber(){ npm run cucumber_base -- \"$@\"/docassemble/*/data/sources/*.feature; }; run_cucumber",
```
plocket: I believe your use case, but I don't quite understand it. Can you tell me how to recreate the circumstances in which it's running?

This is non-breaking and it doesn't have to block this PR, but a path in this location is referring to an internal path of this repository and that path doesn't exist here currently. Are you using it as a sub-module somewhere?

plocket (approved):
These are some great improvements! Thank you! I've got one question that doesn't have to be blocking if you have a mo for it. It looks good to go! 🚀

package.json
BryceStevenWilley: If I'm running locally and want to run tests on a different interview that I'm writing the tests out for, there are only three things that I can do:
* run `npm run cucumber` from the ALKiln dir, which can't find the *.feature files from the interview I'm actually writing tests for
* run `npm run cucumber` from the other interview's dir, but then I get  `Missing script "cucumber"`, or `Cannot find package.json`. There's probably a way to tell npm "Hey, use the package.json from this other dir, not the one I'm currently in", but I can't find it.
* run this command, `npm run cucumber_interview ../path/to/other/interview` from the ALKiln dir, which correctly searches for and finds the feature files from the other docassemble interview.

> This is non-breaking and it doesn't have to block this PR, but a path in this location is referring to an internal path of this repository and that path doesn't exist here currently.

It's not an internal path: the `$@` takes the last arg given to `npm run cucumber_interview`, which is the path to another locally installed docassemble interview, i.e. `npm run cucumber_interview ../docassemble-EFSPIntegration`, and then cucumber finds all of the features files in that package.

plocket: Ah, I see. `..` does a lot to clarify what's going on, thank you. If I understand correctly, running arbitrary tests for some other arbitrary interview. Sounds good. Is this an artifact of how the repos don't have `package.json` files of their own now? It may be worth thinking more about that problem. I have my own ways of handling it, but this seems like a good one too.

We might want to think about the command name to help future people understand its use, but that can happen later. A couple ideas (just to get them down while they're on my mind) are `npm run external_files` and `externa_tests`.

Btw, you can also enter any of this code into the command line and it should be able to execute appropriately. Nice to have a built-in command, though.

BryceStevenWilley: > We might want to think about the command name to help future people understand its use

My suggestion after seeing yours is `cucumber_external`. 

> you can also enter any of this code into the command line

I know, but it's a bad idea to have any non-trivial bash command not written down somewhere

BryceStevenWilley: I'm actually running into some more issues using this: internally, we call `process.pwd()` to find where the potential list of uploadable files are, which this solution doesn't address. I might take this back out work on it a bit more before making another PR specifically for that.

PR #536, opened 2022-03-29
Secret vars

BryceStevenWilley:
Depends on #535.

Adds a step that gets the value to set a variable to from an environment variable. Also, prevents that variable from showing up in the report, replacing the actual value with `${ENV_VAR_NAME}`.

Still need to:
* [x] add a test interview in https://github.com/plocket/docassemble-ALAutomatedTestingTests, and a feature test to https://github.com/SuffolkLITLab/ALKiln/blob/main/tests/features/interactive_steps.feature.
* [ ] add documentation to the docs page

### DIFF
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -486,7 +486,7 @@ module.exports = {
       // See tests > unit_tests > tables.fixtures.js > tables.old_to_current_formatting
       let actual_var_value = row.value;
       if (is_secret_var(row)) {
-        let actual_var_value = process.env[row.flags.secret_name];
+        actual_var_value = process.env[row.flags.secret_name];
         if ( actual_var_value === undefined ) {
           let not_in_env_msg = `The GitHub SECRET "${ row.flags.secret_name }" doesn't exist. Check for a typo and check your workflow file.`;
           await scope.addToReport( scope, { type: `error`, value: not_in_env_msg })
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -1531,8 +1531,7 @@ module.exports = {
           let source_row = row;
           if ( row.source ) { source_row = row.source; }
           let original_row = source_row.original;
-          let reported_val = printable_var_value(original_row); 
-          let msg = `Did not find a field on this page for the variable \`${ original_row.var }\` that could be set to \`${ reported_val }\``;
+          let msg = `Did not find a field on this page for the variable "${ original_row.var }" that could be set to "${ original_row.value }"`;
           await scope.addToReport(scope, { type: `warning`, value: msg });
         }
       }
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -490,7 +490,7 @@ module.exports = {
         if ( actual_var_value === undefined ) {
           let not_in_env_msg = `The GitHub SECRET "${ row.flags.secret_name }" doesn't exist. Check for a typo and check your workflow file.`;
           await scope.addToReport( scope, { type: `error`, value: not_in_env_msg })
-          throw ReferenceError(`Could not find "${row.flags.secret_name }": "${not_in_env_msg}".`)
+          throw ReferenceError(`Could not find "${row.flags.secret_name }": ${not_in_env_msg}.`)
         }
       }
 
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -22,13 +22,18 @@ let escape_regex = function( to_escape ) {
   return to_escape.replace(/[-\/\\^$*+?.()|[\]{}]/g, '\\$&');
 };
 
-let printable_var_value = function(row) {
+let get_original_row = function (row) {
   /* `row` might be a row, a source row, or an original row. */
 
   // Until we flatten row data a litte, we need to find the `orignal` row, which has the flags
   let source_row = row.source || row;
   let original = source_row.original || source_row;
+  return original;
+}
 
+let printable_var_value = function(row) {
+  /* `row` might be a row, a source row, or an original row. */
+  let original = get_original_row(row);
   if ( is_secret_var(original) ) {
     return original.flags.secret_name;
   }
@@ -37,11 +42,7 @@ let printable_var_value = function(row) {
 
 let is_secret_var = function(row) {
   /* `row` might be a row, a source row, or an original row. */
-
-  // Until we flatten row data a litte, we need to find the `orignal` row, which has the flags
-  let source_row = row.source || row;
-  let original = source_row.original || source_row;
-  
+  let original = get_original_row(row);
   return original.flags !== undefined && original.flags.secret_name !== undefined;
 }
 
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -22,7 +22,7 @@ let escape_regex = function( to_escape ) {
   return to_escape.replace(/[-\/\\^$*+?.()|[\]{}]/g, '\\$&');
 };
 
-let display_var_value = function(row) {
+let printable_var_value = function(row) {
   if ( is_secret_var(row) ) {
     return '${' + row.flags.secret_name +'}';
   }
@@ -81,9 +81,6 @@ module.exports = {
     let original = source_row.original;
     let var_str = await scope.toTrimmedJSON( scope, { str: original.var });
     let value_str = await scope.toTrimmedJSON( scope, { str: original.value });
-    if (is_secret_var(original)) {
-      value_str = display_var_value(original);
-    }
     let trigger_str = await scope.toTrimmedJSON( scope, { str: original.trigger });
     if ( trigger_str === scope.trigger_not_needed_flag ) {
       // This text was added by ALKiln itself. It's not part of the user's data
@@ -481,19 +478,29 @@ module.exports = {
 
   normalizeTable: async function ( scope, { var_data, from_story_table=true }) {
     /* Return data derived from cucumber variable-setting data.
-    * Table must have headers. Arrays are not supported. */
+     * Table must have headers. Arrays are not supported. */
 
     // Support tables with no 'trigger' column
     let supported_table = [];
     for ( let row of var_data ) {
       // See tests > unit_tests > tables.fixtures.js > tables.old_to_current_formatting
+      let actual_var_value = row.value;
+      if (is_secret_var(row)) {
+        let actual_var_value = process.env[row.flags.secret_name];
+        if ( actual_var_value === undefined ) {
+          let not_in_env_msg = `The GitHub SECRET "${ row.flags.secret_name }" doesn't exist. Check for a typo and check your workflow file.`;
+          await scope.addToReport( scope, { type: `error`, value: not_in_env_msg })
+          throw ReferenceError(`Could not find "${not_in_env_msg}".`)
+        }
+      }
+
       let result = {
         original: row,
         // Give fake data if needed to prevent some kind of hidden input field from being found
         // See bug at https://github.com/plocket/docassemble-cucumber/issues/79
         // May be able to remove this once converted to `getAllFields`
         var: row.var || 'al_no_var_name',
-        value: row.value,
+        value: actual_var_value,
         trigger: row.trigger || '',
         times_used: 0,
       }
@@ -501,7 +508,9 @@ module.exports = {
       supported_table.push( result );
     }
 
-    if ( session_vars.get_debug() ) { console.log( 'scope.normalizedTable():\n', JSON.stringify( supported_table )); }
+    if ( session_vars.get_debug() ) { 
+      console.log( 'scope.normalizedTable():\n', JSON.stringify( supported_table )); 
+    }
 
     return supported_table;
   },  // Ends scope.normalizeTable()
@@ -1522,7 +1531,7 @@ module.exports = {
           let source_row = row;
           if ( row.source ) { source_row = row.source; }
           let original_row = source_row.original;
-          let reported_val = display_var_value(original_row); 
+          let reported_val = printable_var_value(original_row); 
           let msg = `Did not find a field on this page for the variable \`${ original_row.var }\` that could be set to \`${ reported_val }\``;
           await scope.addToReport(scope, { type: `warning`, value: msg });
         }
@@ -1626,7 +1635,7 @@ module.exports = {
     * normalization.
     */
     let value = row.value;
-    let reported_value = display_var_value(row); 
+    let reported_value = printable_var_value(row); 
     let int_val = parseInt( value );
     
     // Step stuff. Take into account a lot of possible values
@@ -1810,16 +1819,10 @@ module.exports = {
 
     set_secret_var: async ( scope, var_name, set_to_env_name ) => {
       /** Sets a non-story table variable to a "secret" value (i.e. an env var) */
-      let set_to = process.env[set_to_env_name];
-      if ( set_to === undefined ) {
-        let set_to_env_msg = `The GitHub SECRET "${ set_to_env_name }" doesn't exist. Check for a typo and check your workflow file.`
-        await scope.addToReport( scope, { type: `error`, value: set_to_env_msg })
-      }
-
       let var_data = await scope.normalizeTable( scope, {
               var_data: [{
                 var: var_name,
-                value: set_to,
+                value: set_to_env_name,
                 trigger: scope.trigger_not_needed_flag,
                 flags: {secret_name: set_to_env_name}
               }],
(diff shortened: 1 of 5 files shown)


lib/scope.js
```diff
@@ -70,6 +81,9 @@ module.exports = {
     let original = source_row.original;
     let var_str = await scope.toTrimmedJSON( scope, { str: original.var });
     let value_str = await scope.toTrimmedJSON( scope, { str: original.value });
+    if (is_secret_var(original)) {
```
plocket: If I understand correctly, this is redundant now.

plocket:
This is a great improvement. We might actually need to ensure this behavior in several more places.

lib/scope.js
```diff
@@ -70,6 +81,9 @@ module.exports = {
     let original = source_row.original;
     let var_str = await scope.toTrimmedJSON( scope, { str: original.var });
     let value_str = await scope.toTrimmedJSON( scope, { str: original.value });
```
plocket: I _think_ `original.value` is already the secret's name, not its value. I might be mixed up on this. I may have been mixed up during the conversation we had the other day. Maybe worth discussing.

It occurs to me now that one potential approach (with some verification that it makes sense before implementing) might be to always use `original.value` whenever we save a variable to a file or to print to the GitHub console. Unless there are specific places where we feel it would be ideal to print as much data as we can, I suppose.

lib/scope.js
```diff
@@ -1792,6 +1808,27 @@ module.exports = {
       await scope.setFields( scope, { var_data, ensure_navigation: false, ensure_all_vars_used: true });
     },  // Ends set_regular_var()
 
+    set_secret_var: async ( scope, var_name, set_to_env_name ) => {
+      /** Sets a non-story table variable to a "secret" value (i.e. an env var) */
+      let set_to = process.env[set_to_env_name];
+      if ( set_to === undefined ) {
+        let set_to_env_msg = `The GitHub SECRET "${ set_to_env_name }" doesn't exist. Check for a typo and check your workflow file.`
+        await scope.addToReport( scope, { type: `error`, value: set_to_env_msg })
```
plocket: If the secret doesn't exist, should we throw an error here? It's possible that the field will require text input, in which case the test might be able to move on just entering 'undefined' or some such. Do we want to be that permissive?

lib/scope.js
```diff
@@ -1792,6 +1808,27 @@ module.exports = {
...
+      }
+
+      let var_data = await scope.normalizeTable( scope, {
+              var_data: [{
+                var: var_name,
+                value: set_to,
+                trigger: scope.trigger_not_needed_flag,
```
plocket, line 1843: Non-blocking - it may be worth adding a comment that we intend to avoid letting people set secret variables in a regular story table/as part of a group of other fields, so they should never need a trigger variable.

plocket: Never mind, this is explicitly the `set_secret_var` step. Lost my place.

lib/steps.js
```diff
@@ -112,7 +113,8 @@ Given(/I start the interview at "([^"]+)"(?: in lang "([^"]+)")?/, {timeout: -1}
   while ( num_attempts < max_load_attempts ) {
 
     num_attempts++;
-    let timeout_message = `Failed to load "${ file_name }" after ${ num_attempts } tries. `
+    let url = `${ await get_base_interview_url() }${ file_name }`;
+    let timeout_message = `Failed to load "${ file_name }" at URL ${ url } after ${ num_attempts } tries. `
```
plocket: Just fyi, this is technically a GitHub secret as well. Do we want to save it to files? I can't see why someone would want to keep their server url secret, but it's inconsistent treatment of secrets. Should we use the name of the secret instead? Wish GitHub had non-secret config env vars.

lib/steps.js
```diff
@@ -735,6 +737,15 @@ When(/I set the var(?:iable)? "([^"]+)" to "([^"]+)"/, { timeout: -1 }, async (
   );
 });
 
+When(/I set the var(?:iable)? "([^"]+)" to secret "([^"]+)"/, { timeout: -1 }, async ( var_name, set_to_env_name ) => {
```
plocket: "the" isn't necessary, but we've been trying to make these all very human sentences, so it might be more consistent to allow a "the" in here. I also made it case insensitive. I've been trying to remember to do that more just to give people some wiggle room.

```suggestion
When(/I set the var(?:iable)? "([^"]+)" to ?(?:the)? secret "([^"]+)"/i, { timeout: -1 }, async ( var_name, set_to_env_name ) => {
```

plocket: Might be worth considering allowing "the GitHub", as in "the GitHub secret", to allow people to give themselves and others some context.

```suggestion
When(/I set the var(?:iable)? "([^"]+)" to ?(?:the)? ?(?:GitHub)? secret "([^"]+)"/i, { timeout: -1 }, async ( var_name, set_to_env_name ) => {
```

lib/scope.js
```diff
@@ -22,6 +22,17 @@ let escape_regex = function( to_escape ) {
   return to_escape.replace(/[-\/\\^$*+?.()|[\]{}]/g, '\\$&');
 };
 
+let display_var_value = function(row) {
```
plocket: Maybe something with "printable" or "public" instead of "display"? I know we use "display" a lot in AL core, but I generally find it a bit confusing.

lib/steps.js
BryceStevenWilley: Ugh, didn't think about that. I made this change just because it's hard to get the URL right if you can't see it, especially if you're running on localhost (you need the `http://`, which isn't clear). It's probably okay to revert, this seemed hacky anyway.

lib/steps.js
BryceStevenWilley: I'm okay with adding `the`, but `Github` seems a bit to specific, and not that helpful? I guess allowing it is fine.

lib/scope.js
BryceStevenWilley: printable sounds good. I can change it to that.

lib/scope.js
BryceStevenWilley: Sounds like more bugs would be uncovered if your interview proceeds when entering `undefined` 👀 , but I do agree it makes more sense to throw too.

lib/scope.js
BryceStevenWilley: `original.value` isn't the secret's name, it's always the value. 

For the secrets, I could either keep `value` the same, and find everywhere it's printed and change that, or I could change `value` to the secret name, and change everywhere the value was set. I chose the first option, since I thought it was more risky to change the actual test behavior, possibly breaking things if I missed a line somewhere.

lib/scope.js
BryceStevenWilley: I don't think it is, because I didn't want to change the output to be not wrapped in `scope.toTrimmedJSON` (not sure what exactly it does), but I also didn't want the env var name display to be wrapped in that call either.

lib/scope.js
plocket: I see where this confusion is coming from now. It's from what the new function sends to `normalizeTable()`. That's a conundrum because `.original` should probably always be the original input to stay consistent with all the other types of rows. Would it make sense to put the logic in `normalizeTable()`? Add either add the flag as an `options` argument?

lib/scope.js
plocket: Just a note [incase this becomes the foundation of future conversations] that I think the bugs would end up being harder for the developer to track down. This will tell them right where to look for this particular mistake as opposed to triggering some other error further down the line.

lib/scope.js
plocket: Not sure we're talking about the same thing. I mean that `display_var_value()` already uses `is_secret_var(original)`.

BryceStevenWilley: If I changed it to `let value_str = display_var_value(original);`, it would return `original.value`, and not `scope.toTrimmedJSON(scope, { str: original.value });`.  I don't really know why we're JSON.stringify-ing it so I don't want to wrap the `'${' + secret_name + '}'` with it, but I didn't want to change that behavior in case there was some case I wasn't considering.

`display_var_value` is used in several other places okay, but it wouldn't quite work in this case.

lib/scope.js
BryceStevenWilley: Haha, I know. I was kinda joking, I agree throwing is the much clearer option.

lib/scope.js
BryceStevenWilley: I see. I guess this makes sense for laying the ground work for having env vars be in tables. But it'll make things a bit harder to make sure the right value is being set in every case. I'll take a shot at it now.

BryceStevenWilley: Works fine. Hides a bit of the env var work, but that's probably fine.

lib/scope.js
BryceStevenWilley: Now irrelevant because of changes from a different discussion: https://github.com/SuffolkLITLab/ALKiln/pull/536#discussion_r839677374, now we don't need to bother hiding the original value at all.

lib/scope.js
plocket: Just a note for future people: I still agree we necessarily want env vars to be in a story table, but we're still taking advantage of the table to set these variables and so we probably want to keep behavior consistent.

BryceStevenWilley:
Tests are failing because https://github.com/plocket/docassemble-ALAutomatedTestingTests/pull/185 needs to be merged.

tests/features/interactive_steps.feature
```diff
@@ -83,3 +83,20 @@ Scenario: I set text-type values
   And I tap to continue
   # Next page
   Then the question id should be "the end"
+
+@fast @i2
+Scenario: handles settings from Github secrets
+  Given I start the interview at "text_tests"
+  When I set the var first_text_entry to the secret "SECRET_VAR1"
```
plocket: I don't quite have bandwidth to review everything at the moment, and sorry I didn't see this before. This isn't passing, but not just because the file doesn't exist on the other server.  I think the var name is missing quotes:

[Edit: I made a full edit instead of these little partial ones, so removing suggestions from here.]

```
  When I set the var "first_text_entry" to the secret "SECRET_VAR1"
```

That might seem out of the blue, but I'm getting it from the tests. In the cucumber output I can see
```
   ? When I set the var first_text_entry to the secret "SECRET_VAR1"
       Undefined. Implement with the following snippet:
```
which means it's not recognizing the cucumber step you're using.

tests/features/interactive_steps.feature
```diff
@@ -83,3 +83,20 @@ Scenario: I set text-type values
   And I tap to continue
   # Next page
   Then the question id should be "the end"
+
+@fast @i2
+Scenario: handles settings from Github secrets
+  Given I start the interview at "text_tests"
```
plocket: This has to expect a failure. See `reports.feature` for examples:
https://github.com/SuffolkLITLab/ALKiln/blob/main/tests/features/reports.feature#L16

tests/features/interactive_steps.feature
```diff
@@ -83,3 +83,20 @@ Scenario: I set text-type values
...
+  When I tap to continue
+  And I set the var second_text_entry to the github secret "SECRET_VAR2"
+  And I tap to continue
+  Then I see the phrase "secret-var2-value"
+  When I tap to continue
+  And I set the variable third_text_entry to secret "SECRET_NOT_THERE"
+  Then the Scenario report should include:
+  """
+  Could not find "SECRET_NOT_THERE"
+  """
```
plocket: No Steps can happen after a failure, so this will have to go before the failing step.

tests/features/interactive_steps.feature
plocket: Forgot I could include multiple lines for a suggestion, duh.

[Edit: I made a full edit instead of these little partial ones, so removing suggestions from here. In which case we don't need this at all and I'll resolve it.]

```
Scenario: handles settings from Github secrets
  Given the final Scenario status should be "failed"
  Given I start the interview at "text_tests"
```

tests/features/interactive_steps.feature
plocket: I've been putting it at the start after the expected status, but it's flexible.

plocket: Expected status [position] is also flexible [as long as it's before the failing Step]. Also making me think we could probably combine expected report contents and expected status.

plocket: Why the heck am I not just offering the [full] solution. Sorry about that.

```suggestion
@fast @i2
Scenario: handles settings from Github secrets
  Given the final Scenario status should be "failed"
  And the Scenario report should include:
  """
  Could not find "SECRET_NOT_THERE"
  """
  Given I start the interview at "test_secret_vars"
  When I set the var "first_text_entry" to the secret "SECRET_VAR1"
  And I tap to continue
  Then I see the phrase "secret-var1-value"
  When I tap to continue
  And I set the var "second_text_entry" to the github secret "SECRET_VAR2"
  And I tap to continue
  Then I see the phrase "secret-var2-value"
  When I tap to continue
  And I set the variable "third_text_entry" to secret "SECRET_NOT_THERE"
```

plocket:
A couple more things I need to explore, but I'll need to work on it tomorrow.

lib/scope.js
```diff
@@ -1508,7 +1531,8 @@ module.exports = {
           let source_row = row;
           if ( row.source ) { source_row = row.source; }
           let original_row = source_row.original;
-          let msg = `Did not find a field on this page for the variable \`${ original_row.var }\` that could be set to \`${ original_row.value }\``;
+          let reported_val = printable_var_value(original_row); 
+          let msg = `Did not find a field on this page for the variable \`${ original_row.var }\` that could be set to \`${ reported_val }\``;
```
plocket: From what I can tell, this won't result in what you'd hope. E.g. for a row like this:
```
| some_var | SOME_SECRET |  |
```
Testing it out in the console, it resulted in "Did not find a field on this page for the variable `some_var` that could be set to `${SOME_SECRET}`". I'm also taking the opportunity to get rid of my own past inconsistency with those backticks unless someone objects.

```suggestion
          let msg = `Did not find a field on this page for the variable "${ original_row.var }" that could be set to "${ original_row.value }"`;
```

plocket: The same situation shows up elsewhere and I can grab those too if we determine that I'm not missing something.

lib/steps.js
```diff
@@ -14,12 +14,12 @@ const wrapPromiseWithTimeout = require('./cucumber_8_shim');
 const fs = require('fs');
 const path = require('path');
 const { expect } = require('chai');
-const puppeteer = require('puppeteer');
 const { v4: uuidv4 } = require('uuid');
 
 // Ours
 const scope = require('./scope');
 const session_vars = require('./utils/session_vars');
+const get_base_interview_url = require(`./docassemble/get_base_interview_url`);
```
plocket: I think this is not being used anymore.

```suggestion
```

lib/scope.js
```diff
@@ -467,27 +478,39 @@ module.exports = {
...
     let supported_table = [];
     for ( let row of var_data ) {
       // See tests > unit_tests > tables.fixtures.js > tables.old_to_current_formatting
+      let actual_var_value = row.value;
+      if (is_secret_var(row)) {
+        actual_var_value = process.env[row.flags.secret_name];
+        if ( actual_var_value === undefined ) {
+          let not_in_env_msg = `The GitHub SECRET "${ row.flags.secret_name }" doesn't exist. Check for a typo and check your workflow file.`;
+          await scope.addToReport( scope, { type: `error`, value: not_in_env_msg })
+          throw ReferenceError(`Could not find "${row.flags.secret_name }": "${not_in_env_msg}".`)
```
plocket: This one's not super crucial either. I think some extra quotes may have snuck in, but it's hard to read this kind of stuff and keep it all in mind, so maybe I'm wrong.

```suggestion
          throw ReferenceError(`Could not find "${row.flags.secret_name }": ${not_in_env_msg}.`)
```

plocket (approved):
The tests are passing and this'll be a great improvement. I think because of the ways responsibilities shifted around between functions, there's a bit more to clean up. I'll do that and if it's acceptable, we can go ahead and merge.

lib/scope.js
BryceStevenWilley: I don't think it matters too much. The backticks are shown in console, and IMO backtics represent any kind of verbatim code. But I'll take the suggestions anyway.

plocket (approved):
Looks great and a great addition! Thanks for helping work out the kinks.

PR #538, opened 2022-03-29
526 change logfile to debug_log.txt v4

### LINKED ISSUE #526: Change log.txt to debug_log.txt

mcdonaldcarolyn:


.github/workflows/test_tests.yml
```diff
@@ -75,12 +75,12 @@ jobs:
         with:
           name: report
           path: ./**/*_report_*
-      - name: log artifact
+      - name: debug_log artifact
```
plocket: A possible improvement could be changing the name to something like "upload debug_log". A bit more friendly to humans who don't know what "artifact" means in this context.

.gitignore
```diff
@@ -128,4 +128,4 @@ _al_test_project_name*
 # So that our tests will behave more like our users' tests
 package-lock.json
 #report logs
-/log.txt
+debug_log.txt
```
plocket: We do still want the slash at the start.

plocket (approved):
Thank you for the improvement! I left two possible improvements that we can do in future if needed. If you want to do them now, let me know when you're done and I'll review again.

Also, if you could link this PR to issue #526, that would be useful.

Thank you!

.github/workflows/test_tests.yml
mcdonaldcarolyn: great idea!

.gitignore
mcdonaldcarolyn: ok, got it. Thank you for the helpful feedback!

log.txt
```diff
@@ -0,0 +1,984 @@
+Report:
```
plocket: Just need to delete this file and then I think we're good!

plocket (changes requested):
After the next comment is resolved I think we're good to go!

plocket:
Oh, also link #526 if you have the time.

mcdonaldcarolyn:
Closes #526

plocket (approved):
Looks great! Thanks for fine-tuning!

PR #540, opened 2022-03-31
Append cucumber report

### LINKED ISSUE #491: Logs: show cucumber (and other?) error/all messages
Not sure about the cucumber error stacktrace, but can we save a bit of pain by adding cucumber's own [output] ~scenario failure messages~ to the report. We're already doing a lot of work with the scenario object. I bet the message is in there somewhere...

[This would be useful to add to the reports so the user and we can see it later without being on GitHub.]

[From #398, put cucumber output in report (which we've changed to putting in debug_log.txt now):

> I think that err will appear in the cucumber output, though I do wish we had a better way to test that.
> 
> In our convo, we brought up that the cucumber output may be at times useful for a different set of users, so maybe that should be included at the bottom of the report file somehow.
> 
> Also, the pros and cons of having the both more detailed and less detailed cucumber output in place of the report. Cucumber gives output such as expected x to exist, but it did not and then the stack trace. That isn't usually helpful for people trying to understand why their interview failed. The stack trace just traces the ALKiln code and is a red herring, and the failure reason is often not specific enough about what went wrong with the interview itself.
> 
]

BryceStevenWilley:
Fixes #491.

* adds a cucumber.json to the root dir and to the action.yml
  This makes cucumber print multiple reports. See:
  * https://github.com/cucumber/cucumber-js/blob/main/docs/configuration.md
  * https://github.com/cucumber/cucumber-js/blob/main/docs/formatters.md
* adds some bash commands to the npm script to concatenate the summary report to the
  end of the alkiln report. To keep the best bash experience, we still print with colors,
  and then strip out the colors when concatenating.
* added .vscode and the extra cucumber reports to the gitignore

### DIFF
--- a/.github/workflows/test_tests.yml
+++ b/.github/workflows/test_tests.yml
@@ -48,6 +48,7 @@ jobs:
       - run: npm install
       - run: npm run setup
       - run: npm run test -- ${{ github.event.inputs.tags && format('"--tags" "{0}"', github.event.inputs.tags) }}
+        shell: bash
       - name: upload error artifacts
         uses: actions/upload-artifact@v2
         if: ${{ failure() }}
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@ lerna-debug.log*
 
 # Diagnostic reports (https://nodejs.org/api/report.html)
 report.[0-9]*.[0-9]*.[0-9]*.[0-9]*.json
+cucumber-report.*
 
 # Runtime data
 pids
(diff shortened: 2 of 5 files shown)


BryceStevenWilley:
I have discovered that [node always defaults to using `sh` on unix,](https://github.com/npm/feedback/discussions/115#discussioncomment-210601), and the solution I had come up with was using `bash`. Since there was feedback that the concatenation should be run in the `package.json`, we can't rely on just adding `shell: bash` to all parts of the github aciton. Additionally, I realized that it'll probably [exclude windows users](https://stackoverflow.com/questions/34937724/running-bash-scripts-with-npm#comment57607985_34937724).  A lil frustrating, but it's probably worth just writing a separate JS script that is called directly after the cucumber script, so I'll close this and retry with that option at some point.

PR #544, opened 2022-04-01
Append cucumber report pt2

### LINKED ISSUE #491: Logs: show cucumber (and other?) error/all messages

BryceStevenWilley:
A second attempt of #540, to fix #491.

Adds a JS script that concatenates the printed out cucumber report to the end of the Kiln report, stripping out any terminal specific color commands.

Feel free to critique my JS and give improvements, this was mostly from following the tutorials on [Node.js](https://nodejs.dev/learn).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -131,3 +131,4 @@ _al_test_project_name*
 package-lock.json
 #report logs
 /debug_log.txt
+/log.txt
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@ lerna-debug.log*
 
 # Diagnostic reports (https://nodejs.org/api/report.html)
 report.[0-9]*.[0-9]*.[0-9]*.[0-9]*.json
+cucumber-report.*
 
 # Runtime data
 pids
@@ -117,6 +118,7 @@ dist
 tests/chromedriver
 *DS_Store*
 node_modules*
+.vscode/**
 .env
 npm-debug.log
 error*.jpg
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "main": "lib/index.js",
   "scripts": {
     "test": "npm run unit && npm run langs_setup && npm run cucumber -- $@",
-    "cucumber_base": "run_cucumber(){ ./node_modules/.bin/cucumber-js \"$@\"; ./lib/utils/clean_reports.js; }; run_cucumber",
+    "cucumber_base": "run_cucumber(){ ./node_modules/.bin/cucumber-js \"$@\"; status=$?; ./lib/utils/clean_reports.js; return $status; }; run_cucumber",
     "cucumber": "run_base(){ npm run cucumber_base -- \"$@\" tests/features/*.feature; }; run_base",
     "langs": "npm run langs_setup && ./node_modules/.bin/cucumber-js --require ./lib/index.js tests/features/*_lang*.feature",
     "langs_setup": "./lib/utils/langs.js 'tests/features/*.feature'",
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "main": "lib/index.js",
   "scripts": {
     "test": "npm run unit && npm run langs_setup && npm run cucumber -- $@",
-    "cucumber_base": "./node_modules/.bin/cucumber-js --require ./lib/index.js",
+    "cucumber_base": "./node_modules/.bin/cucumber-js && ./lib/utils/clean_reports.js",
     "cucumber": "run_cucumber(){ npm run cucumber_base -- \"$@\" tests/features/*.feature; }; run_cucumber",
     "langs": "npm run langs_setup && ./node_modules/.bin/cucumber-js --require ./lib/index.js tests/features/*_lang*.feature",
     "langs_setup": "./lib/utils/langs.js 'tests/features/*.feature'",
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "scripts": {
     "test": "npm run unit && npm run langs_setup && npm run cucumber -- $@",
     "cucumber_base": "run_cucumber(){ ./node_modules/.bin/cucumber-js \"$@\"; ./lib/utils/clean_reports.js; }; run_cucumber",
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -5,7 +5,7 @@
   "main": "lib/index.js",
   "scripts": {
     "test": "npm run unit && npm run langs_setup && npm run cucumber -- $@",
-    "cucumber_base": "./node_modules/.bin/cucumber-js --require ./lib/index.js",
+    "cucumber_base": "run_cucumber(){ ./node_modules/.bin/cucumber-js \"$@\"; ./lib/utils/clean_reports.js; }; run_cucumber",
```
plocket: I'm not completely sure how scope works in bash. Can we give this function another name? It's also ok if we change the name of the other function instead. For that reason and because it's strange to me that they're both "running cucumber". They're not doing the same thing, so what are they each doing?

plocket: Your code makes me wonder if we really need the `--` in the other function. Would appreciate your thoughts on that.

plocket (commented):
The tests are passing and I'm not completely sure how to test their functionality. I'm not seeing the cucumber part of the report inside the report file on GitHub.

I have some very bad guesses. For example, the async nature of cucumber.js (which we might call cucumber-report.js?). Not sure the node command line is waiting for it. That could have worked locally because it should have finished executing by the time a human person checks the local file, but might not work on GitHub, which closes everything up. I believe it's possible to do all those operations synchronously, so that might be something to try, though I admit I prefer knowing the source of a problem for sure.

I'm also trying to read up on `readdirSync` to see if it can use relative paths. I've mostly seen people use `__dirname`. I suspect that's not the case or you would have noticed it breaking locally.

If you already know the answers to some of these, then it'd be helpful in saving some time. This also may be a more productive discussion in real time.

package.json
BryceStevenWilley: I think it's less about bash and more about how npm is actually running it. I assume it just copy pastes everything in the string directly to shell, which does overwrite it with the latest line. A quick example:
 ```bash
$ run_hi() { echo 'hi 1'; }
$ run_hi
hi 1
$ run_hi() { echo 'hi 2'; }
$ run_hi
hi 2
 ```
 Not opposed to changing the name though, I just needed to wrap everything in some function to make sure `$@` was stable. (might not be necessary, but idk for sure).`
 
 > Your code makes me wonder if we really need the -- in the other function
 
TBH, I don't know what exactly `--` does in this case, it's very program dependent, and I can't find what it does for cucumber at all.

BryceStevenWilley:
I do think it's working! The last lines of the report that I can see are:
> 54 scenarios (54 passed)
> 313 steps (313 passed)
> 5m15.634s (executing steps: 5m15.349s)

Which is what is printed by the cucumber summary process when nothing fails. Compare that with [the latest merge on releases/v4](https://github.com/SuffolkLITLab/ALKiln/actions/runs/2080517700), which doesn't have that summary of scenarios/steps passed and the time it took everything.

plocket:
Ah, my report file wasn't loading all the way somehow in finder! Looks great!

plocket:
We want it in the log, though, right? Not the report?

plocket:
That is, when cucumber errors get triggered, we discussed having them in the logs, not in the report.

BryceStevenWilley:
Apologies, that fact got a bit lost in trying to figure all of the mess, but it's a quick fix. Waiting for the tests to confirm we're adding to the `debug_log.txt` now.

plocket:
debug_log.txt is confirmed 👍 [And sorry I lost track of a couple things as well. Such is the way the cookie crumbles sometimes.]

plocket:
Newest wrinkle: I ran a deliberately failing tests. The tests failed correctly, but the run passed for some reason: https://github.com/SuffolkLITLab/ALKiln/runs/5835449414?check_suite_focus=true

BryceStevenWilley:
Good catch, I wasn't sure how to test that on GitHub, and bash exit statuses are always silent and hard to find for me.

The issue was that `clean_report.js` is the last thing to run, and it should always succeed, so it overwrites the bash exit status for cucumber, which is the thing that says whether or not the task failed (downside of having clean_report.js run in npm, and not as a part of the GH action). Fix was to manually save the exit status from cucumber, and just return that instead.

BryceStevenWilley:
Confirmed that manually making the workflow fail does make the tests fail! https://github.com/SuffolkLITLab/ALKiln/actions/runs/2098401294

.gitignore
```diff
@@ -129,3 +131,4 @@ _al_test_project_name*
 package-lock.json
 #report logs
 /debug_log.txt
+/log.txt
```
plocket, line 134: Sorry, at this point log stuff needs to be removed. I didn't catch anything elsewhere, but this can be cleaned up.

```suggestion
/debug_log.txt
```

plocket (approved):
So close and ever closer! Just a little cleanup and then this is a great addition to output! It's a non-breaking cleanup, so I'll approve and you can remove or not and merge as you see fit.

.gitignore
BryceStevenWilley: This was intentional, I had rebased and accudentally commited my log.txt to the repo. I think that others will likely make the same mistake, and IMO it's doesn't need to be removed from gitignore anyway.

plocket: I think there are only three of us with the problem and we've all dealt with it, but it won't do any harm. Sound good then 👍

PR #545, opened 2022-04-01
Check all types of rows for is_secret_var too

plocket:
Sorry I missed this before. Let me know if this seems unnecessary.


plocket:
We could abstract this process, but it would be an extra delay.

plocket:
Eh, I suppose I could abstract it right now by just editing.

plocket:
I can see using that abstraction elsewhere, even though what I really want to do is flatten the row data.

PR #548, opened 2022-04-05
pin versions of dependencies in action.yml and package.json

### LINKED ISSUE #518: Freeze all library versions at patches in package.yml and action.yml (v4)
For the developer's workflow, we won't be freezing the developer's version of our own action (created through the setup interview) because we've decided that's too confusing for most users.

mcdonaldcarolyn:
Closes #518

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,7 +38,10 @@ Format:
 ### Security
 - 
 -->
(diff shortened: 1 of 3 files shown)


plocket (approved):
Looks great! Thanks! Go ahead and close the issue.

PR #556, opened 2022-04-19
Tap element2

BryceStevenWilley:
Found a decent solution to tapping tabbed templates, and for tapping arbitrary elements. Not urgent, so leaving this in drafts until I can iterate on it for a bit longer.

### DIFF
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -326,8 +326,6 @@ module.exports = {
     let start_url = await scope.page.url();  // so we can later check for navigation
 
     let clickWait;
-    // Error loads page, so no need to detect to keep things short
-    // Click with no navigation will end immediately
     // Can't use elem[ scope.activate ](). It works locally, but errors on GitHub
     // with "Node is not visible or not an HTMLElement". We're not using a custom Chromium version
     // Unfortunately, this won't catch an invisible element. Hopefully we'd catch it before now...?
@@ -336,6 +334,7 @@ module.exports = {
     // and other convos around it. We haven't made it public that the device
     // can customized. Until we do that, we'll just use "click" all the time.
     if (!tabToWaitFor) {
+      // Click with no navigation will end immediately
       clickWait = Promise.all([
         elem.evaluate( (el) => { return el.click(); }),
         scope.page.waitForNavigation({ waitUntil: 'domcontentloaded' }),
@@ -344,6 +343,7 @@ module.exports = {
       // Same thing as above, but the promise doesn't wait for navigation.
       clickWait = elem.evaluate( (el) => {return el.click(); });
     }
+    // Error loads page, so no need to detect to keep things short
     let winner = await Promise.race([
       clickWait,
       // This is meant to detect user-visible alerts/errors,
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -420,6 +420,15 @@ module.exports = {
     await scope.tapElement( scope, elem );
   },
 
+  tapElementId: async function ( scope, element_id ) {
+    let elem = await scope.page.waitForSelector(`#${element_id}`);
+    console.log(`about to tap elem ${ element_id }`)
+    await elem.evaluate( (el) => { return el.click(); });
+    // A bit hacky, but there's not a better way to handle it TBH. We don't know the next step
+    // (fails if "sees phrase", and what you need to do is to wait for a selector, which we can't really do unless we know what's gonna show up, but we don't get info about that.
+    await scope.page.waitForTimeout(2000);
+  },
+
   load: async function ( scope, file_name ) {
     /* Try to load the page. Should we also pass a `timeout` arg or just
     *    use scope.timeout? */
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -244,7 +244,15 @@ module.exports = {
     if ( showif ) { await scope.page.waitForTimeout( scope.showif_timeout ); }
   },
 
-  afterStep: async function afterStep(scope, options = {waitForShowIf: false, navigated: false, waitForTimeout: 0}) {
+  waitForTab: async function waitForTab(scope, tab_to_show) {
+    // Implementation linked to https://github.com/SuffolkLITLab/docassemble-ALToolbox/blob/4aa1069389d2ca9cadb586f309dd73d791f081d1/docassemble/ALToolbox/misc.py#L127
+    // I think `.$()` does not use a timeout
+    if ( tab_to_show.length == 0) { return }
+    let tab = await scope.page.$(`div[aria-labelledby="${tab_to_show}"`);
+    if ( tab ) { await scope.page.waitForTimeout( scope.showif_timeout ); }
+  },
+
+  afterStep: async function afterStep(scope, options = {waitForShowIf: false, waitForTab: '', navigated: false, waitForTimeout: 0}) {
     /* Common things to do after a step, including take care of errors.
     *    TODO: discuss splitting into `afterField` and `afterStep` or something. Or just change name
     *    TODO: Update to latest cucumberjs, which as `AfterStep` */
@@ -274,6 +282,7 @@ module.exports = {
 
       // Wait for elements that might get revealed
       if ( options.waitForShowIf ) { await scope.waitForShowIf( scope ); }
+      if ( options.waitForTab ) { await scope.waitForTab( scope, options.waitForTab ); }
     }  // ends if scope.page
 
     // Reset navigation (TODO: discuss use for observational steps)
@@ -284,7 +293,7 @@ module.exports = {
     if ( options.waitForTimeout ) { await time.waitForTimeout( options.waitForTimeout ); }
   },  // Ends afterStep
 
-  tapElement: async function tapElement(scope, elem) {
+  tapElement: async function tapElement(scope, elem, isTab=false, waitForTimeout=0) {
     /* Tap a given element, detect navigation, and do appropriate related things. */
 
     // This check might never be needed now as we try to match fields on the page
@@ -329,8 +338,12 @@ module.exports = {
       // If navigation, wait for 200ms to let things settle down
       await scope.afterStep(scope, {waitForTimeout: 200, navigated: navigated});
     } else {
-      // If stay on page, wait for `show if` or feedback elements to be move around
-      await scope.afterStep(scope, {waitForShowIf: true, navigated: navigated});
+      // If stay on page, wait for `show if`, the new tab to be loaded, or feedback elements to be move around
+      options = {waitForShowIf: !isTab, waitForTab: isTab, navigated: navigated};
+      if (waitForTimeout > 0) {
+        options = {waitForTimeout: waitForTimeout, navigated: navigated};
+      }
+      await scope.afterStep(scope, options); 
     }
 
     return winner;
@@ -420,13 +433,14 @@ module.exports = {
     await scope.tapElement( scope, elem );
   },
 
-  tapElementId: async function ( scope, element_id ) {
-    let elem = await scope.page.waitForSelector(`#${element_id}`);
-    console.log(`about to tap elem ${ element_id }`)
-    await elem.evaluate( (el) => { return el.click(); });
-    // A bit hacky, but there's not a better way to handle it TBH. We don't know the next step
-    // (fails if "sees phrase", and what you need to do is to wait for a selector, which we can't really do unless we know what's gonna show up, but we don't get info about that.
-    await scope.page.waitForTimeout(2000);
+  tapTab: async function ( scope, tab_id ) {
+    let elem = await scope.page.waitForSelector(`#${tab_id}`);
+    await scope.tapElement( scope, elem, isTab=true );
+  },
+
+  tapElementById: async function ( scope, selector, waitForTimeout ) {
+    let elem = await scope.page.waitForSelector(`${selector}`);
+    await scope.tapElement (scope, elem, waitForTimeout );
   },
 
   load: async function ( scope, file_name ) {
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -460,13 +460,23 @@ module.exports = {
   },
 
   tapTab: async function ( scope, tab_id ) {
-    let elem = await scope.page.waitForSelector(`#${tab_id}`);
+    let elem = await scope.page.$(`#${tab_id}`); 
+    let error_msg = `Couldn't find the tab with id "#${tab_id}" on the page, is there a typo?`;
+    if ( !elem ) {
+      await scope.addToReport(scope, { type: `error`, value: error_msg }); 
+    }
+    expect( elem, error_msg ).to.exist;
     await scope.tapElement( scope, elem, tabToWaitFor=tab_id );
   },
 
   tapElementBySelector: async function ( scope, selector, waitForTimeout ) {
-    let elem = await scope.page.waitForSelector(`${selector}`);
-    await scope.tapElement (scope, elem, false, waitForTimeout );
+    let elem = await scope.page.$(`${selector}`);
+    let error_msg = `Couldn't find "${selector}" on the page, is there a typo?`;
+    if ( !elem ) {
+      await scope.addToReport(scope, { type: `error`, value: error_msg });
+    }
+    expect( elem, error_msg ).to.exist;
+    await scope.tapElement( scope, elem, '', waitForTimeout );
   },
 
   load: async function ( scope, file_name ) {
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -249,13 +249,14 @@ module.exports = {
     // I think `.$()` does not use a timeout
     if ( tab_to_show.length == 0) { return }
     let tab = await scope.page.$(`div[aria-labelledby="${tab_to_show}"`);
-    if ( tab ) { await scope.page.waitForTimeout( scope.showif_timeout ); }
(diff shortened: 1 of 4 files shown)


BryceStevenWilley:
I moved over some fixes from the accessibility branch that I had working on recently, but this should be ready for review. I'll make sure the tests are passing before pinging anyone there.

lib/scope.js
```diff
@@ -244,10 +244,19 @@ module.exports = {
     if ( showif ) { await scope.page.waitForTimeout( scope.showif_timeout ); }
   },
 
-  afterStep: async function afterStep(scope, options = {waitForShowIf: false, navigated: false, waitForTimeout: 0}) {
+  waitForTab: async function waitForTab(scope, tab_to_show) {
+    // Implementation linked to https://github.com/SuffolkLITLab/docassemble-ALToolbox/blob/4aa1069389d2ca9cadb586f309dd73d791f081d1/docassemble/ALToolbox/misc.py#L127
+    // I think `.$()` does not use a timeout
```
plocket, line 253: I think you're right.

lib/scope.js
```diff
@@ -244,10 +244,19 @@ module.exports = {
     if ( showif ) { await scope.page.waitForTimeout( scope.showif_timeout ); }
   },
 
-  afterStep: async function afterStep(scope, options = {waitForShowIf: false, navigated: false, waitForTimeout: 0}) {
+  waitForTab: async function waitForTab(scope, tab_to_show) {
```
plocket, line 247: I've been making the second argument an object. That does complicate factors in some way, but it may be worth talking about at some point. Could you add a description of what the expected value is?

plocket: Or maybe it just could have a different name, like `tab_id`.

lib/scope.js
```diff
@@ -244,10 +244,19 @@ module.exports = {
...
+    let tab = await scope.page.$(`div[aria-labelledby="${tab_to_show}"`);
+    // TODO(brycew): consider waiting for the transitionend event: https://stackoverflow.com/a/57955251/11416267
+    if ( tab ) { await scope.page.waitForTimeout( scope.showif_timeout * 2 ); }
+  },
+
+  afterStep: async function afterStep(scope, options = {waitForShowIf: false, waitForTab: '', navigated: false, waitForTimeout: 0}) {
     /* Common things to do after a step, including take care of errors.
     *    TODO: discuss splitting into `afterField` and `afterStep` or something. Or just change name
-    *    TODO: Update to latest cucumberjs, which as `AfterStep` */
+    *    TODO: Update to latest cucumberjs, which has `AfterStep` */
```
plocket, line 277: This [already exists in v7 of cucumber](https://github.com/cucumber/cucumber-js/blob/7.x/docs/support_files/api_reference.md), a version which we use, we just haven't made the conversion yet. That said, if I recall correctly some of this doesn't actually belong after every step (hence the various arguments) and we have to think more carefully about how to handle this functionality.

lib/scope.js
```diff
@@ -244,10 +244,19 @@ module.exports = {
...
+  waitForTab: async function waitForTab(scope, tab_to_show) {
+    // Implementation linked to https://github.com/SuffolkLITLab/docassemble-ALToolbox/blob/4aa1069389d2ca9cadb586f309dd73d791f081d1/docassemble/ALToolbox/misc.py#L127
+    // I think `.$()` does not use a timeout
+    if ( tab_to_show.length == 0) { return }
+    let tab = await scope.page.$(`div[aria-labelledby="${tab_to_show}"`);
+    // TODO(brycew): consider waiting for the transitionend event: https://stackoverflow.com/a/57955251/11416267
+    if ( tab ) { await scope.page.waitForTimeout( scope.showif_timeout * 2 ); }
+  },
+
+  afterStep: async function afterStep(scope, options = {waitForShowIf: false, waitForTab: '', navigated: false, waitForTimeout: 0}) {
```
plocket: Just noting that here too `waitForTab` might be `tab_id` or `tab_id_to_wait_for` or `tab_to_wait_for`. `waitForTab` currently sounds like a boolean to me. That said, this isn't a blocking change.

lib/scope.js
```diff
@@ -297,25 +307,34 @@ module.exports = {
...
+    let inner;
+    // Error loads page, so no need to detect to keep things short
+    // Click with no navigation will end immediately
+    // Can't use elem[ scope.activate ](). It works locally, but errors on GitHub
+    // with "Node is not visible or not an HTMLElement". We're not using a custom Chromium version
+    // Unfortunately, this won't catch an invisible element. Hopefully we'd catch it before now...?
+    // "Tap" on mobile is more complex to implement ourselves.
+    // See https://stackoverflow.com/a/56547605/14144258
+    // and other convos around it. We haven't made it public that the device
+    // can customized. Until we do that, we'll just use "click" all the time.
```
plocket: Is this a good place for these comments? I'm not sure the previous place was the best place for all of these, but I think it may have been slightly better.

lib/scope.js
```diff
@@ -297,25 +307,34 @@ module.exports = {
...
-      Promise.all([  // Error loads page, so no need to detect to keep things short
-        // Click with no navigation will end immediately
-        // Can't use elem[ scope.activate ](). It works locally, but errors on GitHub
-        // with "Node is not visible or not an HTMLElement". We're not using a custom Chromium version
-        // Unfortunately, this won't catch an invisible element. Hopefully we'd catch it before now...?
-        // "Tap" on mobile is more complex to implement ourselves.
-        // See https://stackoverflow.com/a/56547605/14144258
-        // and other convos around it. We haven't made it public that the device
-        // can customized. Until we do that, we'll just use "click" all the time.
+    let inner;
```
plocket: Just personal confusion experience - what is the name `inner` meant to describe?

lib/scope.js
```diff
@@ -420,6 +443,16 @@ module.exports = {
     await scope.tapElement( scope, elem );
   },
 
+  tapTab: async function ( scope, tab_id ) {
+    let elem = await scope.page.waitForSelector(`#${tab_id}`);
+    await scope.tapElement( scope, elem, isTab=true );
+  },
+
+  tapElementById: async function ( scope, selector, waitForTimeout ) {
```
plocket: Might change this name to `tabElementBySelector` since it really can be any selector, right? It would also let people know that they have to include `#` before the selector if they want to us an id - this isn't going to add it automatically.

lib/steps.js
```diff
@@ -390,7 +390,7 @@ When(/^(?:the user gets)?(?:I get)? to ?(?:the)?(?: ?question)?(?: id)? "([^"]+)
 //#####################################
 //#####################################
 
-When(/I wait (\d*\.?\d+) seconds?/, { timeout: -1 }, async ( seconds ) => {
+When(/I wait ?(?:for)? (\d*\.?\d+) seconds?/, { timeout: -1 }, async ( seconds ) => {
```
plocket, line 393: I like it!

lib/steps.js
```diff
@@ -706,6 +706,27 @@ When('I tap to continue', { timeout: -1 }, async () => {
...
+When(/I tap the "([^"]+)" tab/i, { timeout: -1 }, async ( tabId ) => {
+  return wrapPromiseWithTimeout(
+    scope.tapTab( scope, tabId ),
+    scope.timeout
+  )
+});
+
+When(/I tap the "([^"]+)" element$/i, { timeout: -1 }, async ( elemSelector ) => {
+  return wrapPromiseWithTimeout(
+    scope.tapElementById( scope, elemSelector, 2000 ),
```
plocket: Why have a hard-coded timeout? Does it make sense to use the timeout the user can set themselves? That is, `scope.timeout`?

lib/scope.js
```diff
@@ -297,25 +307,34 @@ module.exports = {
...
+      inner = Promise.all([
         elem.evaluate( (el) => { return el.click(); }),
         scope.page.waitForNavigation({ waitUntil: 'domcontentloaded' }),
-      ]),
+      ]);
+    } else {
+      // Same thing as above, but the promise doesn't wait for navigation.
+      inner = Promise.all([
+        elem.evaluate( (el) => {return el.click(); }),
+      ]);
```
plocket: Just a style note - `.evaluate()` returns a promise, so you shouldn't need to use `Promise.all()` here.

lib/scope.js
plocket: Also, some of this should be done while the story table runs between pages, etc, and a story table is only one Step, so we'll have to manage that manually anyway.

Not easy stuff to know about this system. I'm not yet sure how to clarify it, but a name change would probably help.

plocket (approved):
Great addition! None of these comments are blocking changes, so I'll approve, but maybe we can make some into issues. I think the two main concerns that I noted are:

First, the hard-coded timeout in the `tap "..." element`. Even though there's a way to have a custom timeout, I still don't see why we'd have a hidden hard-coded timeout.

Second, the name of `tapElementById`, which I think is a misleading name.

Thanks for all the work!

plocket:
Actually, is there a way to test any of these at the moment? [We can make that into an issue for the future, but we do want to at least note that.]

Edit: Also, let me know if you have bandwidth to make an [issue in the documentation repo](https://github.com/suffolkLITLab/docassemble-assemblyline-documentation/issues) for the various new steps.

lib/steps.js
BryceStevenWilley: That's a separately defined step (that's just a different interface to this same code) that lets someone specify a timeout. IMO I didn't want to specify the timeout for every single step that I used this tap feature in, and unfortunately we do still need the time out. It's for browser events, so I can't imagine it being too flaky, since it's not based on server events. Ideally, the author / running shouldn't have to know that this timeout exists, so I don't want to make it super configurable elsewhere.

I looked into possibly using the [transition to end](https://stackoverflow.com/questions/51547994/wait-for-transition-to-end-in-puppeteer) value, but I couldn't actually get it to work for the one example I had (tapping on tabs in the tabbed interface), and I think what the actual transition that we should wait for will vary and I can't actually find?

BryceStevenWilley:
> Actually, is there a way to test any of these at the moment?

You're right, there's nothing stopping that from being added to the test repo (other than the friction of doing so). I'll do that and ping back to this PR when it's ready.

Can also add things to the documentation repo too, just wanted a confirmation that this is a wanted feature for the main repo.

lib/steps.js
plocket: I'm willing to go with that choice.

Regarding catching transition events, you can go down the rabbit hole if you want, but it depends on the implementation being used and, yes, it would be specific to the situation. Also, tapping some things may not cause transitions, etc. It's pretty complicated even before you hit the difficulties in the code itself.

lib/scope.js
BryceStevenWilley: This is straight copied from the above function, I believe you wrote this, haha.

lib/scope.js
BryceStevenWilley: So it is the tab id, but not being used to get the tab, it's the labelledby text on the tab panel (where the contents are actually shown, not the item you click to show it, which is the tab). I'll add this description, don't think there's a better name that's not extremely long and unwieldy though.

I'm going to leave the arguments as is too, I don't think it needs that many options right now.

lib/scope.js
BryceStevenWilley: It was a lazy name: it was the inner promise of the promise race below. I've changed it to be a bit more clear.

lib/scope.js
BryceStevenWilley: It's in almost the same place? It was inside the promise.all below, but I had to duplicate that call, so it can't (and shouldn't) stay in the exact same place because it applies both to lines 321 and 326.

lib/scope.js
BryceStevenWilley: This function name isn't something that authors will see, but I like the new name anyway. Changed.

lib/scope.js
BryceStevenWilley: A lot of that is out of scope for this PR: I just corrected a typo when I saw it.

lib/steps.js
BryceStevenWilley: I ran into some problems when testing this to make sure it worked, and ended up trying the transition events and it seems to work pretty well? I want to see if it'll work for the accessibility testing before merging (I remembered that the main reason I needed the wait was because I would tap the button to switch the tabs and then check for accessibility, but the page wouldn't have fully transitioned yet, so the accessibility check would check the colors with the in-progress fade-in. Those colors didn't have enough contrast, but also weren't the true colors of the page and so would be a flaky failure), but it's pretty smooth right now.

BryceStevenWilley:
Test interview file in https://github.com/plocket/docassemble-ALAutomatedTestingTests/pull/187. Tested locally and passing 100%.

lib/steps.js
BryceStevenWilley: Accessibility seems to work fine with the transition end stuff as well. I think it should be good.

lib/scope.js
plocket: It's a tad complex, so we can probably leave it as is for now. Here are thoughts to maybe make into an issue for future clean up.

Line 311: Think it goes under Promise.race, since it explains (I think?) why we're not racing to detect an invalid input error.

Line 312: Probably goes with the Promise that clicks and navigates.

Lines 313-319: not sure where they should go, so we can leave them here, but they should go somewhere else someday if we can keep an eye out for that. Maybe in our own repo's docs.

lib/scope.js
plocket: I think maybe you didn't push that change.

lib/steps.js
plocket: I'll have to see the code to see if it's general enough. To use it here, this would have to be able to detect the transition of tapping any element, not just a tab. If you want to just detect tab tapping transitions, that should probably be a separate Step just for tabs.

plocket: Somehow my page wasn't updating with new code. I'll check it out now and see how it all flows.

lib/scope.js
plocket: Page hadn't updated, looks good to me!

lib/scope.js
BryceStevenWilley: I moved those first two lines to be in the right places. I think the other lines are fine where they are

tests/features/interactive_steps.feature
```diff
@@ -101,3 +101,15 @@ Scenario: handles settings from Github secrets
...
+@i3 @tap-elements @tabs
+Scenario: tap elements
+  Given I start the interview at "test_taps"
+  And I tap the "Tests-first_template-tab" tab
+  Then I see the phrase "Mechanics"
+  And I tap the "Tests-second_template-tab" tab
+  Then I see the phrase "villify"
+  And I tap the "Tests-third_template-tab" tab
+  Then I see the phrase "museum"
+  And I tap the "#special_event" element and wait 5 seconds
```
plocket: Future issue: Would waiting 1 second be enough? I don't see a need for 5 and our tests are going to continue growing. I'd rather keep the time we add manually to a minimum.

Doesn't need to block this PR.

lib/scope.js
```diff
@@ -420,13 +459,23 @@ module.exports = {
...
+  },
+
   load: async function ( scope, file_name ) {
     /* Try to load the page. Should we also pass a `timeout` arg or just
     *    use scope.timeout? */
 
     // If there is no browser open, start a new one
     if (!scope.browser) {
-      scope.browser = await scope.driver.launch({ headless: !session_vars.get_debug() });
+      scope.browser = await scope.driver.launch({ headless: !session_vars.get_debug(), devtools: session_vars.get_debug() });
```
plocket, line 488: Nice find!

BryceStevenWilley: Thanks! I don't think it's necessary as a comment, but I found it from trying to debug some of the puppeteer `evaluate` functions: https://stackoverflow.com/a/49887254/11416267.

tests/features/interactive_steps.feature
BryceStevenWilley: I thought I had a reason for 5, but 1 second seems to work fine locally at least. I changed it to 1 (and added another fix to the step regex to allow for grammatically correct steps), hopefully it's not too flaky.

plocket:
Future: I think one remaining improvement we could make in the future is to better handle when the tab doesn't exist. For example, for when the user has a typo.

lib/scope.js
plocket: Just under that I see a link to https://github.com/puppeteer/puppeteer#debugging-tips with more tips.

BryceStevenWilley:
> Future: I think one remaining improvement we could make in the future is to better handle when the tab doesn't exist. For example, for when the user has a typo.

How you you suggest handling it? `waitForSelector` just waits for `scope.timeout` seconds and if it's not there returns null, which will fail later. Would it just be an additional check for null afterwards and a better log message? If that's all, I'll do it now.

https://puppeteer.github.io/puppeteer/docs/puppeteer.page.waitforselector/

plocket:
> How you you suggest handling it?

I think I'd try to detect the tab that has the actual ID. If it exists, can we assume its content exists too?

plocket:
I suggest not using `waitForSelector` as that will slow everything down. Unless there's a reason you can think that the element would take a while to appear on the page. [Use `.$()` instead.]

[That is, don't check afterwards, check beforehand more quickly.]

BryceStevenWilley:
Everything should be good (addressed the last comment, and added info to the changelog). The documentation is ready in https://github.com/SuffolkLITLab/docassemble-AssemblyLine-documentation/pull/244 and will be merged soon. I'll squash and merge now.

PR #561, opened 2022-05-16
Accessibility

BryceStevenWilley:
Adds accessibility checking ability to Kiln, using the popular `aXe-core` library behind everything from [Deque](https://www.deque.com/axe/) to [Chrome's Lighthouse](https://developers.google.com/web/tools/lighthouse/).

Adding it to an existing scenario is as simple as adding the following:
```
    And I check all pages for accessibility issues
```
and Kiln will then check every new page that it navigates to for accessibility issues, specifically WCAG 2.0 and 2.1, A and AA. It will take several seconds for each page, some to wait for all transitions to finish, and some to check for the actual issues.

You can also check individual pages by inserting the following:
```
    And I check the page for accessibility issues
```
when on a specific page.

Everything else (except documentation) is in, and I want to have a full production CI running for EFSPIntegration to make sure things are stable and workable before documenting, so I can document the whole process.

### DIFF
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -481,34 +481,38 @@ module.exports = {
     await scope.tapElement( scope, elem, tab_id );
   },
 
-  checkForAccessibility: async function (scope ) {
+  checkForA11y: async function (scope ) {
     /** Return if there were any violations and the dir with more details of the checks */
     let { id } = await scope.examinePageID( scope, 'none to match' );
     if ( scope.page_id !== id ) {
       await scope.addToReport(scope, { type: `page_id`, value: `${ id }`, });
       scope.page_id = id;
     }
-    let dirname = await scope.getSafeScenarioFilename( scope, {
-      prefix: `aXe_${ scope.page_id }`
-    });
     // HACK(brycew): wait for transitions to stop before checking for accessibility
     waitForTimeout(2000);
-    await scope.page.screenshot({path: `testing_a11y${scope.page_id}.jpg`, type: `jpeg`, fullPage: true})
     const axe = new AxePuppeteer(scope.page);
     axe.options({
       runOnly: ["wcag2a", "wcag2aa", "wcag21a", "wcag21aa"],
       resultTypes: ['violations', 'incomplete']
     });
     let axe_result = await axe.analyze();
-    axe_result.page_id = `ALKiln Accessible: ${ scope.page_id }`;
+    axe_result.page_id = scope.page_id;
     axe_result.passes_ids = axe_result.passes.map(pass_obj => pass_obj.id);
     axe_result.inapplicable_ids = axe_result.inapplicable.map(inapp_obj => inapp_obj.id);
     axe_result.passes = [];
     axe_result.inapplicable = [];
-    addToDebugLog(`Accessibility:\nReport for ${ scope.page_id }: ${ JSON.stringify(axe_result, null, 2)}\n`); 
+
+    let json_output = null;
+    const passed = axe_result.violations.length == 0;
+    if (!passed) {
+      json_output = await scope.getSafeScenarioFilename( scope, {prefix: `aXe_failure_${ scope.page_id }`}) + '.json';
+      fs.writeFileSync( json_output, JSON.stringify( axe_result, null, 2 ));
+    } else {
+      await scope.addToReport(scope, { type: `info`, value: `Accessibility on ${ scope.page_id } passed!`})
+    }
     return {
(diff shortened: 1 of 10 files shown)


BryceStevenWilley:
Tests are failing because https://github.com/plocket/docassemble-ALAutomatedTestingTests/pull/189 isn't merged yet, but everything else is ready for review!

lib/scope.js
```diff
@@ -46,6 +48,10 @@ let is_secret_var = function(row) {
   return original.flags !== undefined && original.flags.secret_name !== undefined;
 }
 
+let addToDebugLog = function(value) {
```
BryceStevenWilley: Move this to logs.js

lib/scope.js
```diff
@@ -475,6 +481,41 @@ module.exports = {
     await scope.tapElement( scope, elem, tab_id );
   },
 
+  checkForA11y: async function (scope ) {
+    /** Return if there were any violations and the dir with more details of the checks */
```
BryceStevenWilley: This is outdated: mention what we do now

lib/scope.js
```diff
@@ -475,6 +481,41 @@ module.exports = {
...
+    let { id } = await scope.examinePageID( scope, 'none to match' );
+    if ( scope.page_id !== id ) {
+      await scope.addToReport(scope, { type: `page_id`, value: `${ id }`, });
+      scope.page_id = id;
+    }
+    // HACK(brycew): wait for transitions to stop before checking for accessibility
+    waitForTimeout(2000);
```
BryceStevenWilley: See if I can remove this. Otherwise, reduce to 500ms / directly use showif_timeout.

lib/scope.js
```diff
@@ -1615,7 +1656,7 @@ module.exports = {
         }
       }
       if ( num_unused_rows > 0 ) {
-        let msg = `Missing variable or variables on page.`;
+        let msg = `Missing variable or variables on page (should be unused rows):`;
```
BryceStevenWilley: Take out, unrelated.

lib/scope.js
```diff
@@ -475,6 +481,41 @@ module.exports = {
...
+    axe_result.inapplicable = [];
+
+    let json_output = null;
+    const passed = axe_result.violations.length == 0;
+    if (!passed) {
+      json_output = await scope.getSafeScenarioFilename( scope, {prefix: `aXe_failure_${ scope.page_id }`}) + '.json';
+      fs.writeFileSync( json_output, JSON.stringify( axe_result, null, 2 ));
+    } else {
+      await scope.addToReport(scope, { type: `info`, value: `Accessibility on ${ scope.page_id } passed!`})
+    }
```
BryceStevenWilley: See if adding to the report on failure can be moved here.

lib/scope.js
```diff
@@ -1641,9 +1682,21 @@ module.exports = {
...
+      did_navigate = true;
       process.stdout.write(`\x1b[36m${ '>' }\x1b[0m`);  // continued successfully
     }  // ends if !did_navigate && ensure_navigation
 
+    if ( did_navigate && scope.check_all_for_a11y ) {
+      let { passed, json_output} = await scope.checkForA11y(scope);
+      if (!passed) {
+        await scope.addToReport(scope, {
+          type: `error`,
+          value: `Page ${ scope.page_id } is not accessible! See ${ json_output }`
```
BryceStevenWilley: `json_output` is really a file. Make the name better.

package.json
```diff
@@ -34,6 +35,7 @@
   },
   "homepage": "https://suffolklitlab.org/docassemble-AssemblyLine-documentation/docs/automated_integrated_testing/",
   "dependencies": {
+    "@axe-core/puppeteer": "^4.4.2",
```
BryceStevenWilley: Remove the carrot.

lib/scope.js
```diff
@@ -1687,12 +1689,8 @@ module.exports = {
     }  // ends if !did_navigate && ensure_navigation
 
     if ( did_navigate && scope.check_all_for_a11y ) {
-      let { passed, json_output} = await scope.checkForA11y(scope);
+      let { passed, _ } = await scope.checkForA11y(scope);
```
plocket: Should this be `axe_file`? And does `axe_file` mean the axe filename?

BryceStevenWilley: That is intentionally nothing, as in the parameter is unused. But I'm realizing that you can just do `let { passed } = ...` and completely not mention the other dict attribute. I really like that, so I'll change it to that instead.

PR #564, opened 2022-05-27
Handle everything when object drop downs are added

BryceStevenWilley:
Changes that are necessary to get tests working again (after https://github.com/plocket/docassemble-ALAutomatedTestingTests/pull/188 is merged; will fail before those are merged).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,7 +38,11 @@ Format:
 ### Security
 - 
 -->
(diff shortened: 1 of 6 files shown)


plocket (approved):
Looking good I think! Great addition and great fix found. Once the changelog is updated, this should be good to go. Maybe we should merge this before the previous PR to properly make sure all tests are passing.

PR #565, opened 2022-06-09
Fix broken Action.yml and linter on test_tests.yml

BryceStevenWilley:
Ran into an error when trying to setup IL on the testing framework. For action.yml the issue was:

"The template is not valid. SuffolkLITLab/ALKiln/releases/v4/action.yml (Line: 65, Col: 7): 'run' is already defined"

From https://github.com/IllinoisLegalAidOnline/docassemble-Appearance/runs/6813721097?check_suite_focus=true

Also installed a GitHub action schema linter and fixed an error in `test_tests.yml`:
"Missing property required" on the `extra_languages` and `tags` inputs

### DIFF
--- a/.github/workflows/test_tests.yml
+++ b/.github/workflows/test_tests.yml
@@ -5,9 +5,11 @@ on:
   workflow_dispatch:
     inputs:
       extra_languages:
+        required: False
         description: 'Optional. A list of comma separated language names visible on buttons or links that change the language of the interview. Overrides the EXTRA_LANGUAGES repo secret.'
         default: ''
       tags:
+        required: False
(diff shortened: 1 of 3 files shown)


BryceStevenWilley:
@plocket given that this is a simple change, needed to onboard a new user to the testing framework, and also suggested by the VSCode GitHub linter, if this passes tests, I'm going to directly push to main and close this PR.

plocket (approved):
Very strange we haven't run into these before and thanks for the fix! I gave it a quick once over, but approving quickly for quick fix.

plocket:
~We'll need to go over some of these errors soon, though.~ Result of discussion: These errors are because of the update to the testing interview made for another PR. They'll resolve when we merge that other PR.

Issue #563, opened 2022-05-27
Remove tags from generated artifacts


BryceStevenWilley:
I.e. the changes in https://github.com/SuffolkLITLab/ALKiln/pull/561/files#diff-4b632aedc4c91285de358dd7cbcc91a52724745c5c759b81b3a6d80e8d6d7abcL187-R195.

Adding all of the tags to every output file makes the file names much too long, to the point where it's difficult to see more important information about an artifact. IMO tags are much less useful that the scenario id, which serves the same purpose.

plocket:
Sounds good. I was on the fence about having them around anyway.

plocket:
Closed by #568

PR #568, opened 2022-06-23
Close #552, random input prep to rearrange artifact dirs

### LINKED ISSUE #552: Random input: start to implement new directory structure
From https://github.com/SuffolkLITLab/ALKiln/discussions/531#discussioncomment-2442625

### This issue

First step: Move current artifacts to new structure:
- [x] report
- [x] debug
- [x] [redundant error screenshots (that stay out in the main folder)]
- [x] scenario folder
   - [x] [report for just the scenario]
   - [x] downloaded files
   - [x] error screenshots
   - [x] manually taken screenshots
   - [x] JSON

First, for a test run, everything would be in one artifact folder, so it can all be downloaded at the same time. Name options?
- "alkiln_tests_output-" and timestamp (dashes or underscores?)
- Somehow get the name of the test run itself. Probably possible. Plus timestamp? Plus alkiln?

This will be good in general even before starting with random input and is the first step.

Discussions on what to use in the names of artifacts: https://github.com/SuffolkLITLab/ALKiln/issues/429#issuecomment-1002657179

### Long term plan and comments

Then the below structure for the folders eventually:

> Maybe the error screenshots should be in both places. This is the idea proposed here, though the names are missing a bunch of pieces to avoid a mess and show the arrangement more clearly.
> 
> - report.txt (whole report)
> - error-3pm-scenario 1 description-endingPgID.png
> - error-4pm-scenario 2 description-endingPgID.png
> - ~3pm-endingPgID-~ scenario 1 description (folder)
>    - report-endingPgID.txt (scenario only)
>    - download-3pm-file-1-pgID.pdf
>    - error-3pm-pgID.png (Same pic as "error-3pm-scenario 1 description.png". Same name seems dumb because it would have the scenario name in it too)
>    - json-3pm-pgID.json
>    - json-3pm-pgID.json
>    - scenario 1 description.feature
>    - screenshot-3pm-pgID.png
> - ~4pm-endingPgID-~ scenario 2 description (folder)
>    - report.txt (scenario only)
>    - error-4pm-pgID.png (Same as scenario 1 error pic name rationale)
>    - scenario 2 description.feature
> 
> "3pm" and "4pm" just indicates a timestamp. "PgID" could be question id or screen trigger var, whichever is available, or neither if none available. Not sure which is preferred.

plocket:
Feedback on final filenames (you can see in the issue) appreciated. There's also some weirdness to figure out. For example:

* Saving to the debug log may have to be abstracted differently because the directory name is now dynamic
* Saving unit tests for our own lib is different from saving other artifacts in a weird way and I'm not sure I really hit on the right way to do it
* Saving aXe errors isn't tested (from what I can tell), so I can't see the result of that
* I made the artifacts folder during setup and, after a conversation with @BryceStevenWilley , I'm not so sure that's the right choice. That is, are we going to save any artifacts from setup or takedown or is the info in the GitHub console enough? Should that be moved to `BeforeAll()`.

Probably other questions will arise.

Closes #552 

[Also haven't had a chance to look at the github artifacts of ours or of a dependent yet.]

### DIFF
--- a/.github/workflows/test_tests.yml
+++ b/.github/workflows/test_tests.yml
@@ -39,6 +39,9 @@ jobs:
 
     name: Run Puppeteer tests
     steps:
+      - name: Set name of artifact folder with date and UTC time
+        # https://www.shell-tips.com/linux/how-to-format-date-and-time-in-linux-macos-and-bash/#how-to-format-a-date-in-bash
+        run: echo "ARTIFACT_NAME=alkiln_tests_output $(date +'%Y-%m-%d at %Hhrs %Mmins %Ssecs' -u) UTC" >> $GITHUB_ENV
       - name: Set languages
         run: echo "EXTRA_LANGUAGES=${{ github.event.inputs.extra_languages }}" >> $GITHUB_ENV
         if: ${{ github.event.inputs.extra_languages != '' }}
@@ -53,46 +56,21 @@ jobs:
       - run: npm install
       - run: npm run setup
       - run: npm run test -- ${{ github.event.inputs.tags && format('"--tags" "{0}"', github.event.inputs.tags) }}
-      - name: upload error artifacts
+      # Artifacts
+      - name: upload unit tests artifacts folder
         uses: actions/upload-artifact@v2
-        if: ${{ failure() }}
-        with:
-          name: errors
-          path: ./**/error*.jpg
-      - name: upload screenshot steps
-        uses: actions/upload-artifact@v2
-        # They will upload even if a previosu step has failed
+        # This will upload even if a previous step has failed
         if: ${{ always() }}
         with:
-          name: screenshots
-          path: ./**/screenshot*.jpg
-      - name: upload downloads
-        uses: actions/upload-artifact@v2
-        # They will upload even if a previous step has failed
-        if: ${{ always() }}
-        with:
-          name: downloads
-          path: ./**/downloads_*
-      - name: upload report
-        uses: actions/upload-artifact@v2
-        # They will upload even if a previous step has failed
-        if: ${{ always() }}
-        with:
-          name: report
-          path: ./**/*_report_*
-      - name: upload a11y reports
-        uses: actions/upload-artifact@v2
-        # They will upload even if a previous step has failed
-        if: ${{ always() }}
-        with:
-          name: a11y report
-          path: ./**/aXe_failure_*
-      - name: upload debug_log artifact
+          name: ${{ env.ARTIFACT_NAME }}
+          path: ./alkiln_tests_misc-*
+      - name: upload integration tests artifacts folder
         uses: actions/upload-artifact@v2
         if: ${{ always() }}
         with:
-          name: debug_log file
-          path: ./debug_log.txt
+          name: ${{ env.ARTIFACT_NAME }}
+          path: ./alkiln_tests_output-*
+      # Delete Playground Project and project name file
       - name: Cleanup
         if: ${{ always() }}
         run: npm run takedown
--- a/.github/workflows/test_tests.yml
+++ b/.github/workflows/test_tests.yml
@@ -41,8 +41,9 @@ jobs:
     steps:
       - name: Set name of artifact folder with date and UTC time
         # https://www.shell-tips.com/linux/how-to-format-date-and-time-in-linux-macos-and-bash/#how-to-format-a-date-in-bash
-        run: echo "ARTIFACT_NAME=alkiln_tests_output-$(date +'%Y-%m-%d at %Hhrs %Mmins %Ssecs' -u) UTC" >> $GITHUB_ENV
-        run: echo "UNIT_TESTS_ARTIFACT_NAME=_alkiln_tests_misc-$(date +'%Y-%m-%d at %Hhrs %Mmins %Ssecs' -u) UTC" >> $GITHUB_ENV
+        run: |
+          echo "ARTIFACT_NAME=alkiln_tests_output-$(date +'%Y-%m-%d at %Hhrs %Mmins %Ssecs' -u) UTC" >> $GITHUB_ENV
+          echo "UNIT_TESTS_ARTIFACT_NAME=_alkiln_tests_misc-$(date +'%Y-%m-%d at %Hhrs %Mmins %Ssecs' -u) UTC" >> $GITHUB_ENV
       - name: Set languages
         run: echo "EXTRA_LANGUAGES=${{ github.event.inputs.extra_languages }}" >> $GITHUB_ENV
         if: ${{ github.event.inputs.extra_languages != '' }}
--- a/.github/workflows/test_tests.yml
+++ b/.github/workflows/test_tests.yml
@@ -41,7 +41,8 @@ jobs:
     steps:
       - name: Set name of artifact folder with date and UTC time
         # https://www.shell-tips.com/linux/how-to-format-date-and-time-in-linux-macos-and-bash/#how-to-format-a-date-in-bash
-        run: echo "ARTIFACT_NAME=alkiln_tests_output $(date +'%Y-%m-%d at %Hhrs %Mmins %Ssecs' -u) UTC" >> $GITHUB_ENV
+        run: echo "ARTIFACT_NAME=alkiln_tests_output-$(date +'%Y-%m-%d at %Hhrs %Mmins %Ssecs' -u) UTC" >> $GITHUB_ENV
+        run: echo "UNIT_TESTS_ARTIFACT_NAME=_alkiln_tests_misc-$(date +'%Y-%m-%d at %Hhrs %Mmins %Ssecs' -u) UTC" >> $GITHUB_ENV
       - name: Set languages
         run: echo "EXTRA_LANGUAGES=${{ github.event.inputs.extra_languages }}" >> $GITHUB_ENV
         if: ${{ github.event.inputs.extra_languages != '' }}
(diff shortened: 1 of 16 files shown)


.github/workflows/test_tests.yml
```diff
@@ -39,6 +39,10 @@ jobs:
 
     name: Run Puppeteer tests
     steps:
+      - name: Set name of artifact folder with date and UTC time
+        # https://www.shell-tips.com/linux/how-to-format-date-and-time-in-linux-macos-and-bash/#how-to-format-a-date-in-bash
+        run: echo "ARTIFACT_NAME=alkiln_tests_output-$(date +'%Y-%m-%d at %Hhrs %Mmins %Ssecs' -u) UTC" >> $GITHUB_ENV
+        run: echo "UNIT_TESTS_ARTIFACT_NAME=_alkiln_tests_misc-$(date +'%Y-%m-%d at %Hhrs %Mmins %Ssecs' -u) UTC" >> $GITHUB_ENV
```
BryceStevenWilley: You can't have two keys of the same name in workflow YAML, github has complained to be about that before in #565.

Not sure how it's working fine now, but this should be the same thing (and a bit easier to read):
```suggestion
        run: |
          echo "ARTIFACT_NAME=alkiln_tests_output-$(date +'%Y-%m-%d at %Hhrs %Mmins %Ssecs' -u) UTC" >> $GITHUB_ENV
          echo "UNIT_TESTS_ARTIFACT_NAME=_alkiln_tests_misc-$(date +'%Y-%m-%d at %Hhrs %Mmins %Ssecs' -u) UTC" >> $GITHUB_ENV
```

lib/utils/session_vars.js
```diff
@@ -107,14 +113,14 @@ session_vars.save_project_name = function ( project_name ) {
...
+  /* Get the name of the current docassemble Project  */
   let project_name =  JSON.parse( fs.readFileSync( project_name_file_path ));
-  if ( session_vars.get_debug() ) { console.log( `Project name from file is "${ project_name }".` ); }
+  if ( session_vars.get_debug() ) { console.log( `ALKiln debug: Project name from file is "${ project_name }".` ); }
   return project_name;
 };  // Ends get_project_name()
 
 session_vars.delete_project_name = function () {
-  /* Delete the file storing the name of the docassemble project */
+  /* Delete the file storing the name of the docassemble Project */
```
BryceStevenWilley:
```suggestion
  /* Delete the file storing the name of the docassemble project */
```

lib/utils/session_vars.js
```diff
@@ -107,14 +113,14 @@ session_vars.save_project_name = function ( project_name ) {
 };  // Ends save_project_name()
 
 session_vars.get_project_name = function () {
-  /* Get the name of the current project  */
+  /* Get the name of the current docassemble Project  */
```
BryceStevenWilley, line 120:
```suggestion
  /* Get the name of the current docassemble project  */
```

lib/utils/session_vars.js
```diff
@@ -127,6 +133,29 @@ session_vars.delete_project_name = function () {
   }
 };
 
+// The file storing the artifacts path contains the name of the current artifacts folder.
+// The directory name it contains will be created or changed with each run of all the tests.
+const file_storing_artifact_path = `_alkiln_test_artifacts_dir_name.json`;
+session_vars.save_artifacts_path_name = function ( path_name ) {
```
BryceStevenWilley: If we're changing how `_al_test_project_name.json` works in this PR, I'd prefer changing it to something that only has one file, like a `runtime_config.json` or `setup_info.json`.

These functions could still stay the same, they'd just all read and write a JSON map with keys like `test_artifacts_dir_name` and `test_project_name` to the same file.

Mostly because I can imagine this sort of thing (passing information from setup to the actual runtime) might continue to happen, and IMO we don't want a ton of extra files hanging around that all serve kinda the same purpose.

lib/utils/session_vars.js
```diff
@@ -21,9 +21,15 @@ const session_vars = {};
 module.exports = session_vars;
 
 // These default to null
+// TODO: How to isolate debug elsewhere so this file can use the logger?
 session_vars.get_debug = function () { return process.env.DEBUG || null; };
 session_vars.get_dev_api_key = function () { return process.env.DOCASSEMBLE_DEVELOPER_API_KEY || null; };
 session_vars.get_repo_url = function () { return process.env.REPO_URL || null; };
+// .is_local more appropriate? To manage storing local results differently considering
```
BryceStevenWilley: IMO .is_dev_env is fine. It describes the why (because we're re-running scenarios without setup each time), and not the how (locally, which could technically be through an SSH'd server and not locally).

lib/steps.js
```diff
@@ -178,7 +202,11 @@ Given(/I start the interview at "([^"]+)"(?: in lang "([^"]+)")?/, {timeout: -1}
   await custom_download_behavior_timeout;
 
   // ---- EVENT LISTENERS ----
-  // Wait for possible download. From https://stackoverflow.com/a/51423688
+
+  // TODO: Wait for download to complete doesn't work. See notes in
+  // scope.js scope.detectDownloadComplete()
```
BryceStevenWilley, line 207: I actually just did some work on this! 😉  see https://github.com/SuffolkLITLab/ALKiln/tree/faster_downloads

lib/steps.js
```diff
@@ -935,8 +963,22 @@ After(async function(scenario) {
...
         // Save/download a picture of the screen that's showing during the unexpected status
-        let safe_filename = await scope.getSafeScenarioFilename( scope, { prefix: `error` });
-        await scope.page.screenshot({ path: `${ safe_filename }.jpg`, type: `jpeg`, fullPage: true });
+        // Save one copy in the outer-most artifact folder
+        let scenario_filename = await scope.getSafeScenarioFilename( scope, { prefix: `error_on` });
+        await scope.page.screenshot({
+          path: `${ scope.paths.artifacts }/${ scenario_filename }.jpg`,
+          type: `jpeg`,
+          fullPage: true
+        });
```
BryceStevenWilley, line 972: Is it worth it to just copy the screen shot (we know the file name and that it's completed) instead taking a duplicate? Should be a small bit faster, minor nit though.

lib/scope.js
```diff
@@ -192,36 +212,47 @@ module.exports = {
     }
 
     let filename = name_parts.join(`-`);
+    let safe_name = safe_filename( filename, { replacement: `_` });
```
BryceStevenWilley, line 212: If we're changing file names, I'd raise #563 again, which to fix is just deleting lines 210-212.  Especially if we're going to shorten the name strings anyway.

BryceStevenWilley (approved):
LGTM, mostly minor comments, but we should at least address the comment for `test_tests.yml`, I'm still paranoid after #565.

lib/utils/session_vars.js
plocket: This is capitalized on purpose to represent that it's a Thing in the docassemble world. Matches the capital case in the UI of docassemble.

lib/utils/session_vars.js
plocket: Same as above.

lib/utils/session_vars.js
plocket: Hmm, good point, but the project name file gets deleted when the project gets deleted (on takedown) in order to make sure it doesn't interfere with new project names. Thoughts?

lib/utils/session_vars.js
plocket: Was going to do a `suggestion`, but it needs to cover more lines.

lib/steps.js
plocket: Awesome! Hopefully we can get rid of this when we integrate those!

lib/steps.js
plocket: It's a tiny bit faster, but I like having them in both places. The screenshot should definitely be in the Scenario folder that errored, but having all of them in the main folder lets you scroll through them really easily, getting an idea of a pattern if there is one instead of hopping back and forth between folders. I'm open to more thoughts though. Do you want to make a discussion item?

lib/scope.js
plocket: Fair enough. I wasn't sure about keeping them in anyway. I'll try to work on that soon.

lib/steps.js
BryceStevenWilley: Sorry, I agree with having two screenshots, I was just saying that we don't need to take the screen shot twice: we can just copy the image file from the first one into the main folder. If we're fine with taking two different screenshots, that's fine, and it doesn't need to be a discussion item.

lib/utils/session_vars.js
BryceStevenWilley: Not sure I see how that's a problem, you can still remove the project name key from the JSON map.

lib/utils/session_vars.js
BryceStevenWilley: Now that you've pointed it out, the docassemble UI is gonna bug me, lol. Not a biggie though

lib/scope.js
plocket: I will say that the tags make it easier for me to find the file the scenario is in since, of course, I don't have the scenario descriptions memorized, but I do know that a tag of @story_tables means it's in the story tables test file. Maybe that's just because I know this library well right now, though. Maybe it's not that helpful to others?

BryceStevenWilley: I don't have the scenario description memorized either, but I also don't remember enough of the existing tests (in any repo) to be able to recognize any of the tags too. And in that case, half of the tags are used on many other tests as well and get in the way of easily recognizing the one you need. I have to search for the description verbatim in the project, which is the same in the file name, unlike the tags, which have dashes added between them.

PR #569, opened 2022-06-23
Fix finding single select fields on a page

BryceStevenWilley:
When a single field page with the dropdown field is present, no fields on the page are detected. Originally found when I was testing the  by the [`method_of_service` var from the SP6A.yml interview](https://github.com/SuffolkLITLab/docassemble-MotionToStayEviction/blob/db2598102cc24b0bd4ec1048b1e5102c4b1d07b7/docassemble/MotionToStayEviction/data/questions/SP6A.yml#L357) in the [MotionToStayEviction](https://github.com/SuffolkLITLab/docassemble-MotionToStayEviction/) interview, and needed to make that work.

Could probably be a more specific selector but I figured broader is better than too narrow (meaning we'd have to patch another bug like this in the future). It's not too broad though, given it's the `form` element.

Partners with https://github.com/plocket/docassemble-ALAutomatedTestingTests/pull/192, will need to be merged after that is merged for tests to pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,7 +38,9 @@ Format:
 ### Security
 - 
 -->
-<!-- ## [Unreleased] -->
+## [Unreleased]
+### Fixed
+- Single dropdown fields on pages are now recognized as fields properly
 
 ## [4.6.0] - 2022-06-10
 ### Added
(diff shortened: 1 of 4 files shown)


plocket (commented):
Great change! I'll review the other repo for the tests to pass. Could you add to the changelog?

plocket (approved):
Looks good, thanks!

plocket:
Ready to merge this one @BryceStevenWilley ?

PR #570, opened 2022-06-30
Close #392, avoid server reload errors more and handle them more gracefully

### LINKED ISSUE #392: Integrate API key to remove/reduce errors caused by server reload

plocket:
This is my first official attempt, and a pretty complex change.

[The problem this is trying to solve: tests sometimes fail because the server is reloading for unrelated reasons. This is a false failure. It requires the dev to figure out what happened then re-run the tests. After a while, people may start ignoring failures, figuring they are server reload issues.]

What should happen (which I admit is hard to test):

1. Tests start to run.
2. Another project is pulled and triggers a server reload
3. Either the test is just trying to get to the interview or is trying to continue to a new page and times out trying to do that.
4. The test detects that the problem is a server reloading.
5. The test waits until the server has reloaded (with a maximum time limit) so that the next test won't immediately fail.
6. The test throws an error (will explain why lower down) that the server was reloading. It also adds that explanation to the report.
7. The command line script knows to retry failed tests once, so it will try that test again later, hopefully with a better result.

Why throw an error even after reload has successfully completed?
1. We don't know whether the page will need reloading or can just continue at that point.
2. If the page needs reloading, there's currently no way for us to replay the user answers for that page. Specifically, we can't re-do non-story table Steps over again as they're in the past and I think cucumber doesn't give us access to them. Even if cucumber was good about that, we don't know for sure which Step was the last navigating step, so we wouldn't know how far to go back.

**Testing:**
I test at least one part of this by starting to run the "@st2" test (a long running one), waiting 2 seconds after the first dot appears, and, in a different terminal, start a new `npm run setup`. If I time it right, I get to see the results I'm looking for. I do have to keep taking down the new docassemble Project (long explanation you can skip: the time it takes to iterate through and get a unique project name takes too long and the test finishes before the timeout can happen). That means running takedown, then editing the "runtime_config.json" file so that it contains the previous Project name. For me, that's `testingdocassembleALAutomatedTestingTestsmain1`. I then save the file.

[I haven't yet devised a way to test the situation where the test is trying to load the interview's first page. Maybe make two local repos and set one up and then time the test and new setup so they overlap? I'm not sure.

Testing on GitHub for this probably won't work as the timing would be impossible - we'd have to start running setup on both tests and hope that the server timeout didn't interfere with the other setup, just with the initial page load once the tests had started. ]

[This doesn't yet take care of download reloading problems[The worst consequence is that the tests won't wait properly for the reload, which could mess up the next test.] ~but that honestly just~ It also means that those won't get the right error message in the report or console, but that's not as big a deal. Not great, but not the worst, especially since no one is using that feature right now - there's no way to compare them to established PDFs. That should be made into an issue, though.

We did say this doesn't have to take care of everything.]

Closes #392

### DIFF
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -439,24 +439,21 @@ module.exports = {
   track_server_status: async function( scope ) {
     /** Updates server_statuses list by checking if server is responding until,
     *    eventually, it is told to stop. */
-    // This code assumes a reasonable response time would be 2 seconds if the
-    // server is up and that a server reload/restart won't take less than 2 seconds.
-    // Also, it's non-blocking, so it doesn't slow down tests
-    let is_up = await server.is_server_responding({ timeout: scope.server_check_interval_ms });
-    // Add status to the beginning of the list to make truncating easier
-    scope.server_statuses.unshift( is_up );
 
-    // To make it easier to check past server statuses, remove defunct server statuses
-    let max_length = Math.round( scope.timeout/scope.server_check_interval_ms );
-    // Trying to extract more items than the actual length is ok
-    scope.server_statuses = scope.server_statuses.slice( 0, max_length );
+    while ( !scope.stop_tracking_server_status ) {
+      // This code assumes a reasonable response time would be 2 seconds if the
+      // server is up and that a server reload/restart won't take less than 2 seconds.
+      // Also, it's non-blocking, so it doesn't slow down tests
+      let is_up = await server.is_server_responding({ timeout: scope.server_check_interval_ms });
+      // Add status to the beginning of the list to make truncating easier
+      scope.server_statuses.unshift( is_up );
 
-    // Unless told to stop, store the next status
-    if ( !scope.stop_tracking_server_status ) {
-      // No await. That's already happened with the server response await. I think.
-      // Will this cause the process to not exit? If so, only for a short time.
-      scope.track_server_status( scope );
-    }
+      // To make it easier to check past server statuses, remove defunct server statuses
+      let max_length = Math.round( scope.timeout/scope.server_check_interval_ms );
+      // Trying to extract more items than the actual length is ok
+      scope.server_statuses = scope.server_statuses.slice( 0, max_length );
+    }  // ends while !scope.stop_tracking_server_status
+    
   },  // Ends scope.track_server_status()
 
   did_server_reload: async function( scope ) {
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -386,28 +386,100 @@ module.exports = {
     }
 
     let winner = null;
-    try {
-      // Error loads page, so no need to detect to keep things short
-      winner = await Promise.race([
-        clickWait,
-        // This is meant to detect user-visible alerts/errors,
-        // not breaking errors in the testing code.
-        scope.page.waitForSelector('.alert-danger', { visible: true }),
-        scope.page.waitForSelector('.da-has-error', { visible: true }),
-      ]);
-    } catch ( error ) {
 
-      let err_msg = `Error occurred when tried to click ${ tapperText }.`
-      if ( error.name === `TimeoutError` ) {
-        let non_reload_report_msg = `Timeout error occurred when tried to click "${ tapperText }".`;
+    let click_timeout = new Promise(function( resolve, reject ) {
+
+      // Set up the timeout that avoids infinite loops
+      let on_click_timeout = async function() {
+
+        let { id } = await scope.examinePageID( scope, `doesn't matter` );
+
+        let non_reload_report_msg = `ALKiln tried to tap "${ tapperText }", but it took `
+        + `too long. The question id was "${ id }".`
+        let error = non_reload_report_msg;
         scope.handle_page_timeout_error( scope, { non_reload_report_msg, error });
-      } else {
-        // Throw any non-timeout error
-        scope.addToReport( scope, { type: `error`, value: err_msg });
-        throw( error );
-      }  // ends if error is timeout error
 
-    }  // ends try/catch
+      };  // Ends on_click_timeout()
+      let click_timeout_id = setTimeout(on_click_timeout, scope.timeout);
+
+      // self-invoke an anonymous async function so we don't have to
+      // deal with a complex abstraction of the details of resolving and rejecting.
+
+      (async function() {
+        try {
+
+          // Error loads page, so no need to detect to keep things short
+          winner = await Promise.race([
+            clickWait,
+            // This is meant to detect user-visible alerts/errors,
+            // not breaking errors in the testing code.
+            scope.page.waitForSelector('.alert-danger', { visible: true }),
+            scope.page.waitForSelector('.da-has-error', { visible: true }),
+          ]);
+          resolve( winner );
+
+          // let { question_has_id, id, id_matches } = await scope.examinePageID( scope );
+
+          // // Add special rows. Basically, do weirder things.
+          // // TODO: Remove from_story_table - no longer needed if only called from in here
+          // let ensured_var_data = await scope.ensureSpecialRows( scope, {
+          //   var_data: supported_table,
+          //   from_story_table: true,
+          // });
+          // // Some errors happen in there to be closer to the relevant code. We'd have to try/catch anyway.
+          // let error = await scope.setFields(scope, { var_data: ensured_var_data, id, ensure_navigation: true });
+          // if ( error.was_found ) { await scope.throwPageError( scope, { id: id }); }
+          // resolve();
+        } catch ( error ) {
+          
+          let err_msg = `Error occurred when ALKiln tried to tap "${ tapperText }".`
+          if ( error.name === `TimeoutError` ) {
+            let non_reload_report_msg = `Timeout error occurred when ALKiln tried to tap "${ tapperText }".`;
+            // debugging reload
+            // awaiting this doesn't seem to help non-story-table server reload set var Step error
+            // await scope.handle_page_timeout_error( scope, { non_reload_report_msg, error });
+            scope.handle_page_timeout_error( scope, { non_reload_report_msg, error });
+          } else {
+            // Throw any non-timeout error
+            scope.addToReport( scope, { type: `error`, value: err_msg });
+            // throw( error );
+            reject( error );
+          }  // ends if error is timeout error
+
+        } finally {
+          clearTimeout( click_timeout_id );  // prevent temporary hang at end of tests
+        }
+      })();
+    });  // ends click_timeout
+
+    // Do we need winner to be defined here?
+    winner = await click_timeout;
+
+    // try {
+    //   // Error loads page, so no need to detect to keep things short
+    //   winner = await Promise.race([
+    //     clickWait,
+    //     // This is meant to detect user-visible alerts/errors,
+    //     // not breaking errors in the testing code.
+    //     scope.page.waitForSelector('.alert-danger', { visible: true }),
+    //     scope.page.waitForSelector('.da-has-error', { visible: true }),
+    //   ]);
+    // } catch ( error ) {
+
+    //   console.log( error.name );
+    //   let err_msg = `Error occurred when ALKiln tried to tap "${ tapperText }".`
+    //   if ( error.name === `TimeoutError` ) {
+    //     let non_reload_report_msg = `Timeout error occurred when ALKiln tried to tap "${ tapperText }".`;
+    //     // debugging reload
+    //     // awaiting this doesn't seem to help non-story-table server reload set var Step error
+    //     await scope.handle_page_timeout_error( scope, { non_reload_report_msg, error });
+    //   } else {
+    //     // Throw any non-timeout error
+    //     scope.addToReport( scope, { type: `error`, value: err_msg });
+    //     throw( error );
+    //   }  // ends if error is timeout error
+
+    // }  // ends try/catch
 
     // TODO: Not sure how to detect navigation landing at same page.
     // Maybe url change + trigger var change? Maybe just trigger var?
@@ -475,8 +547,7 @@ module.exports = {
     if ( server_reloaded_during_interval ) {
       // Wait for the server to reload just to make sure the next test
       // doesn't also fail because of this same server reload.
-      let reload_promise = await scope.get_server_reload_promise( scope );
-      await reload_promise;
+      await scope.get_server_reload_promise( scope );
     } else {
       // Otherwise just throw the error
       scope.addToReport( scope, {
@@ -524,7 +595,8 @@ module.exports = {
       scope.ms_till_server_timeout = parseInt( session_vars.get_server_reload_timeout() );
     } else {
       // Check again every two seconds
-      scope.ms_till_server_timeout = 5 * scope.timeout;
+      // debugging reload. revert to 5 when done
+      scope.ms_till_server_timeout = 20 * scope.timeout;
     }
     
     // Race between overall timeout and server finishing reloading. We'll throw
@@ -570,7 +642,8 @@ module.exports = {
   throw_server_was_reloading_error: async function( scope ) {
     /* Throw the server reload error with appropriate side effects. */
     let reload_failure_msg = `The test was unable to continue because the `
-      + `server was reloading. ALKiln will try this Scenario a total of 2 times.`;
+      + `server was reloading. ALKiln will try this Scenario a total of 2 times. An `
+      + `error will still be printed for this attempt.`;
     scope.addToReport( scope, { type: `error`, value: reload_failure_msg });
     throw( reload_failure_msg );
   },  // Ends scope.throw_server_was_reloading_error()
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -398,31 +398,6 @@ module.exports = {
       ]);
     } catch ( error ) {
 
-      /* ===== Brainstorms =====
-      Store the promise in state and then await that promise in here
-      Who should retry? If this is a story table, retrying shouldn't happen in here
-      Otherwise, this in here should retry.
-      Except! If we reload the page when the dev is using individual steps to set
-      values, the page's values will be blank. There's no way to look back over the
-      previous steps on this page.
-      Would it be crazy to just restart the test?
-        Depends on the test, but probably not
-      Is it even possible to restart the test?
-        I don't know at all
-      Raising a re-tryable error and then detect that in cucumber to somehow
-      trigger restarting the same scenario?
-      How about just failing the test and telling them why?
-        That can be a useful temporary improvement, but I'd rather not teach devs to
-        ignore test failures. It's better than opaque failures, though.
-      TODO: Also have to do this when first loading an interview or going to a
-        url with a link check.
-      MVP possibly: Add the server reload error message to the report, but also
-        add --retry 1 to the node script _and_ wait for the server to reload.
-        Sadly, --retry 1 will rerun actually failing tests too.
-        https://github.com/cucumber/cucumber-js/blob/main/docs/retry.md
-      ==========================
-      */
-
       let err_msg = `Error occurred when tried to click ${ tapperText }.`
       if ( error.name === `TimeoutError` ) {
         // If the server was reloading during that time, for now just give a very
@@ -438,7 +413,7 @@ module.exports = {
           // Otherwise just throw the error
           scope.addToReport( scope, {
             type: `error`,
-            value: `Timeout error occurred when tried to click ${ tapperText }.` });
+            value: `Timeout error occurred when tried to click "${ tapperText }".` });
           throw error;
         }
       } else {
@@ -488,13 +463,13 @@ module.exports = {
 
     // To make it easier to check past server statuses, remove defunct server statuses
     let max_length = Math.round( scope.timeout/scope.server_check_interval_ms );
-    // Extracting more items than the length is ok
+    // Trying to extract more items than the actual length is ok
     scope.server_statuses = scope.server_statuses.slice( 0, max_length );
 
     // Unless told to stop, store the next status
     if ( !scope.stop_tracking_server_status ) {
       // No await. That's already happened with the server response await. I think.
-      // Maybe this is causing process to not exit?
+      // Will this cause the process to not exit? If so, only for a short time.
       scope.track_server_status( scope );
     }
   },  // Ends scope.track_server_status()
@@ -510,15 +485,12 @@ module.exports = {
     */
     // console.log(`promise: ${ typeof( scope.server_reload_promise ) }`);
     if ( scope.server_reload_promise !== null ) {
-      console.trace(`=======\nPromise exists\n=======\n`);
-    console.log(`Promise: ${ scope.server_reload_promise }`);
       // Return the existing promise of the func that will check on reloading the server
       return scope.server_reload_promise
+
     } else {
-      console.trace(`=======\nNew Promise\n=======\n`);
       // Create and return the promise of the func that will check on reloading the server
       scope.server_reload_promise = scope.wait_for_server_response( scope );
-    console.log(`Promise: ${ scope.server_reload_promise }`);
       return scope.server_reload_promise;
     }
   },  // Ends scope.get_server_reload_promise()
@@ -536,27 +508,28 @@ module.exports = {
     * This currently assumes we're using scope.timeout as the amount of
     * time in which to detect reload
     */
-    console.log( 1 );
+    
     // Server timeout will always be based on scope.timeout even if it changes
     if ( session_vars.get_server_reload_timeout() !== null ) {
+      // Should we require the expected value to be seconds instead of ms?
       scope.ms_till_server_timeout = parseInt( session_vars.get_server_reload_timeout() );
     } else {
+      // Check again every two seconds
       scope.ms_till_server_timeout = 2 * scope.timeout;
     }
     
+    // Race between overall timeout and server finishing reloading. We'll throw
+    // an error anyway, but it'll be faster if the server does finish reloading.
     let server_promise = new Promise(async function( resolve, reject ) {
 
-      console.log( 2 );
       // to stop possible infinite loops
-        console.log( 3 );
-        let server_timeout_id = setTimeout(async function() {
-          console.log( 4 );
-          let timeout_message = `The server took longer than `
-            + `${ Math.round( scope.ms_till_server_timeout/1000 ) } seconds to respond.`
-          await scope.addToReport( scope, { type: `error`, value: timeout_message });
-          // reject( timeout_message );
-          throw new Error( timeout_message );
-        }, scope.ms_till_server_timeout );
+      let server_timeout_id = setTimeout(async function() {
+        let timeout_message = `The server took longer than `
+          + `${ Math.round( scope.ms_till_server_timeout/1000 ) } seconds to respond.`
+        await scope.addToReport( scope, { type: `error`, value: timeout_message });
+        // reject( timeout_message );
+        throw new Error( timeout_message );
+      }, scope.ms_till_server_timeout );
 
       // self-invoke an anonymous async function so we don't have to
       // abstract the details of resolving and rejecting.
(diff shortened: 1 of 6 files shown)


lib/scope.js
```diff
@@ -411,6 +432,152 @@ module.exports = {
...
+    // To make it easier to check past server statuses, remove defunct server statuses
+    let max_length = Math.round( scope.timeout/scope.server_check_interval_ms );
+    // Trying to extract more items than the actual length is ok
+    scope.server_statuses = scope.server_statuses.slice( 0, max_length );
+
+    // Unless told to stop, store the next status
+    if ( !scope.stop_tracking_server_status ) {
+      // No await. That's already happened with the server response await. I think.
+      // Will this cause the process to not exit? If so, only for a short time.
+      scope.track_server_status( scope );
```
BryceStevenWilley: Is this recursive? Don't know enough about node and stack overflows, but semantically it doesn't seem like this should be a recursive function and should just be called in something like:
```javascript
while (!script.stop_tracking_server_status ) {
 ...
}
```

lib/scope.js
```diff
@@ -411,6 +432,152 @@ module.exports = {
...
+
+    } else {
+      // Create and return the promise of the func that will check on reloading the server
+      scope.server_reload_promise = scope.wait_for_server_response( scope );
+      return scope.server_reload_promise;
+    }
+  },  // Ends scope.get_server_reload_promise()
+
+  // 5 times the default scope.timeout for now. Will be set dynamically later.
+  ms_till_server_timeout: 5 * 30 * 1000,
```
BryceStevenWilley: For simplicity, this could probably be a local variable to just the `wait_for_server_response` function.

BryceStevenWilley (dismissed):
LGTM! Will say I haven't personally run into these error often since using Kiln more, so I don't have any good ideas about corner cases or how to test it. But the code does look good!

plocket:
Trying to be as clear as I can about a confusing result, but probably failing badly. Also, trying to not put this on your shoulders. Hoping this will help me think about it:

In trying to implement some of the comments, I just had a test where I did a bunch of steps, failed on reload. The test got rerun, but failed to load the interview at all. That's confusing to me because the server reload flow should have waited till the server was available again before running. Maybe there was another server reload that interfered? Or maybe there was a problem with the test since it was being run from the same repo? It's the procedure I described in the PR's comment.

Also, the report came out weird. The message about the reload happening should have been in the first test with multiple steps and the story table, but instead it was with the second test where the interview failed to load. The report came out weird for the first one too - printing a failure, then another step, then a failure again. Something maybe wasn't timing out correctly.

[The test was an interactive one, tagged "@i1". The test for a story table was still passing for tag "@st2".

The cucumber report was also failing for "@i1", and in a confusing way - the reload error was also in the wrong test in their report and the `After` function had an error as well. That one was about losing puppeteer context, which implies navigation or page closure at the wrong time.]

The report:
```
===============================
===============================
Failed scenarios:

---------------
Scenario: I set various values
Tags: @interactive @slow @i1
---------------
screen id: upload-files
      | upload_files_visible | some_png_1.png, some_png_2.png |  |
      | show_upload | True |  |
      | upload_files_hidden | some_png_2.png,some_png_1.png |  |
screen id: group-of-complex-fields
      | double_quote_dict['double_quote_key']['dq_two'] | true |  |
      | single_quote_dict['single_quote_key']['sq_two'] | true |  |
screen id: direct-standard-fields
      | checkboxes_yesno | True |  |
      | checkboxes_other['checkbox_other_opt_1'] | true |  |
      | dropdown_test | dropdown_opt_2 |  |
      | radio_yesno | False |  |
      | radio_other | radio_other_opt_3 |  |
      | text_input | Regular text input field value |  |
      | textarea | Multiline text\narea value |  |
      | date_input | today - 1 |  |
      | direct_standard_fields | True |  |
screen id: showifs
      | show_2 | True |  |
      | show_3 | True |  |
      | showif_checkbox_yesno | true |  |
      | showif_checkboxes_other['showif_checkboxes_nota_2'] | true |  |
      | showif_checkboxes_other['showif_checkboxes_nota_3'] | true |  |
      | showif_yesnoradio | True |  |
      | showif_radio_other | showif_radio_multi_2 |  |
      | showif_text_input | Show if text input value |  |
      | showif_textarea | Show if\nmultiline text\narea value |  |
      | showif_dropdown | showif_dropdown_1 |  |
screen id: object-checkboxes
      | object_checkboxes_test['obj_chkbx_opt_1'] | true |  |
      | object_dropdown | obj_opt_2 |  |
screen id: buttons-yesnomaybe
      | buttons_yesnomaybe | None |  |
screen id: buttons-other
      | buttons_other | button_3 |  |
screen id: button-continue
      | button_continue | True |  |
screen id: screen-features
screen id: proxy-vars
      | x[i].name.first | Proxyname1 |  |
screen id: is-there-another-generic
      | x.there_is_another | True |  |
screen id: proxy-vars
      | x[i].name.first | Proxyname2 |  |
screen id: is-there-another-generic
**-- Scenario Failed --**
      | x.there_is_another | False |  |
**-- Scenario Failed --**

---------------
Scenario: I set various values
Tags: @interactive @slow @i1
---------------

ERROR: The test was unable to continue because the server was reloading. ALKiln will try this Scenario a total of 2 times.
**-- Scenario Failed --**
**-- Scenario Failed --**

ERROR: Error occurred when tried to go to the interview url "https://apps-dev.suffolklitlab.org/interview?new_session=1&i=docassemble.playground204testingdocassembleALAutomatedTestingTestsrandom2%3Aall_tests.yml".
```

Here's the cucumber report, which also seems to flip which error happened on which test:
```
Failures:

1) Scenario: I set various values (attempt 2) # tests/features/interactive_steps.feature:7
   ✔ Before # lib/steps.js:96
   ✖ Given I start the interview at "all_tests" # lib/steps.js:137
       UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "The test was unable to continue because the server was reloading. ALKiln will try this Scenario a total of 2 times.".
   - Then the question id should be "upload files" # lib/steps.js:473
   - When I set the var "upload_files_visible" to "some_png_1.png, some_png_2.png" # lib/steps.js:798
   - And I set the var "show_upload" to "True" # lib/steps.js:798
   - And I upload "some_png_2.png,some_png_1.png" to "upload_files_hidden" # lib/steps.js:826
   - And I tap to continue # lib/steps.js:747
   - Then the question id should be "group of complex fields" # lib/steps.js:473
   - And I set the var "double_quote_dict['double_quote_key']['dq_two']" to "true" # lib/steps.js:798
   - And I set the var "single_quote_dict['single_quote_key']['sq_two']" to "true" # lib/steps.js:798
   - And I tap to continue # lib/steps.js:747
   - Then the question id should be "direct standard fields" # lib/steps.js:473
   - When I set the var "checkboxes_yesno" to "True" # lib/steps.js:798
   - And I set the var "checkboxes_other['checkbox_other_opt_1']" to "true" # lib/steps.js:798
   - And I set the var "dropdown_test" to "dropdown_opt_2" # lib/steps.js:798
   - And I set the var "radio_yesno" to "False" # lib/steps.js:798
   - And I set the var "radio_other" to "radio_other_opt_3" # lib/steps.js:798
   - And I set the var "text_input" to "Regular text input field value" # lib/steps.js:798
   - And I set the var "textarea" to "Multiline text\narea value" # lib/steps.js:798
   - And I set the var "date_input" to "today - 1" # lib/steps.js:798
   - When I set the var "direct_standard_fields" to "True" # lib/steps.js:798
   - Then the question id should be "showifs" # lib/steps.js:473
   - When I set the var "show_2" to "True" # lib/steps.js:798
   - And I set the var "show_3" to "True" # lib/steps.js:798
   - And I set the var "showif_checkbox_yesno" to "true" # lib/steps.js:798
   - And I set the var "showif_checkboxes_other['showif_checkboxes_nota_2']" to "true" # lib/steps.js:798
   - And I set the var "showif_checkboxes_other['showif_checkboxes_nota_3']" to "true" # lib/steps.js:798
   - And I set the var "showif_yesnoradio" to "True" # lib/steps.js:798
   - And I set the var "showif_radio_other" to "showif_radio_multi_2" # lib/steps.js:798
   - And I set the var "showif_text_input" to "Show if text input value" # lib/steps.js:798
   - And I set the var "showif_textarea" to "Show if\nmultiline text\narea value" # lib/steps.js:798
   - And I set the var "showif_dropdown" to "showif_dropdown_1" # lib/steps.js:798
   - When I tap to continue # lib/steps.js:747
   - Then the question id should be "object checkboxes" # lib/steps.js:473
   - And I set the var "object_checkboxes_test['obj_chkbx_opt_1']" to "true" # lib/steps.js:798
   - And I set the var "object_dropdown" to "obj_opt_2" # lib/steps.js:798
   - When I tap to continue # lib/steps.js:747
   - Then the question id should be "buttons yesnomaybe" # lib/steps.js:473
   - When I set the var "buttons_yesnomaybe" to "None" # lib/steps.js:798
   - Then the question id should be "buttons other" # lib/steps.js:473
   - When I set the var "buttons_other" to "button_3" # lib/steps.js:798
   - Then the question id should be "button continue" # lib/steps.js:473
   - When I set the var "button_continue" to "True" # lib/steps.js:798
   - Then the question id should be "screen features" # lib/steps.js:473
   - When I tap to continue # lib/steps.js:747
   - Then the question id should be "proxy vars" # lib/steps.js:473
   - When I set the var "x[i].name.first" to "Proxyname1" # lib/steps.js:798
   - And I tap to continue # lib/steps.js:747
   - Then the question id should be "is there another generic" # lib/steps.js:473
   - When I set the var "x.there_is_another" to "True" # lib/steps.js:798
   - Then the question id should be "proxy vars" # lib/steps.js:473
   - When I set the var "x[i].name.first" to "Proxyname2" # lib/steps.js:798
   - And I tap to continue # lib/steps.js:747
   - Then the question id should be "is there another generic" # lib/steps.js:473
   - When I set the var "x.there_is_another" to "False" # lib/steps.js:798
   - Then the question id should be "first signature" # lib/steps.js:473
   - When I sign # lib/steps.js:847
   - And I tap to continue # lib/steps.js:747
   - Then the question id should be "second signature" # lib/steps.js:473
   - When I sign # lib/steps.js:847
   - And I tap to continue # lib/steps.js:747
   - Then the question id should be "simple doc" # lib/steps.js:473
   - When I download "simple-doc.pdf" # lib/steps.js:854
   - And I tap to continue # lib/steps.js:747
   - Then the question id should be "button event action" # lib/steps.js:473
   - And I tap to continue # lib/steps.js:747
   - Then the question id should be "single-dropdown-field" # lib/steps.js:473
   - And I set the var "single_dropdown_field" to "email" # lib/steps.js:798
   - And I tap to continue # lib/steps.js:747
   - Then the question id should be "the end" # lib/steps.js:473
   ✖ After # lib/steps.js:952
       Error: Execution context was destroyed, most likely because of a navigation.
           at rewriteError (/Users/Comfadorable/Desktop/General/coding/suffolk-lit-lab-assembly-line/docassemble-cucumber/node_modules/puppeteer/lib/cjs/puppeteer/common/ExecutionContext.js:265:23)
           at processTicksAndRejections (node:internal/process/task_queues:96:5)
           at async ExecutionContext._evaluateInternal (/Users/Comfadorable/Desktop/General/coding/suffolk-lit-lab-assembly-line/docassemble-cucumber/node_modules/puppeteer/lib/cjs/puppeteer/common/ExecutionContext.js:165:64)
           at async /Users/Comfadorable/Desktop/General/coding/suffolk-lit-lab-assembly-line/docassemble-cucumber/node_modules/puppeteer/lib/cjs/puppeteer/common/DOMWorld.js:106:30
           at async DOMWorld.$eval (/Users/Comfadorable/Desktop/General/coding/suffolk-lit-lab-assembly-line/docassemble-cucumber/node_modules/puppeteer/lib/cjs/puppeteer/common/DOMWorld.js:117:26)
           at async Object.examinePageID (/Users/Comfadorable/Desktop/General/coding/suffolk-lit-lab-assembly-line/docassemble-cucumber/lib/scope.js:1738:23)
           at async Object.getSafeScenarioFilename (/Users/Comfadorable/Desktop/General/coding/suffolk-lit-lab-assembly-line/docassemble-cucumber/lib/scope.js:236:20)
           at async World.<anonymous> (/Users/Comfadorable/Desktop/General/coding/suffolk-lit-lab-assembly-line/docassemble-cucumber/lib/steps.js:979:33)

Warnings:

1) Scenario: I set various values (attempt 1, retried) # tests/features/interactive_steps.feature:7
   ✔ Before # lib/steps.js:96
   ✔ Given I start the interview at "all_tests" # lib/steps.js:137
   ✔ Then the question id should be "upload files" # lib/steps.js:473
   ✔ When I set the var "upload_files_visible" to "some_png_1.png, some_png_2.png" # lib/steps.js:798
   ✔ And I set the var "show_upload" to "True" # lib/steps.js:798
   ✔ And I upload "some_png_2.png,some_png_1.png" to "upload_files_hidden" # lib/steps.js:826
   ✔ And I tap to continue # lib/steps.js:747
   ✔ Then the question id should be "group of complex fields" # lib/steps.js:473
   ✔ And I set the var "double_quote_dict['double_quote_key']['dq_two']" to "true" # lib/steps.js:798
   ✔ And I set the var "single_quote_dict['single_quote_key']['sq_two']" to "true" # lib/steps.js:798
   ✔ And I tap to continue # lib/steps.js:747
   ✔ Then the question id should be "direct standard fields" # lib/steps.js:473
   ✔ When I set the var "checkboxes_yesno" to "True" # lib/steps.js:798
   ✔ And I set the var "checkboxes_other['checkbox_other_opt_1']" to "true" # lib/steps.js:798
   ✔ And I set the var "dropdown_test" to "dropdown_opt_2" # lib/steps.js:798
   ✔ And I set the var "radio_yesno" to "False" # lib/steps.js:798
   ✔ And I set the var "radio_other" to "radio_other_opt_3" # lib/steps.js:798
   ✔ And I set the var "text_input" to "Regular text input field value" # lib/steps.js:798
   ✔ And I set the var "textarea" to "Multiline text\narea value" # lib/steps.js:798
   ✔ And I set the var "date_input" to "today - 1" # lib/steps.js:798
   ✔ When I set the var "direct_standard_fields" to "True" # lib/steps.js:798
   ✔ Then the question id should be "showifs" # lib/steps.js:473
   ✔ When I set the var "show_2" to "True" # lib/steps.js:798
   ✔ And I set the var "show_3" to "True" # lib/steps.js:798
   ✔ And I set the var "showif_checkbox_yesno" to "true" # lib/steps.js:798
   ✔ And I set the var "showif_checkboxes_other['showif_checkboxes_nota_2']" to "true" # lib/steps.js:798
   ✔ And I set the var "showif_checkboxes_other['showif_checkboxes_nota_3']" to "true" # lib/steps.js:798
   ✔ And I set the var "showif_yesnoradio" to "True" # lib/steps.js:798
   ✔ And I set the var "showif_radio_other" to "showif_radio_multi_2" # lib/steps.js:798
   ✔ And I set the var "showif_text_input" to "Show if text input value" # lib/steps.js:798
   ✔ And I set the var "showif_textarea" to "Show if\nmultiline text\narea value" # lib/steps.js:798
   ✔ And I set the var "showif_dropdown" to "showif_dropdown_1" # lib/steps.js:798
   ✖ When I tap to continue # lib/steps.js:747
       Error: Action did not complete within 30000 milliseconds
           at Timeout.<anonymous> (/Users/Comfadorable/Desktop/General/coding/suffolk-lit-lab-assembly-line/docassemble-cucumber/lib/cucumber_8_shim.js:18:14)
           at listOnTimeout (node:internal/timers:559:17)
           at processTimers (node:internal/timers:502:7)
   - Then the question id should be "object checkboxes" # lib/steps.js:473
   - And I set the var "object_checkboxes_test['obj_chkbx_opt_1']" to "true" # lib/steps.js:798
   - And I set the var "object_dropdown" to "obj_opt_2" # lib/steps.js:798
   - When I tap to continue # lib/steps.js:747
   - Then the question id should be "buttons yesnomaybe" # lib/steps.js:473
   - When I set the var "buttons_yesnomaybe" to "None" # lib/steps.js:798
   - Then the question id should be "buttons other" # lib/steps.js:473
   - When I set the var "buttons_other" to "button_3" # lib/steps.js:798
   - Then the question id should be "button continue" # lib/steps.js:473
   - When I set the var "button_continue" to "True" # lib/steps.js:798
   - Then the question id should be "screen features" # lib/steps.js:473
   - When I tap to continue # lib/steps.js:747
   - Then the question id should be "proxy vars" # lib/steps.js:473
   - When I set the var "x[i].name.first" to "Proxyname1" # lib/steps.js:798
   - And I tap to continue # lib/steps.js:747
   - Then the question id should be "is there another generic" # lib/steps.js:473
   - When I set the var "x.there_is_another" to "True" # lib/steps.js:798
   - Then the question id should be "proxy vars" # lib/steps.js:473
   - When I set the var "x[i].name.first" to "Proxyname2" # lib/steps.js:798
   - And I tap to continue # lib/steps.js:747
   - Then the question id should be "is there another generic" # lib/steps.js:473
   - When I set the var "x.there_is_another" to "False" # lib/steps.js:798
   - Then the question id should be "first signature" # lib/steps.js:473
   - When I sign # lib/steps.js:847
   - And I tap to continue # lib/steps.js:747
   - Then the question id should be "second signature" # lib/steps.js:473
   - When I sign # lib/steps.js:847
   - And I tap to continue # lib/steps.js:747
   - Then the question id should be "simple doc" # lib/steps.js:473
   - When I download "simple-doc.pdf" # lib/steps.js:854
   - And I tap to continue # lib/steps.js:747
   - Then the question id should be "button event action" # lib/steps.js:473
   - And I tap to continue # lib/steps.js:747
   - Then the question id should be "single-dropdown-field" # lib/steps.js:473
   - And I set the var "single_dropdown_field" to "email" # lib/steps.js:798
   - And I tap to continue # lib/steps.js:747
   - Then the question id should be "the end" # lib/steps.js:473
   ✔ After # lib/steps.js:952

1 scenario (1 failed)
69 steps (1 failed, 68 skipped)
```

lib/scope.js
plocket: Hmm, there was a reason I was doing this, but I can't remember why. Maybe it was something about clarity, but that seems insufficient. Maybe it was to keep track of it globally for some reason. You may be right.

lib/scope.js
plocket: Hmm, I was worried that a while loop would block execution, but it seems to not do that (probably because it's inside an async function), so a while loop it will be!

plocket:
A new error started showing up at some point. I wrote it up at https://github.com/SuffolkLITLab/ALKiln/issues/585, but haven't been able to track it down so far.

plocket:
This doesn't yet take care of download reloading problems. [The worst consequence is that the tests won't wait properly for the reload, which could mess up the next test.] ~but that honestly just~ It also means that those won't get the right error message in the report or console, but that's not as big a deal. [That should still be made into an issue, though.

We did say this PR doesn't have to take care of everything.]

lib/steps.js
```diff
@@ -1054,28 +1084,34 @@ After(async function(scenario) {
     await scope.addToReport(scope, { type: `outcome`, value: `**-- Scenario Failed --**` });
     if ( scope.page ) {
       if ( scope.disable_error_screenshot ) {
+      // if ( scope.disable_error_screenshot && !scope.disable_error_screenshot ) {
```
```suggestion
```

lib/steps.js
```diff
@@ -1054,28 +1084,34 @@ After(async function(scenario) {
...
+          let { id } = await scope.examinePageID( scope, 'none to match' );
+          screenshot_name += `-${ id }`;
+          await scope.page.screenshot({
+            path: `${ scope.paths.scenario }/${ screenshot_name }.jpg`,
+            type: `jpeg`,
+            fullPage: true
+          });
+        } catch ( error ) {
+          throw error;
+        }
```
BryceStevenWilley: Given that you're immediately throwing the caught error, I don't think that this needs to be wrapped in a try-catch, right?

lib/scope.js
```diff
@@ -2111,7 +2290,7 @@ module.exports = {
       if ( !link ) { await scope.addToReport(scope, { type: `error`, value: msg }); }
       expect( link, msg ).to.exist;
 
-      await link.evaluate( elem => { return elem.click(); });
+      await scope.tapElement( scope, elem );
```
```suggestion
      await scope.tapElement( scope, link );
```
Because `elem` isn't defined in this function, so I'm assuming it's copy-pasta from somewhere else.

lib/scope.js
```diff
@@ -2248,13 +2446,15 @@ module.exports = {
...
+
       expect( email, email_msg ).to.not.equal( undefined );
       expect( password, password_msg ).to.not.equal( undefined );
 
       // Try to log in
       await scope.page.type( `#email`, email );
       await scope.page.type( `#password`, password );
-      await scope.page.$eval( `button[name="submit"]`, (elem) => { return elem.click(); });
+      let elem = await scope.page.$( `button[name="submit"]` );
+      await scope.tapElement( scope, elem );
```
BryceStevenWilley, line 2519: Minor and not worth changing, but just wanted to note that this refactoring of grabbing things and passing them into tapElement is slower, since (from my understanding) it makes puppeteer make 2 separate calls to the chrome engine instead of one. Shouldn't be a huge time loss though.

lib/scope.js
```diff
@@ -413,6 +429,147 @@ module.exports = {
...
+  stop_tracking_server_status: false,
+  track_server_status: async function( scope ) {
+    /** Updates server_statuses list by checking if server is responding until,
+    *    eventually, it is told to stop. */
+
+    while ( !scope.stop_tracking_server_status ) {
+      // This code assumes a reasonable response time would be 2 seconds if the
+      // server is up and that a server reload/restart won't take less than 2 seconds.
+      // Also, it's non-blocking, so it doesn't slow down tests
+      let is_up = await server.is_server_responding({ timeout: scope.server_check_interval_ms });
```
BryceStevenWilley, line 446: I think this assumes that `is_server_responding` will always take `timeout` ms to respond, but it doesn't have to take `timeout` seconds. When it returns sooner, we just happily check if it's responding at a faster rate, which I don't think is what we want.

Not sure the javascripty way of doing things, but I would likely get the current time before we call the function, get how long did pass after we call it (current time again - before time), and then sleep for that long.

lib/scope.js
```diff
@@ -413,6 +429,147 @@ module.exports = {
...
+      // Add status to the beginning of the list to make truncating easier
+      scope.server_statuses.unshift( is_up );
+
+      // To make it easier to check past server statuses, remove defunct server statuses
+      let max_length = Math.round( scope.timeout/scope.server_check_interval_ms );
+      // Trying to extract more items than the actual length is ok
+      scope.server_statuses = scope.server_statuses.slice( 0, max_length );
```
BryceStevenWilley, line 461: Just to clarify, we want exactly 1 element per second in here, as opposed to just a constant number of elements?

lib/scope.js
```diff
@@ -585,18 +742,36 @@ module.exports = {
...
-    // If there is no browser open, start a new one
-    if (!scope.browser) {
-      scope.browser = await scope.driver.launch({ headless: !session_vars.get_debug(), devtools: session_vars.get_debug() });
-    }
-    if ( !scope.page ) { scope.page = await scope.browser.newPage(); }
-
     if ( !file_name.includes('.yml') ) { file_name = `${ file_name }.yml` }
     let base_url = await get_base_interview_url();
-    let interview_url = `${ base_url }${ file_name }`;
+    let interview_url = `${ base_url }${ file_name }`;;
```
```suggestion
    let interview_url = `${ base_url }${ file_name }`;
```

lib/steps.js
```diff
@@ -132,47 +154,49 @@ Before(async (scenario) => {
...
+  //           reject( err );
+  //         } else { resolve( false ); }  // Give it another shot
+  //       } finally {
+  //         clearTimeout( page_timeout );  // prevent temporary hang at end of tests
+  //       }
+  //     })();
+  //   });  // ends loaded promise
+  //   await loaded;
+  //   if ( loaded === true ) { break; }
+  // }  // ends while attempting to load page
```
BryceStevenWilley: Remove commented code

BryceStevenWilley (commented):
Generally looks good, only blocker is the fact we might be querying the server for responses too quickly.

lib/scope.js
plocket: Would we pass selectors to `.tapElement()` instead? I'd have to look more closely at the code to see if that's feasible, but it might work. Not for this PR, but we could make it an issue.

lib/scope.js
plocket: Yes. I was assuming my 'waiting' math was right for checking once per second, but as you've spotted, thankfully, that particular wait time wasn't doing what I thought it was.

This is part of the previous comment, right? - one about the `is_server_responding` timeout? If there's another issue to address here, let me know.

BryceStevenWilley: It's separate from that; I'm fairly confident that we do want one status per second. This question was specifically if we wanted "wait time" number of elements in the statuses array, or if we just wanted a fixed number of things in there. If we always have i.e. 30 elements in the array, no matter what `scope.timeout` is, then the code gets a bit simpler, that's all.

plocket: The number of elements does need to cover the whole timeout that the developer wants. If they gave a timeout of 2 min, I believe we need to check whether the server reloaded during any of that time.

I'm not sure I can find a good way to make an example, but I'll try. Suppose we do only store the last 30 items in a hard-coded way. Suppose the reload happened during the first 15 seconds of it, leaving the "user" hanging out waiting for a timeout screen, and then the server managed to finish reloading during the next minute, and it then left the "user" with an error screen later or some such (one of our servers waits 5 min or something). If we only store the last 30 seconds, it would look like the server was up that whole time and that reloading wasn't part of the issue. If we store the whole 2 min, we'd see that the server did in fact reload and this error could be a reload issue.

Not sure if that made any sense.

BryceStevenWilley: I get the idea, thanks. But this part made me more confused about other things:

> Suppose the reload happened during the first 15 seconds of it, leaving the "user" hanging out waiting for a timeout screen, and then the server managed to finish reloading during the next minute, and it then left the "user" with an error screen later or some such (one of our servers waits 5 min or something)

Is this saying that if there's some / any error we'll just restart, assuming it was caused by the reload? So if there is some non-deterministic error happening but it occurs close to a restart it'll be lost?

plocket: No, sorry it's not clear. This all only pops off if there's a timeout error.  If a server reload was detected during the timeout, the developer will be informed that the server was reloading. Otherwise, they will get a different timeout error message. I believe that for a timeout error, if the server wasn't reloading there is nothing else we can detect. There's [meant to be] a screenshot [if possible] if no secret values were used. Can you think of more we can do or detect for timeout errors?

BryceStevenWilley: Gotcha, apologies for the confusion. I can't think of any additional information that we could get there.

lib/scope.js
plocket: I made [a commit](https://github.com/SuffolkLITLab/ALKiln/pull/570/commits/17cc7ec3d104767f351ba6eb6d6bc790bc6611f2) addressing this. Do you think it's a better approach?

lib/scope.js
plocket: #592

BryceStevenWilley: We'd want to check if it's actually slowing things down first, but an issue is fine

lib/scope.js
BryceStevenWilley: Yep, that looks good to me!

plocket: Is this ready for approval?

BryceStevenWilley: sorry, yeah

BryceStevenWilley (approved):

Issue #572, opened 2022-07-01
Random input Step MVP: Get through a simple form by generating random input


plocket:
MVP:
- [x] input field
- [x] textarea
- [x] checkbox
- [x] radio button
- [x] continue button

Stretch:
- [x] date field
- [x] number field
- [x] Custom birthdate
- [x] Custom three-part-date

Others?

Edit:

(#591)

Left to do for MVP:
- [x] Give each screenshots folder a time-stamped name in case the dev gets clever and use the same Scenario to run multiple randomized tests.
- [x] Make sure we can fail on actual errors - situations where there's a complete page or server error. Includes making an interview to test that is missing a variable definition or something.

plocket:
MVP closed by #591

PR #576, opened 2022-07-08
Address #572, random input testing. Not MVP.

plocket:
~Possibly an MVP for random input testing.~ Enters random input for an interview until it can't continue anymore or reaches a maximum number of pages. No`.feature` files yet - for now just a set of screenshots for every page visited. Does handle some of our AL custom data type fields.

Not sure how it would handle a system error or how to test that. Maybe have a test where we try to show a non-existent template file?

Can't pass tests until https://github.com/plocket/docassemble-ALAutomatedTestingTests/pull/193 is merged.

Addresses #572 .

Left for MVP:
- Make sure we can fail on actual errors. Make an interview to test that is missing a variable definition or something.
- Give each screenshots folder a time-stamped name in case the dev gets clever and use the same Scenario to run multiple randomized tests.

### DIFF
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -4,6 +4,8 @@ const path = require('path');
 const fs = require('fs');
 const { AxePuppeteer } = require('@axe-core/puppeteer');
 const safe_filename = require("sanitize-filename");
+// "en" is faster to import: https://github.com/faker-js/faker/issues/1114#issuecomment-1169532948
+const faker = require('@faker-js/faker/locale/en').faker;
 
 // Ours
 const session_vars = require('./utils/session_vars');
@@ -12,7 +14,6 @@ const time = require('./utils/time');
 const log = require('./utils/log');
 const { waitForTimeout } = require('./utils/time');
 const files = require('./utils/files' );
-const faker = require('@faker-js/faker').faker;
 
 
 let ordinal_to_integer = {
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -12,6 +12,7 @@ const time = require('./utils/time');
 const log = require('./utils/log');
 const { waitForTimeout } = require('./utils/time');
 const files = require('./utils/files' );
+const faker = require('@faker-js/faker').faker;
 
 
 let ordinal_to_integer = {
@@ -488,7 +489,7 @@ module.exports = {
 
     // Do your own waiting if you need to in the step
     return to_manipulate;
-  },  // Ends setVar()
+  },  // Ends scope.setVar()
 
   continue: async function ( scope ) {
     /* Presses whatever button it finds that might lead to the next page. */
@@ -500,7 +501,16 @@ module.exports = {
     await elem.evaluate( el => { return el.className });
     // Waits for navigation or user error
     await scope.tapElement( scope, elem );
-  },
+  },  // Ends scope.continue()
+
+  continue_exists: async function ( scope ) {
+    let elem = await Promise.race([
+      scope.page.$( `fieldset.da-field-buttons button[type="submit"]` ),  // other pages (this is the most consistent way)
+      scope.page.$( `fieldset .dasigsave` ),  // signature page
+    ]);
+
+    return elem !== null;
+  },  // Ends scope.continue_exists()
 
   tapTab: async function ( scope, tab_id ) {
     let elem = await scope.page.$(`#${tab_id}`); 
@@ -967,6 +977,9 @@ module.exports = {
       // names.var_names.push( layer_2_decoded );
     }
 
+    // Detecting valid variable names/characters
+    // https://stackoverflow.com/a/23377268
+
     return var_names;
   },  // Ends scope.getPossibleVarNames()
 
@@ -1233,13 +1246,15 @@ module.exports = {
       await scope.page.waitForSelector('#dasigcanvas');
     }
 
-    let handle = await scope.page.evaluateHandle(( selector, field ) => {
-      let elem = document.querySelector( selector );
-      // `elem` has never been null
-      if ( elem.nextSibling && elem.nextSibling.tag === `label` ) {
-        return elem.nextSibling;
-      } else { return elem; }
-    }, field.selector, field );
+    // let handle = await scope.page.evaluateHandle(( selector, field ) => {
+    //   let elem = document.querySelector( selector );
+    //   // `elem` has never been null
+    //   if ( elem.nextSibling && elem.nextSibling.tag === `label` ) {
+    //     return elem.nextSibling;
+    //   } else { return elem; }
+    // }, field.selector, field );
+
+    let handle = await scope.get_handle_from_selector( scope, { field: field });
 
     let disabled = await handle.evaluate(( elem )=> {
       // da's header covers the top of the section containing the fields which means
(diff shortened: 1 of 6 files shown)


lib/steps.js
```diff
@@ -780,6 +780,109 @@ When(/I check all pages for (?:accessibility|a11y) issues/i, {timeout: -1}, asyn
...
+            // stop or continue for the next loop
+            buttons_exist = await scope.continue_exists( scope );
+            if ( buttons_exist ) {
+              await scope.continue( scope );
+            }  // ends if continue exists
+          }
+
+          if ( buttons_exist ) {
+            process.stdout.write(`\x1b[36m${ 'v' }\x1b[0m`);  // pressed button
+          }
```
BryceStevenWilley, line 855: What is this doing? It looks like it's just printing `v` to the screen, is it for the progress output?

lib/steps.js
```diff
@@ -780,6 +780,109 @@ When(/I check all pages for (?:accessibility|a11y) issues/i, {timeout: -1}, asyn
...
+    await custom_timeout;
+
+  }  // ends while continue_exists
+
+  // // Take a screenshot of the last page
+  // await scope.page.screenshot({
+  //   path: `${ scope.paths.random_screenshots }/${ Date.now() }.jpg`,
+  //   type: `jpeg`,
+  //   fullPage: true
+  // });
```
```suggestion
```

lib/scope.js
```diff
@@ -12,6 +12,7 @@ const time = require('./utils/time');
 const log = require('./utils/log');
 const { waitForTimeout } = require('./utils/time');
 const files = require('./utils/files' );
+const faker = require('@faker-js/faker').faker;
```
BryceStevenWilley: Haven't tested myself, but apparently `faker-js/faker/locale/en` is [much faster to import](https://github.com/faker-js/faker/issues/1114#issuecomment-1169532948), though it might not affect us. I'll try to check if it does affect us soon.

lib/scope.js
```diff
@@ -500,7 +501,32 @@ module.exports = {
...
-  },
+  },  // Ends scope.continue()
+
+  continue_exists: async function ( scope ) {
+    /** Return true if a button that will allow continuing appears
+    *    on this page. Otherwise return false. */
+
+    // PROBLEM: because we don't have a page id that can stop us from moving
+    // forward. We have to detect a continue button, but that may be impossible
+    // TODO: Story table requires a question id or it similarly might not be able to stop
```
BryceStevenWilley: Not sure I understand why the `page_id` issue is here. It affects what might call this, sure, but it doesn't affect this function specifically.

lib/scope.js
```diff
@@ -500,7 +501,32 @@ module.exports = {
...
+    // <button type="submit" class="btn btn-da btn-primary" name="Zm9v" value="True">Continue</button>
+    // `buttons:` can be used in question blocks as choices
+    let regular = await scope.page.$( `fieldset.da-field-buttons button[type="submit"]` );
+    let signature = await scope.page.$( `fieldset .dasigsave` );
+
+    if ( regular !== null || signature !== null ) {
+      return true;
+    } else {
+      return false;
+    }
```
```suggestion
    return regular !== null || signature !== null;
```

lib/scope.js
```diff
@@ -2170,10 +2273,244 @@ module.exports = {
...
+      // Get the values of all the options of the field
+      let handle = await scope.get_handle_from_selector( scope, { field: field });
+      // Get the values of all the choices
+      let answer_choices = await handle.evaluate(function ( elem ) {
+        let values = Array.from(elem.querySelectorAll(`option`)).map(element=>element.value);
+        return values;
+      });
+
+      // Remove the "no selection" option from the list ('')
+      // Should we allow '' as a valid choice - no selection?
```
BryceStevenWilley: IMO we shouldn't; only if it's an optional field, and that's def out of scope for MVP

lib/scope.js
```diff
@@ -2170,10 +2273,244 @@ module.exports = {
...
+      } else {
+
+        // `file` type inputs come next to an input with a type of ''.
+        // Maybe others do too.
+        if ( type !== `` && type !== `hidden` ) {
+          // Field types that we can't yet handle, but plan to in future
+          if ( type === `file` ) {
+            await scope.addToReport( scope, {
+              type: `warning`,
+              value: `Sorry, ALKiln cannot yet handle "${ type }" fields.`
```
```suggestion
              value: `Sorry, ALKiln's random-input cannot yet handle "${ type }" fields.`
```

lib/scope.js
```diff
@@ -2170,10 +2273,244 @@ module.exports = {
...
+
+        // * @param fields - data representing one DOM field
+        // * @param fields[n].selector {str} - used to find the DOM node
+        // * @param fields[n].tag {str} - HTML tag of node
+        // * @param fields[n].type {str} - HTML type of node if it has one
+        // * @param fields[n].guesses {arr} - arr of objects that could
+        // *    possibly match a table row's `var` column
+        // * @param fields[n].guesses[n].var {str} - name of variable
+        // * @param fields[n].guesses[n].value {str} - value of a choice
+        // * @param <fields[n].guesses[n].trigger> {str} - var name that triggered this page
```
BryceStevenWilley: Probably useful for writing the code, but doesn't need to be here.  You could leave the `field[n].tag` explanation if you wanted, since it's the only attribute used here.
```suggestion
        // fields[n].tag is HTML tag of node, as a string
```

lib/scope.js
```diff
@@ -1331,18 +1403,36 @@ module.exports = {
...
+  enter_answer: {
     button: async function ( scope, { handle, set_to }) { await scope.tapElement( scope, handle ); },
     textarea: async function ( scope, { handle, set_to }) {
       await scope.setText( scope, { handle, set_to });
       await scope.afterStep(scope);  // No showifs for this one?
     },
     select: async function ( scope, { handle, set_to }) {
+      // Note: I don't remember what's really going on in here.
+      // I will note that values of `select` elements are strings,
+      // not variables, so we don't need to check for variable names.
```
BryceStevenWilley, line 1427: Unless it's an object dropdown, which will have variable names as the option values.

lib/scope.js
```diff
@@ -1260,15 +1283,39 @@ module.exports = {
...
 
-    // Detect a custom type and get the element of that custom type if it exists
+    await scope.funnel_the_answer( scope, { handle, set_to, field });
+
+    return row;
+  },  // Ends scope.setVariable()
+
+  get_handle_from_selector: async function ( scope, { field }) {
+    /** Get the element's puppeteer handle using the field's selector. */
+
```
BryceStevenWilley: Nit: a better name for this would be something like `get_handle_from_field`. Selector is too vague to be useful, as anything on the page could be a selector. Saying what the selector will get you is good though.

lib/scope.js
```diff
@@ -1287,27 +1334,52 @@ module.exports = {
...
 
-    // Set field
-    if ( scope.setCustomDatatype[ custom_datatype ] === undefined ) {
-      // If no custom datatype, set a regular field
-      await scope.setField[ tag ]( scope, { handle, set_to });
-    } else {
+    return custom_datatype;
+  },  // Ends scope.is_custom_datatype()
+
+  // TODO: change `set_to` to the name `answer`
```
BryceStevenWilley: +1 for this change, can we do it in this PR?

BryceStevenWilley (commented):
Did a quick review, haven't started using yet, but looks exciting!

A handful of comments, mostly nits though I think.

lib/steps.js
plocket: Yes, progress output. Hoping we get the progress output to print in GitHub too, but it's at least useful locally. This is for a button being pressed. The '*' is for an input being interacted with.

lib/scope.js
plocket: Great find! How the heck did you find this? Thanks!

lib/scope.js
plocket:
```suggestion
```

lib/scope.js
plocket: Variable names encoded in base64? I'll have to check on Slack. If so, I think we may need a new test to see if we need to sometimes encode `select` values.

lib/scope.js
plocket:
```suggestion
      // TODO: blue sky - for optional fields allow no selection
```

lib/scope.js
BryceStevenWilley: Haha, I found it when randomly clicking through the faker library when I saw we were using it, just to get a sense of it. Definitely wasn't looking for it, lol.

lib/scope.js
BryceStevenWilley: Yep! After making this comment, I actually ran into an bug that shows we don't handle this correctly, PR should be incoming.

lib/steps.js
BryceStevenWilley: Okay. Probably for another PR, but it might make sense to move those to the logging file.

lib/scope.js
plocket: I'll just add it. Why not :)

BryceStevenWilley (approved):
LGTM! 🎉

PR #577, opened 2022-07-13
Prints the page id that story table failed on

BryceStevenWilley:
Gets the last available page id when there's a missing variable. Much more
useful at a glance when trying to figure out where your problem is.

Personally, I've been running into `ProtocolError: Protocol
error (Page.captureScreeshot): Unable to capture screenshot`, which means
there's no screen shot available. Without this change, it's impossible to tell
what page the missing variable was on.

### DIFF
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -1788,7 +1788,7 @@ module.exports = {
       }
       if ( num_unused_rows > 0 ) {
         let { id } = await scope.examinePageID( scope, 'none to match' );
-        let msg = `Missing variable or variables on page (${id})`;
+        let msg = `Missing variable or variables on page "${id}"`;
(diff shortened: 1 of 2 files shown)


lib/scope.js
```diff
@@ -1691,7 +1691,8 @@ module.exports = {
         }
       }
       if ( num_unused_rows > 0 ) {
-        let msg = `Missing variable or variables on page.`;
+        let { id } = await scope.examinePageID( scope, 'none to match' );
+        let msg = `Missing variable or variables on page (${id})`;
```
```suggestion
        let msg = `Missing variable or variables on page "${id}"`;
```

BryceStevenWilley:
Looked over and LGTM'd on a zoom call (~20-30 into the stream)!

plocket (approved):
Looks great! Merge when tests pass.

PR #578, opened 2022-07-13
Normalize a table that doesn't have the header row

BryceStevenWilley:
We require the header row to set the right attributes (`var`, `value`, and
`trigger`) to the row objects. If you don't have the header row, you'll get an
obscure error, usually that "no variables" matched. This gives a much clearer
error and can be resolved more quickly

### DIFF
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -671,7 +671,9 @@ module.exports = {
         var_data = [];
         for (let row_raw of raw_array) {
           if (row_raw.length < 2) {
-            let bad_table_msg = `Your Story Table definition needs to be changed: the first row should be "| var | value | trigger |", but it's actually ${ raw_array[0] }`;
+            let bad_table_msg = `Your Story Table definition needs to be changed. The best practice`
+                + ` is to make your table 3 columns wide with a first row in this format: "| var | value | trigger |".`
+                + ` The first row of your table is actually "${ raw_array[0] }"`;
             await scope.addToReport( scope, { type: `error`, value: bad_table_msg})
             throw Error(bad_table_msg);
           } else if (row_raw.length == 2) {
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -657,6 +657,11 @@ module.exports = {
     // Support tables with no 'trigger' column
     let supported_table = [];
     for ( let row of var_data ) {
+      if (!('var' in row) || !('value' in row)) {
+        let bad_table_msg = `Your table definition is bad: the first row should be var | value | trigger`;
+        await scope.addToReport( scope, { type: `error`, value: bad_table_msg})
+        throw ReferenceError(bad_table_msg);
+      }
       // See tests > unit_tests > tables.fixtures.js > tables.old_to_current_formatting
       let actual_var_value = row.value;
       if (is_secret_var(row)) {
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -650,18 +650,38 @@ module.exports = {
     }
   },  // Ends scope.setLanguage()
 
-  normalizeTable: async function ( scope, { var_data, from_story_table=true }) {
-    /* Return data derived from cucumber variable-setting data.
-     * Table must have headers. Arrays are not supported. */
+  normalizeTable: async function ( scope, { var_data=null, raw_var_data=null, from_story_table=true }) {
+    /* Return data derived from cucumber variable-setting data, from either an array or object format */
+
+    if (!var_data && ! raw_var_data) {
+      let no_data_msg = `ALKiln Internal error: normalizeTable called without var_data or raw_var_data`;
+      await scope.addToReport( scope, { type: `error`, value: no_data_msg });
+      throw Error(bad_table_msg);
+    }
+
+    // Always use the raw var data if present; it has the full first row
+    if (raw_var_data) {
+      var_data = raw_var_data.hashes();
+      if (!('var' in var_data[0]) || !('value' in var_data[0])) {
+        let raw_array = raw_var_data.raw();
+        var_data = [];
+        for (let row_raw of raw_array) {
+          if (row_raw.length < 2) {
+            let bad_table_msg = `Your Story Table definition is bad: the first row should be "| var | value | trigger |", but it's actually ${ raw_array[0] }`;
+            await scope.addToReport( scope, { type: `error`, value: bad_table_msg})
+            throw Error(bad_table_msg);
+          } else if (row_raw.length == 2) {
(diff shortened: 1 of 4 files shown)


lib/scope.js
```diff
@@ -637,6 +637,11 @@ module.exports = {
     // Support tables with no 'trigger' column
     let supported_table = [];
     for ( let row of var_data ) {
+      if (!('var' in row) || !('value' in row)) {
```
BryceStevenWilley: (there should be an issue for this, but) consider just directly getting the first column as "var", the second as "value", etc.

plocket:
Just FYI, in the future (I think we have an issue) we can probably transform a table without a header row by using `.raw()`:

https://github.com/cucumber/cucumber-js/blob/main/docs/support_files/data_table_interface.md

lib/scope.js
```diff
@@ -637,6 +637,11 @@ module.exports = {
...
+        let bad_table_msg = `Your table definition is bad: the first row should be var | value | trigger`;
```
```suggestion
        let bad_table_msg = `Your Story Table definition is bad: the first row should be "| var | value | trigger |" `;
```

lib/scope.js
```diff
@@ -637,6 +637,11 @@ module.exports = {
...
+        await scope.addToReport( scope, { type: `error`, value: bad_table_msg})
+        throw ReferenceError(bad_table_msg);
```
```suggestion
        throw Error(bad_table_msg);
```

BryceStevenWilley (commented):
- [ ] Use raw() to smartly get 1st column, etc.
- [ ] Add test

tests/features/story_table_formats.feature
```diff
@@ -75,3 +74,52 @@ Scenario: Table MISSING trigger column
...
+    | checkboxes_yesno | True |
+    | dropdown_test | dropdown_opt_2 |
+    | radio_other | radio_other_opt_3 |
+    | radio_yesno | False |
+    | single_quote_dict['single_quote_key']['sq_two'] | true |
+    | text_input | Regular text input field value |
+    | textarea | Multiline text\narea value |
+
+@fast @stf7
+Scenario: Table has no header row, one row only has one column
```
plocket:
```suggestion
Scenario: Fails when table has no header row, rows have only one column
```

lib/scope.js
```diff
@@ -650,9 +650,34 @@ module.exports = {
     }
   },  // Ends scope.setLanguage()
 
-  normalizeTable: async function ( scope, { var_data, from_story_table=true }) {
-    /* Return data derived from cucumber variable-setting data.
-     * Table must have headers. Arrays are not supported. */
+  normalizeTable: async function ( scope, { var_data=null, raw_var_data=null, from_story_table=true }) {
+    /* Return data derived from cucumber variable-setting data, from either an array or object format */
```
plocket: Maybe this should include a note about why some things would be passing in just `var_data` or something? Maybe not necessary, but it took me a little while to understand why that was sticking around.

lib/scope.js
```diff
@@ -650,9 +650,34 @@ module.exports = {
...
+
+    if (!var_data && ! raw_var_data) {
+      let no_data_msg = `ALKiln Internal error: normalizeTable called without var_data or raw_var_data`;
+      await scope.addToReport( scope, { type: `error`, value: no_data_msg });
+      throw Error(bad_table_msg);
```
plocket: Maybe there should be a test for this too.

```suggestion
      throw Error(no_data_msg);
```

lib/scope.js
```diff
@@ -650,9 +650,34 @@ module.exports = {
...
+
+    // Always use the raw var data if present; it has the full first row
+    if (raw_var_data) {
+      var_data = raw_var_data.hashes();
+      if (!('var' in var_data[0]) || !('value' in var_data[0])) {
+        let raw_array = raw_var_data.raw();
+        var_data = [];
+        for (let row_raw of raw_array) {
+          if (row_raw.length < 2) {
+            let bad_table_msg = `Your Story Table definition is bad: the first row should be "| var | value | trigger |", but it's actually ${ raw_array[0] }`;
```
plocket: This looks like the old "bad table msg" that was just for when the first row had the wrong column headings. It looks like this could happen if any row had only one column and that seems like it'd be a different error message.
```suggestion
            let bad_table_msg = `Your Story Table definition needs to be changed. A row in your table only has 1 column when it should have at least 2.`;
```

plocket (changes requested):
Looking like a great improvement! Some suggestions, but at least one thing I think would be a bug. Maybe useful to add a test for it as well. Thank you!

lib/scope.js
BryceStevenWilley: Add to the message that they should explicitly open an issue if they see this

lib/scope.js
BryceStevenWilley: My way of designing this is that it should be very clear what you're encouraged to do, but we should be flexible if someone made a typo, or forgot their header row. We don't have to leave this feature officially undocumented (i.e. not on the docs page, but still in changelog, etc.), but I do think that all of the error messages should assume that you're doing the standard way of doing things, like including a header row with three columns.

plocket: That's not what the if statements tests for, though, right? It would lead to inconsistent behavior from the point of view of the developer as they could have other tests that don't have a header row, but are working. That might make it hard for them to track down the actual problem.

I'm not totally against it, but it at least should have a note as to why the message doesn't reflect the logic test that triggers the message.

As a separate note, I think "is bad" is an unpleasant way for someone to experience this error.

plocket: Maybe something like:
```
`... The best practice is to make your table 3 columns wide with a first row in this format: "| var...
```

BryceStevenWilley: Agreed that "is bad" is unpleasant, I had already removed that. And I had landed on something similar to your last suggestion (I was gonna use "we don't have enough information to know how to use your Story Table"), but I like your "best practice" wording better, should be in now.

tests/features/story_table_formats.feature
```diff
@@ -75,3 +74,52 @@ Scenario: Table MISSING trigger column
...
+    | single_quote_dict['single_quote_key']['sq_two'] | true |
+    | text_input | Regular text input field value |
+    | textarea | Multiline text\narea value |
+
+@fast @stf7
+Scenario: Fails when table has no header row and rows have only one column
+  Given the final Scenario status should be "failed"
+  And the Scenario report should include:
+  """
+  Your Story Table definition needs to be changed:
```
plocket:
```suggestion
  Your Story Table definition needs to be changed.
```

plocket (approved):
Looks great! Thanks!

PR #579, opened 2022-07-13
The `da-retry` can sometimes not have a blockquote element

BryceStevenWilley:
This prevents a full crash of Kiln, though we can't print out as much info.

Sometimes necessary for me when the interview can't be loaded (if something went wrong with the playground installation, or I'm pointing at the wrong playground).

### DIFF
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -606,10 +606,14 @@ module.exports = {
     let error_id_elem = await scope.page.$( `#da-retry` );
     if ( error_id_elem ) {
       let error_elem = await scope.page.$( `blockquote` );
-      let error_handle = await error_elem.getProperty( `textContent` );
-      let system_error_text = await error_handle.jsonValue();
+      if (error_elem !== null) {
+        let error_handle = await error_elem.getProperty( `textContent` );
+        let system_error_text = await error_handle.jsonValue();
 
-      result.error = system_error_text;
+        result.error = system_error_text;
+      } else {
+        result.error = "retry error, but we can't see what it is!"
+      }
     }
 
     return result;
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -612,7 +612,7 @@ module.exports = {
 
         result.error = system_error_text;
       } else {
-        result.error = "retry error, but we can't see what it is!"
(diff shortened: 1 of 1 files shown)


plocket:
How can we recreate this so that we can add a test file? We should probably create tests for both eventualities soon.

plocket:
Also, what does a full crash mean? No more tests are run?

BryceStevenWilley:
> How can we recreate this so that we can add a test file?

It's not a consistent failure. From the docassemble source code (server.py:L21248), error messages are wrapped in `<pre>` instead of `<blockquote>` when there's a newline in the error message (I have no idea when that happens though). I'm not sure the strategy of just searching the whole page for `<pre>` will work though, we want just subelements of the `#da-retry`?

But IMO, this doesn't need a specific unit test: it's just making things more rigorous.

> Also, what does a full crash mean? No more tests are run?

Yes, at the very least that scenario stops running and exits with an inscrutable error.

plocket:
I think we might want to work on the message itself, but I agree it's useful to have a fallback message.

lib/scope.js
```diff
@@ -606,10 +606,14 @@ module.exports = {
...
-      let error_handle = await error_elem.getProperty( `textContent` );
-      let system_error_text = await error_handle.jsonValue();
+      if (error_elem !== null) {
+        let error_handle = await error_elem.getProperty( `textContent` );
+        let system_error_text = await error_handle.jsonValue();
 
-      result.error = system_error_text;
+        result.error = system_error_text;
+      } else {
+        result.error = "retry error, but we can't see what it is!"
```
```suggestion
        result.error = "Docassemble ran into an error on the page, but ALKiln does not know what the error is. If a screenshot was taken, check the screenshot";
```

BryceStevenWilley (commented):
We also need to make an issue to 1) reproduce and 2) take care of the special case of the `pre` tag (mentioned in above comment).

plocket (approved):
I think the requirement for approval was clarifying the error message, which looks done! Thanks!

I did want to note that I would only class something as a "full crash" if the whole test stopped. From what I understand, I think this is just a Scenario failure, which is correct just not clear. Maybe I'm misunderstanding the description, though.

PR #580, opened 2022-07-13
Makes `...there_is_another | false |` okay in tables

BryceStevenWilley:
`there_is_another | true` is invalid in a story table, because it will always be
true, and it will start an infinite loop of adding more elements to whatever list is
gathering.

However, you might be asked to add elements to a pre-filled list. In my case, I
add Individuals to the `users` and `other_parties` from existing EFM case
information, and then ask if the interviewee wants to add more users. Therefore
I can't use `target_number` in the story table, as the number won't include the
number of elements already added to the list (which Kiln can't figure out). I
can however, use `there_is_another = False`, which will terminate.

This does change how things are handled and documented, but IMO is a good
change. Let me know if we want a discussion here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,7 +38,12 @@ Format:
 ### Security
 - 
 -->
-<!-- ## [Unreleased] -->
+## [Unreleased]
+### Changed
+- Don't mark `...there_in_another | False |` in story tables as invalid, as it's necessary for some
+  workflows (see https://github.com/SuffolkLITLab/ALKiln/pull/580 for a longer discussion)
+  - Explicitly **not** documented, as we don't want to encourage people to use it if it's not
+    necessary for their interviews.
 
 ## [4.9.3] - 2022-09-07
 ### Fixed
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -2146,7 +2146,7 @@ module.exports = {
         enhanced_var_data.push( special_row );
 
       // Neutralize a `.there_is_another` row in a story table
-      } else if ( from_story_table && var_datum.var.match( /\.there_is_another$/ ) ) {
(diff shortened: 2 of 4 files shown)


plocket (commented):
This is permissive, but adds branching complexity that may not be necessary. Adding a test can help with that. If this is implemented, I'd rather leave it out of the documentation to avoid user confusion. That way, if it happens accidentally (e.g. with the Story Table generator), it's not a blocker, but it avoids people trying to explicitly use it.

plocket:
The changelog comment should mention that this should not be documented [and why it's not being documented].

plocket (approved):
Update the tests and the changelog (as described) and this should be good to go. Let me know if you want another review.

plocket:
Sorry to waffle on this, but regarding documentation we might want to have another conversation. I'm now rethinking. It might be ok to document with a hefty warning. Otherwise I'm concerned we'll forget this and be surprised by it later. Now that I look at it again, the case you made for the need is compelling.

BryceStevenWilley:
>  Otherwise I'm concerned we'll forget this and be surprised by it later.

I'm not super concerned about this, given

1) I don't think it's that surprising? Given that it'll do exactly what you expect it to do, and it's only surprising compared to our 2nd-order rules about using it.
2) For things like this, I'm more likely to check the code than I am to check the documentation, and I think as maintainers we should be doing that. 

Still okay with adding this in the documentation though. My concern isn't that it needs a warning (IMO it doesn't), but that it's adding complexity to our main suggested solution, which is "just use target_number".

plocket:
Fair enough. Let's just leave it out for now in that case.

BryceStevenWilley:
If I'm reading your last comment correctly, you agree that this doesn't need extra documentation outside of the CHANGELOG, and this is ready for merging, correct? There is a test added for it, and it's in the CHANGELOG now, so I think everything else is ready.

Issue #574, opened 2022-07-03
In the page's detected var names, dig into object variable names until no more nested objects are found


plocket:
In [`getPossibleVarNames`](https://github.com/SuffolkLITLab/ALKiln/blob/4be7351bf979aad2bfffb33995808de1ac101551/lib/scope.js#L604), we're only decoding dicts/objects down one level. Those objects may have objects of their own. Continue decoding objects in a while loop until no more objects are detected.

Can the parts of the objects get doubly encoded? If so, what does that look like? Maybe a doubly encoded obj would end up looking like this:
```js
foo[kwf23LER[B'EDckdleEsd]]
```
That first part (kwf23LER) would have to be decoded again separately?

#575 might be harder, but might be useful to make this work more reliably.

BryceStevenWilley:
Adding all of the info I've learned about how docassemble does these encodings.

On the "decoding" side (i.e. on the server when receiving a POST from proceeding in the interview), it uses these regexes to search for variables with brackets:
```python
match_brackets = re.compile(r'\[[BR]?\'[^\]]*\'\]$')
match_inside_and_outside_brackets = re.compile(r'(.*)(\[[BR]?\'[^\]]*\'\])$')
match_inside_brackets = re.compile(r'\[([BR]?)\'([^\]]*)\'\]')
```
This answers the questions about different parts being encoded separately: the answer is no, only the element inside the last set of brackets will be encoded. 

As for the question of that one part of the element being encoded more than once, I don't see any recursive encoding things on the docassemble side, meaning things are likely finite. The only example I can find is that on fields where the datatype is:
* `object`
* `object_radio`
* `object_multiselect`
* `object_checkboxes`

, there's an extra encoding level: e.g. the value for an object multiselect in the HTML `<option>` is `b2JqZWN0X211bHRpc2VsZWN0W0InWWpKS2NWZ3lPWGRrUmpoNidd`, which decodes to `object_multiselect[B'YjJKcVgyOXdkRjh6']`, and `YjJKcVgyOXdkRjh6` decodes to `b2JqX29wdF8z` which decodes to `obj_opt_3`, the actual name of the object. The docassemble code is really weird, but some smoking gun parts are the for loop at parse.py:5896 (commit d6fdcdacf), where the selection key is decoded an extra time.

plocket:
This sounds promising I think. A couple questions:

[...]

PR #581, opened 2022-07-13
Fix object radio

BryceStevenWilley:
The values of object radio buttons are variable names encoded as base64.
Convert those values into the human readable versions of themselves if they can
be converted.

* fromBase64 would accept invalid base64 strings, and just convert it to what it
  wanted to. This does a rough check to make sure that the value contains only
  valid base64 characters before trying to convert it. TODO: there could be more
  work done here to only add values that once decoded, end up as ascii
  characters (since we're working in python), but that can be future work.
* Added more docstrings to the toBase64 and fromBase64 functions
* Made unit tests pass by removing bad base64 conversions from the fixtures
* switched one of the test argument orders, since the expected and actual values
  are backwards. TODO: should change the other as well
  
Tests pass locally (and in the second commit), and don't pass even when the interview steps are kept, meaning 
that the fix does actually fix a DA related bug. The third commit (last one as of writing) fails because it includes extra steps not yet merged in ALAutomatedTestsingTests.

### DIFF
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -56,8 +56,8 @@ let is_secret_var = function(row) {
 
 let fromBase64 = function(base64_str) {
   /* Takes a base64 encoded string and returns it decoded.
-   * Note that this accept take strings with invalid base 64 characters,
-   * changes those characters to `/`s instead.
+   * Note that this can take strings with invalid base 64 characters,
+   * but will change those characters to `/`s instead.
    */
   let buffer = Buffer.from( base64_str, 'base64' );
   return buffer.toString( `utf-8` );
@@ -1238,7 +1238,7 @@ module.exports = {
   },  // Ends scope.getPossibleVarNames()
 
   nestedDecode: function ( scope, { to_handle, field_like_names = {}, nota } ) {
-    // A some fields have a key hidden inside them. (not yesno type checkboxes)
+    // Some fields have a key hidden inside them. (not yesno type checkboxes)
     let checkbox_match = to_handle.match( /^(.*)\[[BR]'(.*)'\]/ );
 
     // Of the format `foo[B'encoded']` or `foo[R'encoded']`
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -1075,21 +1075,17 @@ module.exports = {
 
       // button, radio, checkbox yesno, and maybe others need the value of the `value` attribute
       let values = [];
-      let multiple = false;
       if ( node.name === `select` ) {
         values = await scope.getOptionValues( scope, { $select: $node });
-        multiple_attr = await $node.attr('multiple');
-        field_data.multiple = typeof multiple_attr == 'string';
       }
       if ( values.length === 0 && node.attribs.value ) {
         // TODO: Add links to docs about da YAML values that devs never fill in themselves and
         // so don't know what they are. E.g. yesnomabye -> True/False/None.
         // TODO: Text inputs and textareas must ignore the value, etc. How?
         values.push( node.attribs.value );
-        field_data.multiple = false;
       }
 
-      field_data.guesses = await scope.getPossibleTableRowMatchers( scope, { var_names, values, multiple:field_data.multiple });
+      field_data.guesses = await scope.getPossibleTableRowMatchers( scope, { var_names, values });
       fields.push( field_data );
     }  // ends for every node
 
@@ -1141,7 +1137,7 @@ module.exports = {
     return selector;
   },  // Ends scope.getSelector()
 
-  getPossibleTableRowMatchers: async function ( scope, { var_names, values, multiple }) {
+  getPossibleTableRowMatchers: async function ( scope, { var_names, values }) {
     /* Return variable names and values combined in ways that can
     * try to match table rows. Tables can have two properties:
     * variable name and value. */
@@ -1150,9 +1146,7 @@ module.exports = {
       // Inputs like textareas don't have values
       if ( values.length > 0 ) {
         for ( let value of values ){
-          if (!multiple) {
-            possible_table_rows.push({ var: var_name, value: value, });
-          }
+          possible_table_rows.push({ var: var_name, value: value, });
           // Ignore blanks, and values that certainly aren't base64 encoded like True and False
           // Modified from https://stackoverflow.com/a/8571649/11416267
           // DA doesn't use `==` to pad base 64
@@ -1161,11 +1155,7 @@ module.exports = {
             let value_decoded = fromBase64( value);
             let values_decoded = scope.nestedDecode( scope, {to_handle: value_decoded });
             for (let possible_val of values_decoded) {
-              if (multiple) {
-                possible_table_rows.push({var: possible_val, value: 'True'})
-              } else {
-                possible_table_rows.push({var: var_name, value: possible_val, });
-              }
+              possible_table_rows.push({var: var_name, value: possible_val, });
             }
           }
         }
@@ -1270,18 +1260,14 @@ module.exports = {
         possible.push( `${ var_name }['None']` ); 
       } else {
         if (checkbox_match[2].match(da_base64_regex)) {
-          console.log(`${ checkbox_match[2] } matches da regex?`)
           // Get the choice name. E.g. option_1, option_2
           let box_to_handle = fromBase64( checkbox_match[2] );
-          console.log(`  pushed ${ box_to_handle }`)
           possible.push( `${ var_name }['${ box_to_handle }']` );
           if (box_to_handle.match(da_base64_regex)) {
-            console.log(`${ box_to_handle } matches da regex?`)
             try {
               // The part inside `[B'encoded']` could need one more decoding (or more?)
               // It may have invalid chars if it's not a one-word var name, though, like '_'
               let box_layer_2_decoded = fromBase64( box_to_handle );
-              console.log(`  pushed ${ box_layer_2_decoded }`)
               possible.push(`${ var_name }['${ box_layer_2_decoded }']`);
             } catch ( err ) {}  // It wasn't double encoded. Do nothing.
           }
@@ -1637,7 +1623,7 @@ module.exports = {
       await scope.setCustomDatatype[ custom_datatype ]( scope, { handle: custom_handle, answer });
     } else {
       // If no custom datatype, set a regular field
-      await scope.enter_answer[ field.tag ]( scope, { handle, answer, field });
+      await scope.enter_answer[ field.tag ]( scope, { handle, answer });
     }
 
   },  // Ends scope.funnel_the_answer()
@@ -1705,7 +1691,7 @@ module.exports = {
       await scope.setText( scope, { handle, answer });
       await scope.afterStep(scope);  // No showifs for this one?
     },
-    select: async function ( scope, { handle, answer, field=null }) {
+    select: async function ( scope, { handle, answer }) {
       // Note: I don't remember what's really going on in here.
 
       // A dropdown's option value can be one of two things
@@ -1720,19 +1706,7 @@ module.exports = {
       // There could be multiple fields on the page with the same option
       // values.
       // let option = await handle.$(`option[value="${ answer }"]`);
-      if ( session_vars.get_debug() ) {
-        console.log(`in select: answer: ${ answer }, handle: ${ handle }, field:`, field);
-      }
-      if (field && field.multiple) {
-        // We do the same as checkboxes for multiselect: the field name will be
-        // the option to use, and the answer will be "true" or "false"
-        console.log("in field && field.multiple");
-        if (answer === 'True') {
-          answer = field.matching_input_rows[ matching_input_rows.length - 1 ].var;
-          console.log(`new answer: ${ field }`);
-        }
-      }
-      let option = await handle.$(`option[value="${ answer }"]`);
+      let option = await scope.page.$(`option[value="${ answer }"]`);
       if ( option ) {
         await handle.select( answer );  // And use that value to set it
 
@@ -1740,9 +1714,6 @@ module.exports = {
       // TODO: Can these be double encoded? Can they be encoded like objects?
       } else {
         let base64_name = toBase64( { utf8_str: answer });
-        if ( session_vars.get_debug() ) {
-          console.log(`in select: base64'd: ${ base64_name }`);
-        }
         await handle.select( base64_name );
       }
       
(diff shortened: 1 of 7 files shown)


BryceStevenWilley:
TODO:
* [ ] make a test for checkboxes, specifically with a weird unicode string as a choice and make sure it's choosable
* [ ] make a new docassemble YAML file for the above test (the "decoding" test)

plocket:
Some of the fixtures are failing. A bit braindead to dig into why right now, sorry. Fixtures are sadly fragile. I hesitated to use them in the first place. It seemed like a shame not to unit test them when the output was a string, though. We might discuss removing them.

BryceStevenWilley:
I am aware and know why, see https://github.com/SuffolkLITLab/ALKiln/pull/581/commits/f81a49fb4cb4e56efcc36731da3f3ee5f8be44bb.  

> unit tests are broken because I added `multiple` as an attribute to the fields, so now the deep equals are wrong

Was also having issues with chai not actually printing the diff at all (it just says the assert failed and doesn't print anything else), which was pretty annoying. 

In general I don't have the bandwidth to work on this for a bit. This was supposed to be a smaller commit (fixing just object radio buttons), but has ballooned in scope to try to handle all of the base64 fields, as well as handling object multiselects and various other fields that I don't have the experience with.

BryceStevenWilley:
Since we mentioned (offline) that we should still include some of the changes from #604 and this PR in the stripped down PR, it's just easier to take out the code trying to handle multiples. That code is now in the `fix_object_multiselect` branch (https://github.com/SuffolkLITLab/ALKiln/pull/new/fix_object_multiselect).

I just discovered that I can't run puppeteer in WSL, so unfortunately I can't get things testing and running fully. I anticipate that i'll have to comment out all lines in `test_decoding.yml` that aren't `object_radio` for the tests to fully work, but I can confirm that the fixtures are fine now. Next week I'll have a better setup and will be able to test this fully and get to a point where it's mergable.

BryceStevenWilley:
I've got tests working on this, with the changes in https://github.com/plocket/docassemble-ALAutomatedTestingTests/pull/197.

Should be ready for a last review, @plocket

lib/scope.js
```diff
@@ -54,7 +54,33 @@ let is_secret_var = function(row) {
...
+let fromBase64 = function(base64_str) {
+  /* Takes a base64 encoded string and returns it decoded.
+   * Note that this accept take strings with invalid base 64 characters,
+   * changes those characters to `/`s instead.
+   */
+  let buffer = Buffer.from( base64_str, 'base64' );
+  return buffer.toString( `utf-8` );
+}
+
+let toBase64 = function ( { utf8_str }, exclude_padding=true) {
```
plocket: Nothing terribly wrong, but the function trigger of `toBase64()` is different than `fromBase64()` in a complex way. The "from" takes a string as an argument, the "to" takes an object with a string as an argument. It might be good to make those more similar for less confusion.

lib/scope.js
```diff
@@ -1200,96 +1234,65 @@ module.exports = {
       // TODO: Should we move this back into `getAllFields()`?
     }
 
-    // A lot of checkbox fields have a key hidden inside them. (not yesno type checkboxes)
-    let checkbox_match = layer_1_decoded.match( /^(.*)\[[BR]'(.*)'\]/ );
+    return scope.nestedDecode(scope, { to_handle: layer_1_decoded, field_like_names, nota });
+  },  // Ends scope.getPossibleVarNames()
+
+  nestedDecode: function ( scope, { to_handle, field_like_names = {}, nota } ) {
+    // A some fields have a key hidden inside them. (not yesno type checkboxes)
```
plocket: Not crucial
```suggestion
    // Some fields have a key hidden inside them. (not yesno type checkboxes)
```

plocket: More important than the others (to me) - I'd love an explanation of the different patterns of encoding that different configurations of  variable names can have. For example, how `foo[bar][bash.zoo][boo]` would be double encoded, but not triple encoded (or would it be triple and not quadruple?). I'd rather not lose the information that you so kindly dug up.

lib/scope.js
```diff
@@ -1200,96 +1234,65 @@ module.exports = {
...
+    let checkbox_match = to_handle.match( /^(.*)\[[BR]'(.*)'\]/ );
 
     // Of the format `foo[B'encoded']` or `foo[R'encoded']`
     // See https://github.com/jhpyle/docassemble/blob/595ad0a08572c0188f9ab49b1a6af99a3621cd42/docassemble_base/docassemble/base/standardformatter.py#L2108
 
-    // What about something like foo[R'encoded1'][B'encoded2']? Is that a
-    // thing? Can they be index numbers? Or just `foo['encoded1']`? What
-    // would their format be? The above? `foo[R'actual_choice'][B'encoded']`?
-    // `foo['something'][B'encoded']`?
+    // The B means it's a string, the R means it's 'something else', that was repr'd to a string.
```
plocket: Not totally crucial, but I think it would be helpful for future me: I think this is generally understandable by context, but not the specifics - `repr` is not a js thing from what I can tell. Is there a js way of saying this? Also, what's an example of 'something else'?

lib/scope.js
plocket: Sorry I'm nitpicking about comments. I just think this is a complex part of the concepts. Maybe the changes I'm suggesting wouldn't even be helpful and I'm just imagining they would. That said, next suggestion:

May be worth adding param descriptions too. You can probably copy a few of them from the function above. Even I'm struggling with it and I originally wrote it.

lib/scope.js
```diff
@@ -1200,96 +1234,65 @@ module.exports = {
...
   getOptionValues: async function ( scope, { $select }) {
     /* Returns all the option node values of the select node. */
     let options = [];
     let option_nodes = $select.find( `option` );
     for ( let opt of option_nodes ) {
       options.push( opt.attribs.value );
-      // Dropdown options created with objects will be encoded
-      let decoded = await scope.fromBase64( scope, { base64_str: opt.attribs.value });
-      options.push( decoded );
+      // Dropdowns created w/ objects are base64 encoded; but it's handled in getPossibleTableRowMatches
```
plocket, line 1304: Penny for your thoughts: Does it make sense to make an issue to change that so all the decoding goes on when we get the fields? Or maybe an issue to research that?

tests/features/interactive_steps.feature
```diff
@@ -132,3 +132,17 @@ Scenario: tap element with an error
   """
   Given I start the interview at "test_taps"
   And I tap the "Tests-not_there-tab" tab
+
+@i5 @decoding-test @object-multiselect
+Scenario: Runs through interview designed to use base64 encoding corner cases
```
plocket:
```suggestion
Scenario: Base64 encoded corner cases are decoded correctly
```

lib/scope.js
```diff
@@ -1200,96 +1234,65 @@ module.exports = {
...
-          // It may have invalid chars if it's not a one-word var name, though, like '_'
-          let box_layer_2_decoded = await scope.fromBase64( scope, { base64_str: box_layer_1_decoded });
-          var_names.push(`${ var_name }['${ box_layer_2_decoded }']`);
-        } catch ( err ) {}  // It wasn't double encoded. Do nothing.
+      if ( nota ) { 
+        possible.push( `${ var_name }['None']` ); 
+      } else {
+        if (checkbox_match[2].match(da_base64_regex)) {
+          // Get the choice name. E.g. option_1, option_2
+          let box_to_handle = fromBase64( checkbox_match[2] );
```
plocket: Can we change `box_to_handle` back to `box_layer_1_decoded`? I find the first option too vague and hard to understand what it's doing. The second one seems much more specific to what it's doing and matches the later `box_layer_2_decoded`. Or something that otherwise indicates that we're peeling back layers and this is just the first one? I know the second one may not happen in the end, but it's still the first later, even if it turns out to be the only one.

I do realize I made the terrible variable name `to_handle`, but I wish I hadn't done that.

lib/scope.js
```diff
@@ -54,7 +54,33 @@ let is_secret_var = function(row) {
   return original.flags !== undefined && original.flags.secret_name !== undefined;
 }
 
+let fromBase64 = function(base64_str) {
+  /* Takes a base64 encoded string and returns it decoded.
+   * Note that this accept take strings with invalid base 64 characters,
```
plocket: Very good stuff to know!

```suggestion
   * Note that this can take strings with invalid base 64 characters,
```

lib/scope.js
```diff
@@ -1200,96 +1234,65 @@ module.exports = {
       // TODO: Should we move this back into `getAllFields()`?
     }
 
-    // A lot of checkbox fields have a key hidden inside them. (not yesno type checkboxes)
-    let checkbox_match = layer_1_decoded.match( /^(.*)\[[BR]'(.*)'\]/ );
+    return scope.nestedDecode(scope, { to_handle: layer_1_decoded, field_like_names, nota });
+  },  // Ends scope.getPossibleVarNames()
+
+  nestedDecode: function ( scope, { to_handle, field_like_names = {}, nota } ) {
```
plocket: Not a huge thing, but is there a better name for this? If it's decoding nested values, maybe `decodeNested` instead? Some kind of action/verb instead of a noun?

plocket (approved):
The code looks good! Great changes! Some comments on clarity and some typo corrections, but this is probably good to go as far as functionality is concerned. There is one comment I feel is more important to add - the note about capturing da's encoding patterns.

This is a complex constellation of concepts to tackle. Kudos.

lib/scope.js
```diff
@@ -54,7 +54,33 @@ let is_secret_var = function(row) {
   return original.flags !== undefined && original.flags.secret_name !== undefined;
 }
 
+let fromBase64 = function(base64_str) {
+  /* Takes a base64 encoded string and returns it decoded.
+   * Note that this accept take strings with invalid base 64 characters,
+   * changes those characters to `/`s instead.
```
```suggestion
   * but will change those characters to `/`s instead.
```

lib/scope.js
BryceStevenWilley: I don't know of an equivalent in javascript, but I do want to stick with exact descriptions: docassemble is using the `repr` function in python. I don't think trying to describe what's happening would make sense in javascript, as it's very specific to the python objects internal to docassemble. I can make it clearer that it's talking about python process internal to docassemble though.

Something else is literally just anything that's not a string in docassemble. I'll make that clearer, but I don't have any specific examples. None of this is documented in docassemble, so it's hard to tie my understanding of docassemble internals to concrete examples.

lib/scope.js
BryceStevenWilley: Good catch, I had meant to simplify things by taking away the object wrapper.

lib/scope.js
BryceStevenWilley: Not really. `getPossibleTableRowMatches` is called only a few lines after `getOptions`, so they're very closely tied together anyway. 

If you want to, I could remove `getOptions`; it's only used in one place, and the one calling function, `getAllFields`, already has a lot of other implementation details for other types of fields, like checkboxes, so there's not a lot of benefit of abstracting `getOptions` out.

lib/scope.js
BryceStevenWilley: Agreed, `box_to_handle` is a bad name. It's not too hard to just change the name of `to_handle` to be better, so I'll do that.

The thing I don't quite like about the `layer_1_decoded` part is that this is now a separate function, so the layer numbers aren't actually correct. We've already decoded the whole string once outside of this function, so the numbers don't match my mental model that includes the whole process of decoding. Also, box is a vague term too.

I changed `to_handle` to `expr_str`, `box_to_handle` to `dictionary_key_decoded` (dictionary and not object because it's referring to the python dictionary that we're handling), and removed the need to have a named variable for the `layer_2` decoding, take another look and let me know if that addresses your confusion.

lib/scope.js
BryceStevenWilley: I landed on `decodeFully`, since it might or might not be nested, and "nested" has different meanings to me before vs after "decode".

lib/scope.js
BryceStevenWilley: I added param descriptions, and more comments in the function itself. I also removed some older, conflicting comments that weren't quite accurate with our new knowledge. Let me know if those help.

plocket: Some great updates! It doesn't quite cover the different variations in the ways that I wanted, but maybe that can be made into its own issue. e.g.:

Here are the different encoding patterns DA creates for variable names:
- `foo` -> "encoded str"
- `foo[bar]` -> "encoded str" -> "foo[B'encoded str]" -> "foo[bar]"
- `foo[bar][baz][zoo]` -> "encoded str" -> "foo[B'encoded str]" -> "foo[B'encoded str]" - > "foo[bar][baz][zoo]"
- So on for the different patterns you found

Here are the different encoding patterns DA creates for values:
- Etc. Though maybe these are the same.

I think it's worth being comprehensive here because the details really make a difference for how we've decided to do this. I would have expected to need a loop, but we don't, and this shows the justification for that.

Again, this can be an issue for the future, referencing this comment: https://github.com/SuffolkLITLab/ALKiln/issues/574#issuecomment-1205562550. We can maybe use the interview you created in order to use legit encodings.

lib/scope.js
plocket: Ah, B means the encoded value started as a string, R is that it started as a non-string, like a number or array or whatever.

lib/scope.js
plocket: I think we'll just leave this for another time. No need to do unrelated refactoring in this PR and not particularly crucial.

lib/scope.js
plocket: The change is better in some ways and I think we can leave it at that. What I appreciated about the previous variable names was that they expressed right there in the name (as opposed to a comment) that we were digging down into multiple decodings. We can wait for the future to tell us if it needs more clarity.

lib/scope.js
BryceStevenWilley: I think pushing it off is the best idea, I don't quite have the time to do the due diligence on all of the variations right now. Making an issue of your most recent comment.

lib/scope.js
BryceStevenWilley: I think it's more like an object (I guess everything in python is an object, but eh), like DAObjects and stuff, yeah.

BryceStevenWilley:
I think all of the comments were addressed and confirmed to be addressed or, (in the instance of #616) made into issues. Going to merge now!

PR #593, opened 2022-08-16
Close #589 Date for filename to y/m/d format plus ms at end

### LINKED ISSUE #589: Change date in filenames to year-month-day format
See https://stackoverflow.com/a/48729396/14144258 [and https://stackoverflow.com/a/6040556/14144258].
Change in https://github.com/SuffolkLITLab/ALKiln/blob/releases/v4/lib/utils/files.js

[Should we change the date format in the after-test report? Should we make a different issue for that? The main report could use the GitHub action environment variable to better match up with the GitHub test folder name. That wouldn't work for local tests, though. Ugh. How about the individual scenario reports? Maybe not necessary.]

plocket:
This will do a better job keeping files/folders in the right order by name. [I'm open to opinions on that, though.] There's a second note in the issue about the timestamp for the main report, but that should probably be its own issue/discussion.

Close #589.

### DIFF
--- a/lib/utils/files.js
+++ b/lib/utils/files.js
@@ -18,7 +18,7 @@ files.readable_date = function() {
   let secs = ("0" + date.getUTCSeconds()).slice(-2);
   let msecs = ("00" + date.getUTCMilliseconds()).slice(-3);
 
-  let fancy_str = `${ year }-${ month }-${ day } at ${ hours }hrs ${ mins }mins ${ secs }secs ${ msecs }ms UTC`;
+  let fancy_str = `${ year }-${ month }-${ day } at ${ hours }hr ${ mins }min ${ secs }sec ${ msecs }ms UTC`;
   return fancy_str;
 };  // Ends files.readable_date()
 
--- a/lib/utils/files.js
+++ b/lib/utils/files.js
@@ -12,11 +12,11 @@ files.readable_date = function() {
   let date = new Date();  // now
   let day = ("0" + date.getUTCDate()).slice(-2);
   let month = ("0" + date.getUTCMonth() + 1).slice(-2);
-  let year = date.getUTCYear();
+  let year = date.getUTCFullYear();
   let hours = ("0" + date.getUTCHours()).slice(-2);
-  let min = ("0" + date.getUTCMinutes()).slice(-2);
+  let mins = ("0" + date.getUTCMinutes()).slice(-2);
   let secs = ("0" + date.getUTCSeconds()).slice(-2);
-  let msecs = ("00" + date.getUTCgetUTCMilliseconds()).slice(-3);
+  let msecs = ("00" + date.getUTCMilliseconds()).slice(-3);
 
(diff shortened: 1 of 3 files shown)


lib/utils/files.js
```diff
@@ -5,11 +5,20 @@ let files = {};
...
-  let fancy_str = str.replace(/\//g, '-').replace(',', ' at').replace(':', 'hrs ').replace(':', 'mins ') + 'secs UTC';
+  let day = ("0" + date.getUTCDate()).slice(-2);
+  let month = ("0" + date.getUTCMonth() + 1).slice(-2);
+  let year = date.getUTCFullYear();
+  let hours = ("0" + date.getUTCHours()).slice(-2);
+  let mins = ("0" + date.getUTCMinutes()).slice(-2);
+  let secs = ("0" + date.getUTCSeconds()).slice(-2);
+  let msecs = ("00" + date.getUTCMilliseconds()).slice(-3);
+
+  let fancy_str = `${ year }-${ month }-${ day } at ${ hours }hrs ${ mins }mins ${ secs }secs ${ msecs }ms UTC`;
```
BryceStevenWilley: Don't remember if I've formally brought this up so I will now; IMO reading `10hrs 05mins 03secs 104ms` is harder than reading `10:05:03.104`; the latter is just how people generally write time, and notably the former makes the file names an extra 14 characters, which does matter when you're trying to read all of the filename at once.

And then, it's almost exactly the same as one of the answers you linked, and you could change the code to
```javascript
let fancy_str = new Date().toISOString().substr(0, 19).replace('T', ' at ') + ' UTC';
```

I would very strongly support this change, but I don't want to block the already great improvement in this PR for it.

BryceStevenWilley (approved):
If we're changing the datetime string, I think we could make it even better, but I won't block it.

lib/utils/files.js
plocket: I think my concern with `:` was that it was an invalid dir/filename char in some OSs. Maybe see https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#naming-conventions for invalid chars.

Good find on `.toISOString()`. I kept trying to find a format that would follow the rules we wanted and couldn't hunt it down for some reason. I could simplify the whole date part of this using that.

plocket: I asked around on how to shorten the name in other ways, but some people were confused when I used the format of `xxh xxm xxs`. [I do wish we could find a shorter format, though.]

BryceStevenWilley: ugh, windows. Forgot about that. I'll resolve this then.

My preferences on the format is `10:43:55.123` and then `10hrs 43mins 55secs 123ms`, I agree `10h 43m 55s` is both too terse and not standard.

plocket: Quick question - did you deliberately remove 's' from the end of 'min' and 'sec' indicating to change it, or was that just a typo?

BryceStevenWilley: Not intentional, but would be fine with either `10hrs 43mins 55secs 123ms` or `10hr 43min 55sec 123ms`

PR #602, opened 2022-08-31
Close #600. Allow devs to test list of interviews on their server.

### LINKED ISSUE #600: Adjust artifact folder and interview loading to allow devs to test list of interviews from the server
Related to, but separate from #599, as this is needed even if we're not handling the rest of the mess in this repository.

Two things need to be adjusted:
- [x] Testing the list of interviews from the server should avoid loading any package onto the server, so we'll need to create artifact folder even when setup.js isn't run. Artifact folder creation needs to be moved to `BeforeAll()` in steps.js.
- [x] The interview load step needs to be able to go to an arbitrary address. The list of interviews have a different kind of link to access them and it doesn't match our current pattern.

Edit:
- [ ] The runtime_config.json needs to be created even when only the artifact folder is created. It is currently created only when saving the project name.

plocket:
Closes #600.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,7 +38,9 @@ Format:
 ### Security
 - 
 -->
-<!-- ## [Unreleased] -->
+## [Unreleased]
+### Changed
+- Allow devs to test list of interviews on their server. See https://github.com/SuffolkLITLab/ALKiln/issues/600.
 
 ## [4.8.0] - 2022-08-30
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,7 +40,8 @@ Format:
 -->
 ## [Unreleased]
 ### Changed
-- Allow devs to test list of interviews on their server. See https://github.com/SuffolkLITLab/ALKiln/issues/600.
+- Allow devs to test an arbitrary interview. See https://github.com/SuffolkLITLab/ALKiln/issues/600.
+- Create artifacts folder even when setup has been skipped. See https://github.com/SuffolkLITLab/ALKiln/issues/600.
 
 ## [4.8.0] - 2022-08-30
 ### Changed
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -759,7 +759,7 @@ module.exports = {
 
     let interview_url = null;
     // If the file is already a full url, just use that arbitrary url
-    if ( file_name.includes( `http` ) ) {
+    if ( /^http/.test( file_name ) ) {
       interview_url = file_name;
 
     // If the filename is not a full url
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -752,6 +752,11 @@ module.exports = {
     /* Try to load the page. Should we also pass a `timeout` arg or just
     *    use scope.timeout? */
 
+    await scope.addToReport(scope,{
+      type: `row`,
+      value: `Trying to load the interview question page for "${ file_name }"`
+    });
+
     let interview_url = null;
     // If the file is already a full url, just use that arbitrary url
     if ( file_name.includes( `http` ) ) {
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -38,7 +38,9 @@ Format:
 ### Security
 - 
 -->
-<!-- ## [Unreleased] -->
+## [Unreleased]
+### Changed
+- Allow devs to test list of interviews on their server. See https://github.com/SuffolkLITLab/ALKiln/issues/600.
```
BryceStevenWilley: I would say that this lets you test installed interviews, right? Nothing in this PR is specific to getting the full list and testing it.

lib/scope.js
```diff
@@ -752,10 +752,27 @@ module.exports = {
     /* Try to load the page. Should we also pass a `timeout` arg or just
     *    use scope.timeout? */
 
-    if ( !file_name.includes('.yml') ) { file_name = `${ file_name }.yml` }
-    let base_url = await get_base_interview_url();
-    let interview_url = `${ base_url }${ file_name }`;
-    if ( session_vars.get_debug() ) { console.log( interview_url ); }
+    let interview_url = null;
+    // If the file is already a full url, just use that arbitrary url
+    if ( file_name.includes( `http` ) ) {
```
BryceStevenWilley: I would make this more specific. Not sure if JS has a startswith, but something like `file_name.startswith(`http://`) || file_name.startswith(`https://`). Just so we don't have any false positives.

lib/scope.js
```diff
@@ -794,35 +813,55 @@ module.exports = {
...
 
+    // Wait for at least one of these to appear
     let winner = await Promise.race([
       // Signature page
       scope.page.waitForSelector( `#dasigpage`, { timeout: scope.timeout }),
       // Any other valid page
       scope.page.waitForSelector( `#daMainQuestion`, { timeout: scope.timeout }),
       scope.page.waitForSelector( `#da-retry`, { timeout: scope.timeout }),
-    ]);
+    ]).catch( function ( error ) {
```
BryceStevenWilley, line 830: probably out of scope for this PR, but for my own curiosity:

Can we tell when the page is completely finished loading? Would mean that we wouldn't have to wait for the timeout when everything is done and sitting there.

BryceStevenWilley (approved):
LGTM! I think it should be described differently in the CHANGELOG, but the changes themselves look great

lib/scope.js
plocket: We already do that by checking for any of those elements to be present. As soon as one is there, the race stops and the code moves on. The timeout is just in case none of them are found. We don't want to wait for them for forever. Of course, we may be missing some kind of indicator of a page having loaded, but I'm not sure how to track that down.

plocket:
[I decided to instead take the possibly secret url out of the report. The arbitrary url may have the server address, which is possibly given as a secret and thus would be expected to remain secret. The only reason it's a secret is that there's no other way to pass in an environment variable on GitHub. That said, there'd be no way to differentiate between a secret env var and a non-secret env var anyway.]

~Documentation: arbitrary url won't keep secrets secret~

~Arbitrary urls won't be secret in the report - they'll be printed as is. We can't detect what's secret in there.~

~Make this into an issue on the documentation site.~

lib/scope.js
BryceStevenWilley: > The timeout is just in case none of them are found. We don't want to wait for them for forever. 

That was my question, if none of them appear, we just wait for the timeout, which is a little silly IMO, especially for tests with longer timeouts.

This stackoverflow has some code that we should consider using: https://stackoverflow.com/questions/52497252/puppeteer-wait-until-page-is-completely-loaded.

plocket:
I decided to instead take the possibly secret url out of the report. The arbitrary url may have the server address, which is possibly given as a secret and thus would be expected to remain secret [especially when it comes from an organization's env vars]. The only reason it's a secret is that there's no other way to pass in an environment variable on GitHub. That said, there'd be no way to differentiate between a secret env var and a non-secret env var anyway.

lib/scope.js
plocket: Always good to have a new eye on things! When I looked, I only remember finding stuff about `networidle` being unreliable. This looks like an interesting solution. We should back it up with the timeout as well, though, just in case it keeps changing for some reason. See [the comment below the answer](https://stackoverflow.com/questions/52497252/puppeteer-wait-until-page-is-completely-loaded#comment118988927_61304202) for an example.

[Do we already have an issue for this? ~I think I added one.~ That was a different issue about a different aspect of this.]

PR #604, opened 2022-09-02
Adjust where `multiples` appears in fields.fixtures

plocket:
Took a crack at this myself and just want to check if the `fields.fixtures.js` makes sense for the code you wrote @BryceStevenWilley. Not sure if those adjustments are accurate to what expected behavior is.  If you don't have bandwidth for this, that's ok too.


BryceStevenWilley:
I manually integrated these changes into #581, so I don't think we need this PR anymore.

Issue #613, opened 2022-09-07
Speed up the failure of tapable elements to appear on the page


plocket:
> The timeout is just in case none of [the tapable elements] are found. We don't want to wait for them for forever. 

That was my question, if none of them appear, we just wait for the timeout, which is a little silly IMO, especially for tests with longer timeouts.

This stackoverflow has some code that we should consider using: https://stackoverflow.com/questions/52497252/puppeteer-wait-until-page-is-completely-loaded.

_Originally posted by @BryceStevenWilley in https://github.com/SuffolkLITLab/ALKiln/pull/602#discussion_r959993336_

Reply:

> Always good to have a new eye on things! When I looked, I only remember finding stuff about `networidle` being unreliable. This looks like an interesting solution. We should back it up with the timeout as well, though, just in case it keeps changing for some reason. See [the comment below the answer](https://stackoverflow.com/questions/52497252/puppeteer-wait-until-page-is-completely-loaded#comment118988927_61304202) for an example.

BryceStevenWilley:
A partial implementation of this is when searching for continue buttons, as in https://github.com/SuffolkLITLab/ALKiln/pull/620. We know when some of the continue buttons are present, they all should be. So if we iterate through them and don't see a continue button, we can stop immediately and say so.

PR #620, opened 2022-09-29
Only "continue" on btn-primary buttons

BryceStevenWilley:
Not btn-warning or btn-danger, which are the default selections for Restart and Exit buttons on kickout screens. Without this, it's easy for a test with a bad `target id` to get into an infinite loop, getting sent to the kickout screen, restarting back to the beginning, and starting again.

All `next` / continue buttons on have the `btn-primary` class. It's not bullet proof, but it's okay? The only difference I could tell is that the `next` buttons don't have an associated `value` attribute, and `restart` does, though I think that's only for a list of buttons. We can't really tell what the buttons will do from those values, but we do know that they aren't continue buttons.

Needs https://github.com/plocket/docassemble-ALAutomatedTestingTests/pull/198 to be successful.

### DIFF
--- a/tests/features/interactive_steps.feature
+++ b/tests/features/interactive_steps.feature
@@ -146,3 +146,18 @@ Scenario: Base64 encoded corner cases are decoded correctly
     | object_checkboxes_test['obj_chkbx_opt_1'] | True | |
     | object_dropdown | obj_opt_2 | |
     | checkbox_dict['single_quote_key']['🁳🧴🯍🥫🅍🗎🹇🗷🌲🯵'] | True | |
+
+@fast @i6
+Scenario: Doesn't try to 'continue' with a restart button
+  Given the final Scenario status should be "failed"
+  And the Scenario report should include:
+  """
+  The test got stuck on "kickout-screen"
+  """
+  Given the max seconds for each step in this scenario is 10
+  Given I start the interview at "test_kickout"
+  And I get to "doesnt exist" with this data:
+    | var | value | trigger |
+    | user_choice | wrong | |
+  
+
--- a/tests/features/interactive_steps.feature
+++ b/tests/features/interactive_steps.feature
@@ -147,8 +147,11 @@ Scenario: Base64 encoded corner cases are decoded correctly
     | object_dropdown | obj_opt_2 | |
     | checkbox_dict['single_quote_key']['🁳🧴🯍🥫🅍🗎🹇🗷🌲🯵'] | True | |
 
+# This test should fail because it's target id is never found, but that also means
+# it has succeeded because it didn't try to keep looking for the id by pressing 'restart'
+# to continue.
 @fast @i6
-Scenario: Doesn't try to 'continue' with a restart button
+Scenario: Fails as it doesn't try to 'continue' with a restart button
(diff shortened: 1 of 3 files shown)


tests/features/interactive_steps.feature
```diff
@@ -145,4 +145,19 @@ Scenario: Base64 encoded corner cases are decoded correctly
     | object_radio_field | object_radio_choice_one | |
     | object_checkboxes_test['obj_chkbx_opt_1'] | True | |
     | object_dropdown | obj_opt_2 | |
-    | checkbox_dict['single_quote_key']['🁳🧴🯍🥫🅍🗎🹇🗷🌲🯵'] | True | |
\ No newline at end of file
+    | checkbox_dict['single_quote_key']['🁳🧴🯍🥫🅍🗎🹇🗷🌲🯵'] | True | |
+
+@fast @i6
+Scenario: Doesn't try to 'continue' with a restart button
```
plocket: A failing scenario should start with "Fails with". Also, an explanation as to why this both fails and succeeds might be useful as it's a bit of a weird case.
```suggestion
# This test should fail because it's target id is never found, but that also means
# it has succeeded because it didn't try to keep looking for the id by pressing 'restart'
# to continue.
@fast @i6
Scenario: Fails as it doesn't try to 'continue' with a restart button
```

plocket (approved):
Great improvement! Very excited for this case to be handled!

Items that I can see for improvement that I couldn't add as comments:
- Adding this to [`continue_exists()`](https://github.com/SuffolkLITLab/ALKiln/blob/7f7bf438fd058c06daddc5c03bd50464d8108e6d/lib/scope.js#L711) as well so such a test could finish faster under these circumstances. Not sure how to add a test for that since I think it's just used in the random input Step right now. Maybe if the YAML was a one-page test it would be more effective, but I'm not sure how a test can be shown to end correctly.
- Updating the CHANGELOG

Maybe we can make an issue for figuring out how to give this a better failure message than the timeout message? I hesitate to alter the default timeout message to be more vague. No elegant solution springs to mind at the moment. Maybe using `continue_exists()` in `setFields()` would be one way.

Regardless, I think this serves the purpose it describes. Let me know if you want me to review again if you change something you want a second eye on.

tests/features/interactive_steps.feature
BryceStevenWilley: TBH the main reason the tests is there is to avoid an infinite loop, not to particularly test that the failure occurs, which IMO makes it different from the other failure tests, i.e. if there was a way for the test to succeed instead, I would do that. But I guess it's better to match the convention, and just explain in the comments as you did.

BryceStevenWilley:
Added to the change log, and started using the selector in `continue_exists()` by sharing the same selector between the two functions. I'll wait for the new commit to pass tests, and then I'll merge.

> Maybe we can make an issue for figuring out how to give this a better failure message than the timeout message?

I think that's covered by https://github.com/SuffolkLITLab/ALKiln/issues/613, which if implemented would make timeout and not finding the correct interaction on a screen would be different error messages.

plocket:
~Tests passed and ready to merge?~ Needed to refresh the page.

PR #621, opened 2022-09-30
Add `1` to the month int, not the month str

BryceStevenWilley:
Adds parenthesis to correct the order of operations of JS addition and type coersion. Before, the timestamp string would be `81` for September, and now, it's properly `09`.

I don't think this needs tests.


plocket (approved):
Great fix, thanks!

Issue #614, opened 2022-09-09
Security: Freeze both our and the dev's package with package-lock.json


plocket:
I thought we had this issue, but I can't find it. This will ensure fullest security so that the updating/lack of security of upstream packages doesn't affect our packages except when we deliberately update package-lock.json.

For our dev's repo we need to:
- create a file to store their package-lock.json
- use/copy that file to create the dev's package-lock.json in the action (instead of the package.json)

For us and for the dev: configure our repo and their repo (through our action) to install using the package-lock.json.

plocket:
For copying a file from our own repository, possibly see `action_path` at https://docs.github.com/en/actions/learn-github-actions/contexts.

plocket:
This little test in our action works in a dependent repo:
```yml
runs:
  using: "composite"
  steps:
    - name: make file
      run: |
        touch cucumber.js
        cat "${{ github.action_path }}/cucumber.js" > cucumber.js
        file=$(<"cucumber.js")
        echo $file
      shell: bash
```

Note: tmate can't access values of variables like `github.whatever` in your local terminal. E.g. `github.action_path`. It gets the error 'bad substitution'. Instead, you have to (in your action.yml) add an environment variable to contain that value and use it later. Something like:

```yml
    - run: echo "ACTION_PATH=${{ github.action_path }}" >> $GITHUB_ENV
      shell: bash
    - name: make file
      run: |
        touch cucumber.js
        cat "${ACTION_PATH}/cucumber.js" > cucumber.js
        file=$(<"cucumber.js")
        echo $file
      shell: bash
```

plocket:
How to use package-lock.json instead of package.json when doing `npm install`: https://blog.npmjs.org/post/171556855892/introducing-npm-ci-for-faster-more-reliable

plocket:
We can store the package-lock.json in its own folder internally. When we want to update, we make changes in that internal file and run the usual `npm install` from that internal folder to update the package-lock.json (I believe).

plocket:
Not sure how we didn't see this before - I'm not sure how/if we can freeze the developer's package-lock.json while still giving them the option of which npm version of ALKiln to install.

We currently allow the developer to choose their version of ALKiln to install with npm using their ALKILN_VERSION env var. This lets them use npm version `^4.0.0` or `4.3.1` and such:

```yml
      run: |
        cat << EOF > "package.json"
        {
          # scripts...
          "dependencies": {
            "colors": "1.4.0",
[...]

Issue #594, opened 2022-08-17
Make things Windows compatible


BryceStevenWilley:
@trancode has been making great progress using `env-cmd` to make it such that we can run our npm commands that need env variables defined on Windows (ongoing work in https://github.com/SuffolkLITLab/ALKiln/tree/env-cmd).

The current roadblock is that some of the npm scripts do `run_cmd() { ... \"$@\" ... }; run_cmd`, which creates a shell / bash function, and then calls it. This fails in windows, since it doesn't know what to do with `run_cmd`.

Our best bet past this is to write the simple bash commands that we are using in a single JS file, making it automatically cross platfrom. The max things that we do in the bash commands are move around arguments and read in the status from other commands (https://nodejs.dev/en/learn/accept-input-from-the-command-line-in-nodejs might be useful).

BryceStevenWilley:
Should be closed by https://github.com/SuffolkLITLab/ALKiln/pull/624, but I'm gonna double check and make sure it can work on windows.

plocket:
I believe this worked on windows, but re-open if it didn't

PR #623, opened 2022-10-04
Move internal tests from ALAutomatedTestingTests

BryceStevenWilley:
Moves all tests specifically related to testing ALKiln are in this package.

Something that I've mentioned doing a few times, but hadn't gotten around to trying. Didn't take that long, and,~as seen by the passing test below, it works!~

I couldn't seem to do it right in this commit, but I should be able to add the [github commit history](https://github.com/plocket/docassemble-ALAutomatedTestingTests/graphs/contributors) to the first commit, so all of the committers from ALAutomatedTestingTests get proper credit in this repo as well. Will have to wait til tomorrow though, I can't do it on my machine I have with me today.

* (main driving reason): makes implementing https://github.com/SuffolkLITLab/ALKiln/issues/614 and https://github.com/SuffolkLITLab/ALKiln/issues/594 easier, since it makes the logic between `package.json` and
  `action.yml` more similar (specifically the args passed to cucumber become the same, so we wouldn't have to make twice as many arguments for each of those, and we can remove `cucumber-external`)
* makes running tests faster, since installing the tests no longer has a python module that restarts the server.
  Proof of this: the setup step takes [13 seconds on this branch](https://github.com/SuffolkLITLab/ALKiln/actions/runs/3183081132/jobs/5189859713#step:8:1), but it takes [51 seconds on the latest commit to main](https://github.com/SuffolkLITLab/ALKiln/actions/runs/3159765846/jobs/5143401581#step:8:1). 
* simplifies the dual purpose of ALAutomatedTestingTests to just having user-facing interviews to run; no more extra testing interviews
* works with `dainstall`, zip install (removing node_modules), and over GitHub URL
   (through ALKiln setup/ docassemble-API, not sure if the web UI has different logic)


BryceStevenWilley:
Tests didn't seem to fully execute, fixing now.

BryceStevenWilley:
Hadn't known how we were searching through folders to find upload files, also simplifies that code a bit now. Tests are passing!

lib/scope.js
```diff
@@ -1805,34 +1805,35 @@ module.exports = {
...
-        fs.accessSync( dir_features );
-        dir = dir_features;
-      } catch ( error_features ) {
-        await scope.addToReport( scope, { type: `warning`, value: `There are two places where you can store files to upload: "${ dir_source }" and "${ dir_features }". These tests could not find either of those.` });
-        throw error_source;  // Most people will want to know the sources folder path;
+      fs.accessSync(docassemble_pkg_path);
+      // https://stackoverflow.com/a/24594123/11416267
+      let dirs = fs.readdirSync(docassemble_pkg_path, {withFileTypes: true}).filter(dirent => dirent.isDirectory()).map(dirent => dirent.name);
+      if (dirs.length < 1) {
+        let msg = `You can store files to upload in 'docassemble/*/data/source'. These tests could not find a directory like that.`;
```
plocket: Nitpick: Can we include the package name in `'docassemble/*/data/source'` instead of `*`? I think at this point we already have it. For less experienced developers, the `*` would be confusing.

lib/scope.js
```diff
@@ -1805,34 +1805,35 @@ module.exports = {
...
+        await scope.addToReport( scope, { type: `warning`, value: msg});
+        throw Error(msg);  // Most people will want to know the sources folder path;
+      }
+      if (dirs.length != 1) {
+        await scope.addToReport( scope, { type: `warning`, value: `You have multiple directories in 'docassemble' folder (${ dirs }). These tests will use the first directory (${dirs[0]}).` });
```
plocket: Nitpick: Instead of multiple directories, would it be appropriate to say multiple package folders or multiple project folders? I think new developers would be confused by this. I also think we might simplify language a bit by using 'folders' instead of 'directories' in various places. I've probably used 'directories' myself in other places and never noticed it, but I think 'folders' is more clear.

.github/workflows/test_tests.yml
```diff
@@ -28,9 +28,9 @@ jobs:
       PLAYGROUND_PASSWORD: ${{ secrets.PLAYGROUND_PASSWORD }}
       PLAYGROUND_ID: ${{ secrets.PLAYGROUND_ID }}
       DOCASSEMBLE_DEVELOPER_API_KEY: ${{ secrets.DOCASSEMBLE_DEVELOPER_API_KEY }}
-      BRANCH_PATH: main
+      BRANCH_PATH: move_tests_internal
```
plocket: I think we need to change this to be more like our action workflow so that the branch and repo address get created automatically based on the test. As soon as we delete this branch, these tests will fail. See https://github.com/SuffolkLITLab/ALKiln/blob/releases/v4/action.yml#L55-L56.

plocket (changes requested):
One main change needed for this to work and a couple nitpicks about error messages.

I'm ok to go forward with this change after the fix, though I'll say that moving our test package code in here has been debated on and off in the past. I think we decided to separate them mainly since it seemed inappropriate to have our npm install include downloading a docassemble package (I think, though maybe there were other reasons too), but there's also the argument that we should keep our testing files within our repo. Honestly, I've never been sure which way to jump.

.github/workflows/test_tests.yml
BryceStevenWilley: Thanks for pointing me to that, I changed this just to check that it would still work on Github, but hadn't started tackling this problem yet, but `action.yml` looks like it should solve it.

lib/scope.js
BryceStevenWilley: That's fine, I can change it to folders. I think that "package folders" would be more confusing though.

lib/scope.js
BryceStevenWilley: Do you mean from the repo name? It's an okay guess, but TBH if they don't have a `docassemble/*` directory, it's not a docassemble package (you can't put things in top level docassemble namespace), and there are other issues that we won't be able to help with? Essentially, this should be a very rarely seen error.

BryceStevenWilley (commented):
> I think we decided to separate them mainly since it seemed inappropriate to have our npm install include downloading a docassemble package

I agree it is a little awkward, but TBH, those tests are directly related to this npm package and nothing else. If we really don't want them to be published to NPM, we can make a [`.npmignore`](https://zellwk.com/blog/ignoring-files-from-npm-package/#blacklisting-files-with-npmignore), wich won't push those files to npm.

lib/scope.js
BryceStevenWilley: Added it, wasn't able to improve the error message much more in this case, but if they don't have a docassemble folder I do ask if they're in the right directory (printing the current working directory as well).

lib/scope.js
BryceStevenWilley: Changed to just "folders", don't think there's much benefit in trying to be more specific with "subpackage folder" or anything like that.

.github/workflows/test_tests.yml
BryceStevenWilley: Tests haven't quite finished yet, but looks like they work well! https://github.com/SuffolkLITLab/ALKiln/actions/runs/3321851857/jobs/5490068479#step:5:21 gets the right repo URL and the step above prints the right branch.

BryceStevenWilley:
Address the comments, let me know if you want to chat about this at all. My main reasons for suggesting it are the already noticeable speedup in internal test time, and the potential to simplify a lot of the surrounding code we have.

lib/scope.js
```diff
@@ -1805,35 +1805,53 @@ module.exports = {
...
+      fs.accessSync(docassemble_pkg_path);
+      // https://stackoverflow.com/a/24594123/11416267
+      let dirs = fs.readdirSync(docassemble_pkg_path, {withFileTypes: true}).filter(dirent => dirent.isDirectory()).map(dirent => dirent.name);
+      if (dirs.length < 1) {
+        let msg = `You can store files to upload in 'docassemble/${ subpackage_name }/data/source'. These tests could not find a folder like that starting at ${ process.cwd() }.`;
+        await scope.addToReport( scope, { type: `warning`, value: msg});
+        throw Error(msg);  // Most people will want to know the sources folder path;
+      }
+      for (let dir of dirs) {
+        pkg_paths.push(`${docassemble_pkg_path}/${dir}`);
```
BryceStevenWilley, line 1824: I added the ability to actually check which of the folders in `docassemble` have a `data/sources` before just blindly choosing the first one. I ran into the local issue where I had a `.mypy_cache` folder under `docassemble` as well, which would get incorrectly chosen. Minor for others, but it was very helpful for me when developing locally, and does make things a bit more resilient IMO.

plocket: Great! Meant to comment on that. That's a great improvement.

plocket (approved):
These are great. I don't know if I'll get to give this another look, so I'm going to approve. Give me some kind of grace period to look it over, but if I haven't said anything by then, go ahead and merge it anyway 👍

BryceStevenWilley:
I'll probably try to merge this one later this week, just FYI.

PR #624, opened 2022-10-13
Installable Kiln + Working on Windows!

### LINKED ISSUE #594: Make things Windows compatible

BryceStevenWilley:
Very messy, not ready for merge, but wanted to show progress.

[plocket edit: Closes #594, I think]


BryceStevenWilley:
TODO: change the action.yml to use this way as well

BryceStevenWilley:
Note: currently testing this branch on https://github.com/SuffolkLITLab/docassemble-ALAffidavitOfIndigency/tree/NO_MERGE_test_kiln, check there to see how tests have progressed / been running. Nothing more should have to change in that repo, to keep APIs the same.

package.json
```diff
@@ -1,12 +1,12 @@
...
   "name": "@suffolklitlab/alkiln",
-  "version": "4.10.1-pre-cli",
+  "version": "4.11.0",
   "description": "Integrated automated end-to-end testing with docassemble, puppeteer, and cucumber.",
   "main": "lib/index.js",
   "scripts": {
     "test": "npm run unit && npm run langs_setup && npm run cucumber -- $@",
     "cucumber": "./lib/run_cucumber.js",
-    "langs": "npm run langs_setup && ./node_modules/.bin/cucumber-js --require ./lib/index.js docassemble/*/data/sources/*_lang*.feature",
+    "langs": "npm run langs_setup && npm run cucumber",
```
plocket, line 9: Can we avoid adjusting this until we actually start messing with the language features? We haven't for a while, so I'm not even sure they work anymore. From what I can tell/recall this is supposed to run just language feature files and I don't think the above implements that limit.

package.json
```diff
@@ -1,12 +1,12 @@
 {
   "name": "@suffolklitlab/alkiln",
-  "version": "4.10.1-pre-cli",
+  "version": "4.11.0",
```
plocket: No need to change this now, but for the future - avoid adjusting version numbers in PRs. That should be adjusted when the actual publishing happens. That's because by the time this gets published, this version number might not be correct anymore depending on what else happens in the meantime.

package.json
BryceStevenWilley: As it was, it would break with this refactor, so I changed it to at least work. I was going to mention that when this gets out of draft. 

I know that it doesn't run just the language feature files; if we want to support that, I can change things, but I'm not going to leave outdated code that is known to be broken around.

plocket: Good point. In that case, maybe we remove it for now and revisit when we have time to dedicate to it? Maybe less confusing than leaving a remnant whose purpose is unclear?

BryceStevenWilley: I can remove it, but that breaks compatability, so it's a major version change.

IMO the sctript is still useful: it's running a strict superset of the language tests, so it's still running all of the tests that script would run if it was perfect. If your normal tests pass, then we're just doing duplicate work (and if the normal tests don't pass, the other scripts would fail too). 

I acknowledge that it's unclear when reading the code, I can add a comment to avoid that. And if this holds up the feature, I can fix it by changing the arguments to run_cucumber.js. Now is the time to do that if we want.

It feels like the core of the issue is what to do with langs as a feature. It might be worth a discussion before this is merged.

TL;DR, our options:
* leave as is, comment to add clarity when reading code, mark the fact it takes longer to run as a known regression in Changelog
* remove script, make major version change
* fix to be the same functionality as it was before
* consider what we want to do with the language feature going forward, choose one of the above with a roadmap in mind

plocket: Let's just make an issue to get langs to only run lang files. For now if it runs everything, it'll be longer, but at least it'll still run langs.

BryceStevenWilley: Made a quick issue for this: https://github.com/SuffolkLITLab/ALKiln/issues/653

action.yml
```diff
@@ -68,58 +68,31 @@ runs:
...
     # Install node
     - name: "ALKiln: Install node packages"
       uses: actions/setup-node@v1
       with:
         node-version: "17"
-    - run: npm install
+    # npm install -g "@suffolklitlab/alkiln@$ALKILN_VERSION"
+    - name: Install ALKiln directly from npm
+      run: |
+        npm install -g github:SuffolkLITLab/ALKiln#lighter_package_json
```
plocket: We need to update this before publishing.

lib/run_cucumber.js
```diff
@@ -0,0 +1,72 @@
...
+(async function main() {
+  let argv = process.argv;
+
+  let just_args = argv.slice(2);
+
+  const environment = { cwd: process.cwd() };
+  const provided_config = {
+    "publishQuiet": true,
+    "require": "./lib/index.js", 
+    "format": ["progress", "summary:cucumber-report.txt"],
```
plocket: Do we need these if they're in cucumber.js? Where is cucumber.js being used?

lib/run_cucumber.js
```diff
@@ -0,0 +1,72 @@
...
+    }
+    // If we get a list of seemingly no tags, assume they meant to use tags
+    else if (just_args.every((val) => !val.startsWith("@"))) {
+      provided_config.tags = just_args.map((val) => "@" + val).join(' or ');
+    }
+    // If it's a mix of tags and no tags, it's probably a boolean expression!
+    else {
+      provided_config.tags = just_args.join(' ');
+    }
+  }
```
plocket, line 37: Can we just pass the tags through as they are? Not sure we should deviate from the documented behavior of tags/try to manipulate after the fact.

lib/run_cucumber.js
```diff
@@ -0,0 +1,72 @@
...
+  var path = require('path');
+  var codeDir = path.dirname(require.main.filename);
+  const support = await cucumber_api.loadSupport({ support: {
+    requireModules: [`${ codeDir }/index.js`],
+    requirePaths: [ `${ codeDir }/index.js`],
+    importPaths: [],
+  }, sources: {paths: []}}, environment);
+
+  const { success } = await cucumber_api.runCucumber({...runConfiguration, support});
+  console.log("%o", success);
```
plocket: What does "%o" mean?

lib/steps.js
```diff
@@ -92,7 +92,7 @@ BeforeAll(async () => {
   // Make the folder to store all the files for this run of the tests
   scope.paths.artifacts = files.make_artifacts_folder();
   // Save its name in a file. `session_vars` can get it back later.
-  // This is used by at least clean_reports.js.
+  // This is used by at least run_cucumber.js.
```
plocket: Did I miss this? I don't see `save_artifacts_path_name` in run_cucumber.js. Do we need to put it somewhere? Where is the artifacts path name being saved right now? steps.js? Something in the setup?

lib/utils/log.js
```diff
@@ -52,7 +52,8 @@ log.debug = function ({ type='', pre='', data='', post='' }) {
 };
 
 log.add_to_debug_log = function(value) {
-  fs.appendFileSync(log.debug_log_file, value);
+  let artifacts_path = session_vars.get_artifacts_path_name();
+  fs.appendFileSync(`${ artifacts_path }/${ log.debug_log_file}`, value);
```
plocket: Hmm, have we tested this out? Was there a reason we weren't putting this stuff in that folder? Or was it just being added to that folder somewhere else? It's weird that things were functioning when that functionality was missing.

plocket (commented):
Looking good, but some questions remaining.

lib/run_cucumber.js
BryceStevenWilley: Cucumber.js isn't being used anymore, we should be able to remove

lib/run_cucumber.js
plocket: https://developer.mozilla.org/en-US/docs/Web/API/console#outputting_text_to_the_console

plocket: If we include, we'd add more details, but maybe we don't need it. Maybe more logs are better, though.

lib/run_cucumber.js
BryceStevenWilley: We can't pass tags as they are, as it wouldn't run as is. This tries to keep the behavior as or-ing things, which is what the cli currently does.

lib/run_cucumber.js
plocket: Can we somehow log the full command that is constructed through this script?

plocket: We can print the full run configuration and support. Let's add that [since now the full command won't be logged to the console automagically].

lib/steps.js
plocket:
```suggestion
  // The value we are saving is used by at least run_cucumber.js.
```

lib/utils/log.js
BryceStevenWilley: Decided to just remove it; it's not being used anywhere / is probably currently broken, and this was a guess at how to fix to, but didn't start using the function anywhere.

plocket: Issue #659 as it's currently not being used anywhere.

BryceStevenWilley:
Tested on the ALAffidavit repo a few different ways:

* [with no tag arguments](https://github.com/SuffolkLITLab/docassemble-ALAffidavitOfIndigency/actions/runs/4386606290/jobs/7680869428)
* [with one tag argument (`@aaoi1`)](https://github.com/SuffolkLITLab/docassemble-ALAffidavitOfIndigency/actions/runs/4386740691/jobs/7681178780)
* [with two tag arguments (`@alaffidavit and @aaoi1`)](https://github.com/SuffolkLITLab/docassemble-ALAffidavitOfIndigency/actions/runs/4386775622)
* [fancy new way of running with multiple tags (`aaoi1 aaoi2`)](https://github.com/SuffolkLITLab/docassemble-ALAffidavitOfIndigency/actions/runs/4386819125/), and we'll automatically turn that into `@aaoi1 or @aaoi2`

All of them ran the expected scenarios, based an the tag expressions passed in.

lib/run_cucumber.js
BryceStevenWilley: Tested this, and these all work through GitHub actions as well

lib/run_cucumber.js
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/env node
+
+const fs = require('fs');
+const log = require('./utils/log');
+const session_vars = require('./utils/session_vars');
+
+// import {loadConfiguration, loadSupport, runCucumber } from '@cucumber/cucumber/api';
+const cucumber_api = require('@cucumber/cucumber/api');
```
plocket:
```suggestion
const cucumber_api = require('@cucumber/cucumber/api');
```

plocket (approved):
Looking amazing! I think my last (cosmetic) question remaining is `alkiln-run` vs. `alkiln-test` (having represented the steps before as setup/test/takedown). It is pretty cosmetic, though, and might take a bunch of fiddly little fixes. I'm ok either way.

Thank you!

Issue #618, opened 2022-09-19
Make most output file and directory names shorter


BryceStevenWilley:
Windows (which most users will be downloading Github artifacts to if their test fails), has a default limit of full file paths of 260. Trying to extract specific files to view their contents will fail sometimes. We should more generally have shorter artifact names.

## Steps to shorten artifact names:

1. [x] Assess how long names are currently: guess an average amounts of chars for each item, e.g. scenario names, timestamps
2. [x] see by how many characters we would be over the 260 limit
3. [x] Choose from the below ideas of how to best shorten file names to get under 260 (not worth it if we are well above 260 with all changes):
* Time - env vars for timezones, allows use to remove UTC from the end (4 characters per instance, 8 total?)
* "16hrs 14mins 15secs" -> “16h14m15s” (10 characters per instance, 20 total)
* Shorten start of folder name: just alkiln instead of alkin_tests_output (12 characters)
* use timestamp instead of human-readable times in some places (which?) (27 characters, likely would only do that inside the folder, so only in one place) 
* shorten amount of scenario name that's included (context dependent on how much we want to shorten)

BryceStevenWilley:
Ran into this today, so here's some rough calculations for how longs names are in windows, taking the following steps:
* download the Zip from a completed test, like at https://github.com/SuffolkLITLab/ALKiln/suites/8611375456/artifacts/386283955 to my Downloads folder
* open the Zip file with file explorer, which lets you look at the files inside the zip without extracting all of them
  * try to open just one of the images in the zip, without extracting things first, OR
  * extract the whole thing to the default suggested folder path in Downloads.

The prefix of the path is `C:\Users\bwilley\Downloads\`. Most people's user names won't be longer than 25 chars or so, so that's **43 chars max** for the prefix max, **avg closer to 28**.

Then due to how windows unzips files, it's the main zip name + the very similar internal folder name: `alkiln_tests_output-2022-10-04 at 17hrs 25mins 58secs UTC\alkiln_tests_output-2022-91-04 at 17hr 26min 41sec 573ms UTC`. In this case it's **118 characters**, and I think that's what it will always be, no variation.
* 20 for the titles, 
* 20 for the dates, 
* 57 for the timestamps

Here's an example image that won't extract inside that dir: `error_on-direct-standard-fields-Fail with was uexepctedly not able to continue for invalid field input message-2022-91-04 at 17hr 31min 30sec 942ms UTC.jpg`,  which is **155 characters**.
[...]

PR #626, opened 2022-10-26
Shorten artifact names

BryceStevenWilley:
Avoid artifact names whose whole paths are longer than 260 characters,
which causes issues on Windows machines. See
https://github.com/SuffolkLITLab/ALKiln/issues/618 for more details and
analysis. Doesn't fix completely, but does make less likely.


plocket (approved):
Great change! Thanks! Add to the changelog? Whatever you want.

PR #632, opened 2022-12-12
Use Axios's `toJSON` method to make shorter errors

### LINKED ISSUE #605: Remove axios full error log for setup (e.g. when Project is deleted or created), like with a 504
Do print the error message/status, though.

[Maybe duplicate of or addition to #550.]

BryceStevenWilley:
On an example erroring docassemble server (for example, this server returning a [502 to a delete project](https://github.com/SuffolkLITLab/docassemble-EFSPIntegration/actions/runs/3641406524/jobs/6147303425)), the entire axios error will print, which is that case in 1.8k lines, most of it being garbage.

Axios has a `toJSON()` method on errors that should reduce length of the errors being shown, as it will not print the most annoying parts of the long errors, like the internal axios socket objects

The errors look like this now, which IMO is an improvement:

```bash
$ alkiln-takedown
ALKiln takedown: Trying to remove the docassemble Project from the testing account.
ALKiln : Trying to delete Project fake_project

ALKiln  ERROR: Ran into an error when deleting Project "fake_project"

ALKiln takedown ERROR: could not delete_project
{
    "message": "Request failed with status code 400",
    "name": "Error",
    "stack": "Error: Request failed with status code 400\n    at createError (/home/brycew/Developer/LITLab/code/ALKiln/node_modules/axios/lib/core/createError.js:16:15)\n    at settle (/home/brycew/Developer/LITLab/code/ALKiln/node_modules/axios/lib/core/settle.js:17:12)\n    at IncomingMessage.handleStreamEnd (/home/brycew/Developer/LITLab/code/ALKiln/node_modules/axios/lib/adapters/http.js:293:11)\n    at IncomingMessage.emit (node:events:377:35)\n    at endReadableNT (node:internal/streams/readable:1312:12)\n    at processTicksAndRejections (node:internal/process/task_queues:83:21)",
    "config": {
        "transitional": {
            "silentJSONParsing": true,
            "forcedJSONParsing": true,
            "clarifyTimeoutError": false
        },
        "transformRequest": [
            null
        ],
        "transformResponse": [
            null
        ],
        "timeout": 120000,
        "xsrfCookieName": "XSRF-TOKEN",
        "xsrfHeaderName": "X-XSRF-TOKEN",
        "maxContentLength": -1,
        "maxBodyLength": -1,
        "headers": {
            "Accept": "application/json, text/plain, */*",
            "User-Agent": "axios/0.24.0"
        },
        "method": "delete",
        "url": "http://192.168.4.47/api/playground/project",
        "params": {
            "key": "********************************",
            "project": "fake_project"
        }
    },
    "status": 400
}
```

Fix #605.


plocket (changes requested):
Good goal! We decided what we'd like to do is properly isolate the `.toJSON()` calls in the `docassemble_api_REST.js` file. That way, everything else can just throw the error without worrying about it. Only the functions in the REST file will use `.toJSON()`.

lib/docassemble/setup.js
```diff
@@ -20,7 +20,13 @@ const setup = async () => {
 
   // Create a project on the da server's account with a unique name then
   // store it locally for later deleting the project. 
-  let project_name = await da_i.loop_to_create_project();
+  let project_name = null;
+  try {
+    project_name = await da_i.loop_to_create_project();
+  } catch (error) {
+    log.error({type: `setup`, pre: `could not loop_to_create_project`, data: JSON.stringify(error, null, 4)})
+    process.exitCode = 1;
```
plocket, line 31: Maybe add a comment that we're avoiding the unhandled promise rejection error.

lib/docassemble/takedown.js
```diff
@@ -10,9 +10,15 @@ const takedown = async () => {
...
-  session_vars.delete_project_name();
-  log.info({ type: `takedown`, pre: `Success! Test takedown has finished successfully.` });
+  try {
+    await da_i.delete_project();
+    session_vars.delete_project_name();
+    log.info({ type: `takedown`, pre: `Success! Test takedown has finished successfully.` });
+    process.exitCode = 0;
+  } catch (error) {
+    log.error({type: `takedown`, pre: `could not delete_project`, data: JSON.stringify(error, null, 4)})
+    process.exitCode = 1;
```
plocket: Same here - comment why we're using process.exitCode.

Actually not sure why process exiting on success?

lib/docassemble/takedown.js
```diff
@@ -10,9 +10,15 @@ const takedown = async () => {
...
   *   For local development, clean up project name file. */
   log.info({ type: `takedown`, pre: `Trying to remove the docassemble Project from the testing account.` });
-  await da_i.delete_project();
-  session_vars.delete_project_name();
-  log.info({ type: `takedown`, pre: `Success! Test takedown has finished successfully.` });
+  try {
+    await da_i.delete_project();
+    session_vars.delete_project_name();
+    log.info({ type: `takedown`, pre: `Success! Test takedown has finished successfully.` });
+    process.exitCode = 0;
```
plocket: I'm still not quite sure what the ramifications would be downstream (whatever github or our action script is doing). It seems safe, but why not let it complete on its own?

Answer:
1. Setting exit code doesn't stop the process from running, just affects the code.
2. It's implicitly always 0.

Response:
It still seems a bit weird to do it here. If in future we actually handle unhandled promise rejection, we'll be removing these exit code things anyway, so we'll just have to then also remember to clean this up. This `process.exitCode` stuff is, from what I understand, a hack to avoid the useless error. This isn't a new practice for our code in general.

Are we trying to follow some kind of best practices?

Answer: We may end up using `process.exitCode` in the handler

Response: Possibly true, but we won't be using it in takedown.js, etc. When we get rid of line 20, would we get rid of line 17? To which we think the answer is 'yes'. So line 20 is just an intermediate step and we might not want to include line 17 as part of that intermediate step.

Answer: Just thought it added clarity in this specific spot.

Response: With that justification, I'm ok either way.

BryceStevenWilley:
```suggestion
```

Don't really need it.

plocket: I'd possibly want a comment on why we're using it here too because it's an unusual thing to see.

plocket (approved):
I made some notes, but I don't think they're blockers (iirc). I think this is a great change in general.

lib/docassemble/setup.js
```diff
@@ -20,7 +20,13 @@ const setup = async () => {
 
   // Create a project on the da server's account with a unique name then
   // store it locally for later deleting the project. 
-  let project_name = await da_i.loop_to_create_project();
+  let project_name = null;
+  try {
+    project_name = await da_i.loop_to_create_project();
+  } catch (error) {
+    log.error({type: `setup`, pre: `could not loop_to_create_project`, data: JSON.stringify(error, null, 4)})
```
BryceStevenWilley:
```suggestion
    log.error({type: `setup`, pre: `something happened when creating docassemble project`, data: JSON.stringify(error, null, 4)})
```

lib/docassemble/setup.js
```diff
@@ -20,7 +20,13 @@ const setup = async () => {
...
   // Create a project on the da server's account with a unique name then
   // store it locally for later deleting the project. 
-  let project_name = await da_i.loop_to_create_project();
+  let project_name = null;
+  try {
+    project_name = await da_i.loop_to_create_project();
+  } catch (error) {
+    log.error({type: `setup`, pre: `something happened when creating docassemble project`, data: JSON.stringify(error, null, 4)})
+    process.exitCode = 1;
+  }
```
BryceStevenWilley:
```suggestion
  // Explicit try-catch prevents "Unhandled promise rejection error"
  // avoids showing confusing node errors to users that they shouldn't have to fix
  try {
    project_name = await da_i.loop_to_create_project();
  } catch (error) {
    log.error({type: `setup`, pre: `something happened when creating docassemble project`, data: JSON.stringify(error, null, 4)})
    // we don't re-throw the exception, so make sure to exit with 1 to let Github Actions know the script failed
    process.exitCode = 1;
  }
```

lib/docassemble/setup.js
BryceStevenWilley: Should be addressed now!

lib/docassemble/takedown.js
```diff
@@ -18,6 +18,7 @@ const takedown = async () => {
     log.info({ type: `takedown`, pre: `Success! Test takedown has finished successfully.` });
   } catch (error) {
     log.error({type: `takedown`, pre: `could not delete project`, data: JSON.stringify(error, null, 4)})
+    // we don't re-throw the exception, so make sure to exit with 1 to let Github Actions know the script failed
```
plocket, line 21:
```suggestion
    // we don't re-throw the exception to avoid an "unhandled promise rejection", so make sure to exit with 1 to let Github Actions know the script failed
```

plocket: I'm trying out putting the note about the error right where the weird looking code is. You can take it or leave it 👍

BryceStevenWilley: Thanks, but I think I'll pass: there's more clarification at the[level of the try catch](https://github.com/SuffolkLITLab/ALKiln/pull/632/files#diff-a424932832259b654fe8ce9283c6c61a8c7cc72e07daccff8c0488974557af15R13) just a few lines higher. The try-catch is the specific thing that's stopping the unhandled promise rejection, so I'd prefer to keep that comment closer to it.

lib/docassemble/setup.js
```diff
@@ -20,7 +20,17 @@ const setup = async () => {
...
   // Create a project on the da server's account with a unique name then
   // store it locally for later deleting the project. 
-  let project_name = await da_i.loop_to_create_project();
+  let project_name = null;
+  // Explicit try-catch prevents "Unhandled promise rejection error"
+  // avoids showing confusing node errors to users that they shouldn't have to fix
+  try {
+    project_name = await da_i.loop_to_create_project();
+  } catch (error) {
+    log.error({type: `setup`, pre: `something happened when creating docassemble project`, data: JSON.stringify(error, null, 4)})
```
plocket:
```suggestion
    log.error({type: `setup`, pre: `an unexpected error occurred while trying to create a docassemble project in the user's account`, data: JSON.stringify(error, null, 4)})
```

I'm just trying to add something about an error or something about specifically not being able to create the project to be more specific than something happening. I imagine folks like Caroline being a bit more annoyed about the lack of specificity. I do wish we could add a message about what their next step might be. Even something like looking at the previous lines in the log, checking the user account, looking at the docs, or contacting us. Maybe that has to wait till we fill out the docs with more info about specific error messages.

BryceStevenWilley:
Needs a CHANGELOG update 😬

BryceStevenWilley:
Did a changelog update, should be ready to go!

PR #633, opened 2022-12-16
Randomly Answer Question Fixes

### LINKED ISSUE #628: Random input selecting "Back" button?
In some simple random input tests, I found the back button was being pressed quite a bit. For example, in [AppearanceEfiling](https://github.com/SuffolkLITLab/docassemble-AppearanceEfile), the random test got to the 4th question after the first 3 intro screens (landing screen, introduction screen, and terms of use) 2 times in 25 screens. It would keep pressing back instead of next, since the landing screens only had 2 options to choose from.

We should recognize a few specific buttons on the screen (back, the review sections, etc). and click them at a much lower rate than buttons that advance the interview state, like next, etc. Review sections would be nice to test randomly, but they're less likely to crash the interview, and more likely to just have business logic errors, which would be harder to test.

BryceStevenWilley:
Trying to get randomly answer questions working on a real interview, but it's a bit brittle:

*  would occasionally run into null errors. To fix, just added more null checks in [this commit](https://github.com/SuffolkLITLab/ALKiln/commit/1266101d202f4a11518d7d43deb6e1cbd95a6191)
* on the Affidavit of Indigency, the kickout screen has an exit button that takes you to courtformsonline. Unfortunately, Kiln will take you there, but not realize you've left the interview, and thinks it's hit an error page. So we now navigate back to the interview if we leave it somehow.
* Fix #628, only presses the back button 10% of the time, not 50% of the time. The latter made it very improbable that you'd make it very far into the interview.
* Addresses some issues with radio buttons: only the last one would ever be pressed, which always led you to a kickout screen for this interview. Improved it a little bit, but still favors the later buttons due to implementation.

Some existing bugs that are still around:
* each radio button is considered separately, heavily weighting the answers towards the latter radio buttons; the last button has a 50% chance of being clicked, and the 2nd to last has a 25% chance, etc. https://github.com/SuffolkLITLab/ALKiln/issues/637
* doesn't work well with list collect screens, it tries to answer every single list item that it sees on screen (~20 or 30 I think?) and times out. https://github.com/SuffolkLITLab/ALKiln/issues/638
*  still running into weird bugs on certain screens when Kiln thinks there's no content. Unclear what page it's actually trying to query (maybe a not-yet loaded page?) https://github.com/SuffolkLITLab/ALKiln/issues/809

Even with those bugs, this PR should continue, definitely improves the situation a bit, though I wouldn't routinely run it for an interview yet, as there are still a lot of false "found an error screen"s.

### DIFF
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -1246,7 +1246,7 @@ module.exports = {
             + `column, but it is empty. The test might get unexpected results. See `
             + `https://suffolklitlab.org/docassemble-AssemblyLine-documentation/docs/automated_integrated_testing/#trigger.`
             + `\nThe row's data:\n${ row_as_story_table }`
-            + `\nThis field's data:\n${ JSON.stringify( field )}`;
+            + `\nThis field's data:\n${ JSON.stringify( field.guesses )}`;
           reports.addToReport(scope, { type: `warning`, value: row_trigger_warning });
           rows_without_a_trigger_name.push( row_that_matches_this_field );
         } else {
@@ -1401,8 +1401,7 @@ module.exports = {
       let elem = document.querySelector( selector );
       // If the element has a sibling that's a label, we need to
       // interact with that label instead.
-      // `elem` has never been null from what we've seen
-      if ( elem.nextSibling && elem.nextSibling.tag === `label` ) {
+      if ( elem && elem.nextSibling && elem.nextSibling.tag === `label` ) {
         return elem.nextSibling;
       } else { return elem; }
     // Pass in the selector as an argument
@@ -1422,14 +1421,15 @@ module.exports = {
       // future we should look into what happens with a `fieldset` for
       // `choices:`, etc.
       let custom_datatype = null, matches = null;
-      let group = elem.closest( `.da-form-group` );
-      if ( group && group.className ) {
-        matches = ` ${ group.className } `.match(/da-field-container-datatype-([^ ]+) /);
+      if (elem) {
+        let group = elem.closest( `.da-form-group` );
+        if ( group && group.className ) {
+          matches = ` ${ group.className } `.match(/da-field-container-datatype-([^ ]+) /);
+        }
+        if ( matches ) {
+          custom_datatype = matches[1];
+        }
       }
-      if ( matches ) {
-        custom_datatype = matches[1];
-      }
-
       return custom_datatype;
     });
 
@@ -1585,7 +1585,13 @@ module.exports = {
       /* Set value of some `input` element to the given value. */
       let row_was_used = true;
 
-      let type = await handle.evaluate( elem => { return elem.getAttribute('type'); });
+      let type = await handle.evaluate( elem => { 
+        if (elem) {
+          return elem.getAttribute('type'); 
+        } else {
+          return '';
+        }
+      });
       if ( type === `radio` || type === `checkbox` ) {
 
         let [label] = await handle.$x(`following-sibling::label`);
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -2762,11 +2762,15 @@ module.exports = {
         return true;
       });
       let field = null;
-      if (back_btn && Math.random() > 0.9) {
+      if (back_btn != null && Math.random() > 0.9) {
         // 10% chance of clicking the back button
         field = back_btn;
       } else {
         field = faker.helpers.arrayElement( fields );
+      }
+      if (field == null) {
+        // Just skip it? Shouldn't happen but has before
+        return;
       }
       let handle = await scope.get_handle_from_field( scope, { field: field });
       // Send it to be tapped
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -1602,8 +1602,13 @@ module.exports = {
           let name = await label.evaluate( elem => { return elem.getAttribute('for'); });
         }
 
-        if ( type === `radio` ) { await label.evaluate( elem => { return elem.click(); }); }
-        else { await scope.setCheckbox( scope, { label, answer }); }
+        if ( type === `radio` ) {
+          if (answer.toLowerCase() == "true") {
+            await label.evaluate( elem => { return elem.click(); });
+          }
+        } else {
+          await scope.setCheckbox( scope, { label, answer });
+        }
 
         await scope.afterStep(scope, { waitForShowIf: true });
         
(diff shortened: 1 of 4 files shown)


lib/scope.js
```diff
@@ -1742,7 +1742,13 @@ module.exports = {
     input: async function ( scope, { handle, answer }) {
       /* Set value of some `input` element to the given value. */
 
-      let type = await handle.evaluate( elem => { return elem.getAttribute('type'); });
+      let type = await handle.evaluate( elem => { 
+        if (elem) {
+          return elem.getAttribute('type'); 
+        } else {
+          return '';
+        }
```
plocket: If the handle doesn't actually represent an element, should this function stop completely? Should it throw an error? Maybe I've forgotten something else this does. I'll try to take a closer look, but may not get to it.

lib/scope.js
```diff
@@ -2741,9 +2746,8 @@ module.exports = {
 
       // If this is a custom datatype
       let custom_datatype = await scope.get_custom_datatype( scope, { handle });
-      let is_custom_datatype = scope.get_random_input_for[ custom_datatype ] !== undefined;
+      let is_custom_datatype = scope.get_random_input_for[ custom_datatype ] !== undefined && type != 'radio';
       if ( is_custom_datatype ) {
```
```suggestion
      let is_custom_datatype = scope.get_random_input_for[ custom_datatype ] !== undefined;
      if ( is_custom_datatype && type !== `radio` ) {
```

Otherwise, the name `is_custom_datatype` doesn't mean what you'd expect it to mean. I'm not sure why this excludes custom datatypes that are radio buttons. Where do we deal with custom datatypes that are radio buttons?

lib/scope.js
```diff
@@ -1759,8 +1759,13 @@ module.exports = {
           let name = await label.evaluate( elem => { return elem.getAttribute('for'); });
         }
 
-        if ( type === `radio` ) { await label.evaluate( elem => { return elem.click(); }); }
-        else { await scope.setCheckbox( scope, { label, answer }); }
+        if ( type === `radio` ) {
+          if (answer.toLowerCase() == "true") {
```
plocket: After a bit of trial and error, I have a guess as to what happened here. In the non-random flow, if a radio button is going to be pressed, only that radio button will get into here. The `answer` is the DOM `value` of the radio button. A an `answer` for a 'no' radio button will probably be "False". If it's the set of choices, it'll be something like "radio_other_opt_3".

That is, by the time the handle is in here, this radio button should definitely be pressed, no matter what the answer is.

If we want to control the flow of random radio buttons, I think that needs to happen upstream where random answers are dealt with. If we don't want a radio button to be pressed, it shouldn't get sent in here.

lib/scope.js
BryceStevenWilley: First, I don't think we have any custom datatypes that are radio buttons.

Second, this probably shouldn't need to exist, I just know that `get_custom_datatype` and `is_custom_datatype` would say that radio buttons were custom datatypes when they weren't actually. I would guess that `get_custom_datatype` isn't correct, but I don't know how to fix it without breaking other things.

lib/scope.js
BryceStevenWilley: I don't know what it should do. I had to add this fix because `elem` was `null`, just to avoid running into uncertain null errors. Since we don't know the context of the error fully, it's likely something with puppeteer (or our use of puppeteer), and I'd rather just not have that happen. Throwing another error IMO would be the same issue there, we'd have to catch the error everywhere to else to prevent errors propagating.

I'd be willing to add a log, though it can't go right there (because that's JS that's executing in the browser, not this machine).

lib/scope.js
BryceStevenWilley: Hm, I get it. A weird place for a parameter to have no effect, but I don't know if there's a refactor that makes it cleaner.

From what I can tell, there's no existing function to get the elements of a radio button? i.e. which values it can be set to? I'll have to let this sit until I can figure that out.

plocket: There's no way to get the options of a radio button group right now. Maybe separate the radio button stuff from the other stuff so that we can merge the other stuff? I have some shapes of ideas for radio button groups [though they are a bit messy].

BryceStevenWilley: Discussion about this ~35 minutes in Deep dive recording

lib/scope.js
plocket: Did a little experiment in the browser [and a bit of research]. When a type is null, the default type is 'text', so we'll use that instead of an empty string. I can do that when I'm working on peeling out the radio button stuff. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#input_types

lib/scope.js
plocket: Made #807

lib/scope.js
```diff
@@ -2231,9 +2240,14 @@ module.exports = {
     }  // ends act depending values
   },  // Ends scope.getThereIsAnotherValue()
 
+  is_on_interview_still: async function ( scope ) {
```
plocket: Change to `is_on_server_still`

lib/scope.js
```diff
@@ -2737,7 +2751,25 @@ module.exports = {
...
+      });
+      let field = null;
+      if (back_btn != null && Math.random() > 0.9) {
+        // 10% chance of clicking the back button
+        field = back_btn;
+      } else {
+        field = faker.helpers.arrayElement( fields );
+      }
+      if (field == null) {
+        // Just skip it? Shouldn't happen but has before
```
plocket: Add a debug log here that'll give us more info in the future.

lib/scope.js
```diff
@@ -2737,7 +2751,25 @@ module.exports = {
...
+      fields = fields.filter((ff) => {
+        if (ff.selector.includes("daquestionbackbutton")) {
+          back_btn = ff;
+          return false;
+        }
+        return true;
+      });
+      let field = null;
+      if (back_btn != null && Math.random() > 0.9) {
+        // 10% chance of clicking the back button
```
plocket: Explain why we want to check back button presses

lib/scope.js
```diff
@@ -1401,8 +1401,7 @@ module.exports = {
       let elem = document.querySelector( selector );
       // If the element has a sibling that's a label, we need to
       // interact with that label instead.
-      // `elem` has never been null from what we've seen
-      if ( elem.nextSibling && elem.nextSibling.tag === `label` ) {
+      if ( elem && elem.nextSibling && elem.nextSibling.tag === `label` ) {
```
plocket, line 1404: Log non-existent elems for debugging: Not sure how to add a debug log outside of the DOM here, but we should log an element situation that doesn't exist and maybe have some kind of response to that. Or we shouldn't send selectors for elements that don't exist. I'm not sure how that's even happening. Maybe we can send in `reports` to write in or some such.

Same below in `get_custom_datatype()` and `enter_answer.input()`.

plocket: Made https://github.com/SuffolkLITLab/ALKiln/issues/808

plocket (approved):
Some great fixes! I know we won't get another chance to review, so I'm calling it.

PR #642, opened 2023-02-03
Close #133, remove all uses of deprecated scope.setVar()

### LINKED ISSUE #133: Old `scope.setVar()` is still being used for AL name and address fields
[ E.g https://github.com/SuffolkLITLab/ALKiln/blob/releases/v4/lib/scope.js#L2244

`.setVar()` is deprecated and less robust and it doesn't print to the report. Use `.setFields()` instead.]

~I think it may be less robust than `.getField()` and `.setFields()`. It may not handle fields that have `show if` field properties. I'm considering repurposing it. Maybe it can use `.getField()` etc. to get, set, and return whether the field was disabled, enabled, or non-existant (`true`, `false`, `null` or something).~

plocket:
Closes #133

### DIFF
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -636,54 +636,6 @@ module.exports = {
 
     return false;
   },
-  
-  setVar: async function ( scope, var_name, value ) {
-    /* Use variable names and values to interact with buttons that
-    *    have values, text input elements, textareas, and dropdowns.
-    * Returns the element that was found.
-    * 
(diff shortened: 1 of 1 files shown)


BryceStevenWilley (approved):
LGTM! Was going to suggest some variable name changes, but the suggestions just doesn't work well, mostly because of my bad variable naming in the `normalizeTable` function 🙈

PR #652, opened 2023-02-12
Fixes #511, screenshot makes signature pages fail

### LINKED ISSUE #511: Can't take screenshot of signature pages
It causes an error because taking a screenshot erases the signature. This is technically a bug, but not one that's urgent so I'm going to avoid derailing us by leaving off that label. We can't think of a reason anyone other than us would want to take a screenshot of the signature page. It's going to be documented with https://github.com/SuffolkLITLab/docassemble-AssemblyLine-documentation/issues/212.

More details:

Docassemble currently handles signatures in-house, but hasn't accounted for the page zooming in and out, [which clears the drawing in the canvas](https://stackoverflow.com/questions/53263526/canvas-drawing-disappears-on-zoom-in-out/53275532). When kiln takes a screenshot, it uses the "full page" feature of puppeteer, which seems to zoom out on the page to get the whole thing in the viewport. That causes the signature to be erased.

A potential solution for the future is to take screenshots differently on signature pages [- don't take a full page screenshot]. That might be problematic if developers are trying to find out things about the page that are off-screen, though. An upstream fix is probably necessary, so we might focus on that instead. The code is [near the beginning of app.js](https://github.com/jhpyle/docassemble/blob/fdfff85381454fa2dbc44ed869b431cf4a3a4269/docassemble_webapp/docassemble/webapp/static/app/app.js#L105).

plocket:
Also one plain language improvement that didn't need to be in here. Fixes #511.

Sorry about the previous [unmerged] commits. Wanted to get this in to be able to add screenshots to story table runs. We've been needing them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,8 +39,11 @@ Format:
 - 
 -->
 ## [Unreleased]
-## Added
+### Added
 - Contribution docs
+
+### Fixed
+- #511, couldn't take screenshots of signature pages https://github.com/SuffolkLITLab/ALKiln/issues/511
 
 ## [4.10.3] - 2023-01-11
 ### Removed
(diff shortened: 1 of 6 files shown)


BryceStevenWilley (approved):
LGTM! Test passes, signature shows up in the screenshot too

Issue #550, opened 2022-04-07
Add better error messages to api interface delete_project()


plocket:
- [ ] Add better error messages to api interface `delete_project()`. Especially 403 ["access denied" and "project doesn't exist". The devs shouldn't be deleting projects by hand, but it's still very possible they would].
- [x] Maybe avoid showing full [server] response in error message. It's a lot of noise vs. the signal achieved.

PR #662, opened 2023-03-10
Use Axios's `toJSON` method to make shorter errors (#632)

### LINKED ISSUE #605: Remove axios full error log for setup (e.g. when Project is deleted or created), like with a 504

BryceStevenWilley:
Accidentally merged #632 into `releases/v4` instead of `v4` (the PR pre-dated `v4` as the default branch), so this process is backwards; but here's  the changes cherry-picked from `releases/v4`.

* Use Axios's `toJSON` method to make shorter errors

On an example erroring docassemble server (for example, this server returning a [502 to a delete project](https://github.com/SuffolkLITLab/docassemble-EFSPIntegration/actions/runs/3641406524/jobs/6147303425)), the entire axios error will print, which is that case in 1.8k lines, most of it being garbage.

Axios has a `toJSON()` method on errors that should reduce length of the errors being shown, as it will not print the most annoying parts of the long errors, like the internal Axios socket objects

Still need to test, as it's difficult to get a server to 502 on demand.

Fix #605.

* Actually fail when we log the errors
* Move `toJSON()` calls to docassemble_api_REST
* Add better comment to `process.exitCode`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,7 +38,9 @@ Format:
 ### Security
 - 
 -->
-<!-- ## [Unreleased] -->
+## [Unreleased]
+### Changed
+- Shorten Axios errors to make them more readable (https://github.com/SuffolkLITLab/ALKiln/pull/632)
 
 ## [4.10.4] - 2023-02-15
 ### Added
--- a/lib/docassemble/docassemble_api_REST.js
+++ b/lib/docassemble/docassemble_api_REST.js
@@ -24,7 +24,11 @@ da.get_dev_id = async function ( timeout ) {
     timeout: timeout,
   };
 
-  return await axios.request( options );
+  try {
+    return await axios.request( options );
+  } catch (error) {
+    throw error.toJSON();
+  }
 };  // Ends da.get_dev_id()
 
 
@@ -41,8 +45,11 @@ da.create_project = async function ( project_name, timeout ) {
     timeout: timeout,
   };
 
-  return await axios.request( options );
-
+  try {
+    return await axios.request( options );
+  } catch (error) {
+    throw error.toJSON();
+  }
 };  // Ends da.create_project()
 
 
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -38,7 +38,9 @@ Format:
 ### Security
 - 
 -->
-<!-- ## [Unreleased] -->
+## [Unreleased]
+### Changed
+- Shorten Axios errors to make them more readable (https://github.com/SuffolkLITLab/ALKiln/pull/632)
```
plocket, line 43: I think the issue number isn't what you intended?

```suggestion
- Shorten Axios errors to make them more readable (https://github.com/SuffolkLITLab/ALKiln/pull/605)
```

plocket: I also liked the list of changes in the comment so I put them here, but if you don't want to include that, that's fine. Just an alternative version.

```suggestion
- Shorten Axios errors to make them more readable (https://github.com/SuffolkLITLab/ALKiln/pull/605)
- Actually fail when we log the errors
- Move `toJSON()` calls to docassemble_api_REST
- Add better comment to `process.exitCode`
```

plocket (approved):
Great! Thank you! You can take or leave the comment.

plocket:
Also, do you think this also handles #550 sufficiently?

CHANGELOG.md
BryceStevenWilley: I do intentionally link to the PR when I fix things. Issues usually have a lot of extra discussion that don't get implemented, and the issue is still linked to from the PR, usually in the first comment. I've found it's a lot easier to find the issue fixed from a PR than finding the PR from the issue.

I normally would be fine with adding more details to the changelog, but those specific commit comments are very internal, and don't make sense outside the context of the choices made in the PR ("move toJSON calls" move them from where? from a different file they were in earlier in the PR. From the changelog comment, it looks like those toJSON calls were around before that release, but they weren't).

BryceStevenWilley:
TBH, I don't think it addresses #550, at least not the first part. I'll mark the second part as being done there.

CHANGELOG.md
plocket: Those make some sense. I'm good with whatever you choose.

PR #665, opened 2023-03-16
Show axios error reason with JSON

BryceStevenWilley:
I didn't notice until trying to debug some Axios responses that were running with #662, but for some reason, axios doesn't include `error.response.data`, the attribute that says why the API call failed (on docassemble at least) in `error.toJSON()`.

This change adds it back in the returned JS object manually, additionally optional chaining to make sure that it doesn't crash if `error.response` isn't present for some reason. Optional chaining is supported in Node 14 and above, which we are on.


plocket (approved):
Great! Thanks! Pls move to v5 or do another PR for v5

BryceStevenWilley:
v5 PR is #670

PR #672, opened 2023-03-22
Pass API Key through headers, not param

BryceStevenWilley:
Params are printed in the uWSGI logs on docassemble, potentially leaking the API key to other developers.

### DIFF
--- a/lib/docassemble/docassemble_api_interface.js
+++ b/lib/docassemble/docassemble_api_interface.js
@@ -222,7 +222,7 @@ da_i.loop_to_create_project = async function( creation_options ) {
 
     } catch ( error ) {
       // TODO: Make a `normalize_ajax_responses_and_errors()`?
-      let response = error.response;
+      let response = error;
 
       if ( response && response.data ) {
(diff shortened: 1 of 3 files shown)


lib/docassemble/docassemble_api_interface.js
```diff
@@ -222,7 +222,7 @@ da_i.loop_to_create_project = async function( creation_options ) {
 
     } catch ( error ) {
       // TODO: Make a `normalize_ajax_responses_and_errors()`?
-      let response = error.response;
+      let response = error;
```
plocket, line 225: This is unrelated, right? A different bug?

plocket (approved):
Great catch and looks good, thank you!

lib/docassemble/docassemble_api_interface.js
BryceStevenWilley: Yes, different bug but I did need to fix it to test. I think it's related to those 400 errors we've been seeing with Caroline and elsewhere recently.

PR #673, opened 2023-03-22
Pass API Key through headers, not param

BryceStevenWilley:
Params are printed in the uWSGI logs on docassemble, potentially leaking the API key to other developers.

Cherry-picked from #672 onto the v5 branch, since it should be in both.


plocket (approved):
Same here, and thanks for ensuring the PR is on both branches 🌹

PR #679, opened 2023-03-28
596 signature text

### LINKED ISSUE #596: Step: Allows dev to input letters so the robot will draw letters on the signature canvas
Find a package that let the testing framework draw letters on a canvas. Our users, the developers writing the tests, would specify a name for the bot to draw on the canvas. Not at all urgent, but a nice addition for forms that can have multiple signatures.

First find the package, then we'll see how we can integrate it.

jostran14:
Added/modified sign functionality to take a name argument and draw on canvas.
Closes #596


plocket (approved):
Great improvement and great work, thanks!

Issue #663, opened 2023-03-13
Make da Playground Project name unique with timestamp


plocket:
Our test implementation is on [a dev branch/release](https://github.com/SuffolkLITLab/ALKiln/tree/debug_cache), but is mixed in with some other experimental changes.

X of the problem: We're finding we're having some caching issues with @CaroRob's tests - tests on the same branch will use previous yml code, not updating to the newly pulled code. When a branch's code is updated in relevant ways, its tests will continue to behave as they previously had. A new branch checked out of that branch with no code changes will behave properly. Also, after a few days, the problem will disappear and the tests will behave appropriately again. [This last implies that if a Project is being cached, da does eventually clean it up.]

Possibly, through the API the Project's name gets deleted, but the Project itself doesn't get fully deleted, and pulling new code doesn't actually pull new code. That seems strange, but we don't have other guesses right now.

Y of the problem (and the purpose of this issue): We think making sure the Project name is unique will solve this. We implemented it on [a dev branch/release](https://github.com/SuffolkLITLab/ALKiln/tree/debug_cache) and the problem hasn't showed up again yet. We hope that if this is indeed the problem, lingering Projects will clean themselves up - after all, the problem does disappear after a few days.

plocket:
## GitHub ticket
_In case GitHub stuff is involved with this as well._

> Product
> 
> GitHub
> 
> Description
> 
> Actions don't seem to be updating based on code updates - test output for a branch doesn't change even when the code changes until the user makes a new branch with the same code. Not blocking, but our users are getting frustrated working around this. We don't mind that the tests are failing - the new branch's tests are failing in expected ways - we just mind that the action output doesn't change when the code changes.
> 
> What we saw a few weeks ago on a different repo too:
> 
> Branch 1 starts failing specific end-to-end tests (e.g. [#20](https://support.github.com/ticket/20), [#21](https://support.github.com/ticket/21), [#22](https://support.github.com/ticket/22)). Rerunning results in the same consistently failing tests 5 times in a row.
> Checkout a new branch from Branch 1 (Branch 2a). Make no changes.
> Branch 2a (correctly) fails only one test, e.g. [#22](https://support.github.com/ticket/22), consistently.
> Checkout another new branch from Branch 1 (Branch 2b). Make no changes.
> Branch 2b (correctly) fails only one test, e.g. [#22](https://support.github.com/ticket/22), consistently, same as Branch 2a.
[...]

PR #684, opened 2023-04-06
Address #663 for v5, avoid Project caching issues

plocket:
Address #663

### DIFF
--- a/lib/docassemble/docassemble_api_interface.js
+++ b/lib/docassemble/docassemble_api_interface.js
@@ -204,16 +204,17 @@ da_i.loop_to_create_project = async function( creation_options ) {
   let wait_seconds = creation_options.wait_seconds || 5;
 
   let project_name = null;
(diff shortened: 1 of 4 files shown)


lib/docassemble/docassemble_api_interface.js
```diff
@@ -204,16 +204,17 @@ da_i.loop_to_create_project = async function( creation_options ) {
   let wait_seconds = creation_options.wait_seconds || 5;
 
   let project_name = null;
-  let name_incrementor = 0;
+  // Unlikely, but 2 devs could run locally and create the same Project name
```
BryceStevenWilley, line 207: `Date.now` goes to milisecond precision, and IMO it's almost impossible to get the same milisecond. But since the code already handles it, it's fine to keep it there to simplfy the PR

BryceStevenWilley (approved):

Issue #661, opened 2023-03-10
Allow dev to run alkiln in the playground or something similar


plocket:
Right now through an interview.

Will add more, but quick note about sandbox: https://github.com/puppeteer/puppeteer/blob/main/docs/troubleshooting.md#setting-up-chrome-linux-sandbox

[Edit:
- [ ] Complete MVP items
- [ ] Write issues for all items in new repo

See [list of items to complete below](https://github.com/SuffolkLITLab/ALKiln/issues/661#issuecomment-1486046122)
]

BryceStevenWilley:
We have a proof of concept! Kiln and puppeteer can run on docassemble, with few caveats that we need to fix:

* as mentioned above, we are running puppeteer with no sandbox. IMO, this is not as important, because we are only visiting the server that the code is running on; if you don't trust anything that you could download from your server, I've got bad news for you; it's already on your server. Eventually we can look at [puppeteer's suggestions for sandboxing on linux](https://github.com/puppeteer/puppeteer/blob/main/docs/troubleshooting.md#setting-up-chrome-linux-sandbox), the setuid approach might work on docket with some extra setup steps, i.e. login to your docker container and run these commands. Some more sources on how to get things working, all pretty involved: https://stackoverflow.com/questions/62345581/node-js-puppeteer-on-docker-no-usable-sandbox
* we need to add back the env variable for playground_id to Kiln. Right now, we assume that we get that information with the API key, but if we run on the server itself, we don't need the API key. The python side (i.e. running kiln through the interview) should be able to get us the user id ([`user_info().id`](https://docassemble.org/docs/functions.html#user_info)), we just need to be able to pass that info back to Kiln again. This info was previously removed in #513, so we should be able to add it back by looking at those changes.
* getting everything to run through python in the background task. We have it all in bash, so things should be good there
* undoing the hacky-ness that we've had to add in [the branch for this](https://github.com/SuffolkLITLab/ALKiln/tree/cli_playground)
    * be able to pass in the paths that we want to search for `*.feature` files through the command line. docassemble stores the source files in such a strange location that the current glob can only find it in downloaded or installed projects, not in a playground project.
    * making sure that the sandbox option can also be configured; we do not want to start running without a sandbox on GitHub Actions

plocket:
[...]

PR #686, opened 2023-04-06
#661 ALKiln in the Playground

plocket:
Address #661, ALKiln in the Playground.

- Adds new session vars/env vars
- Adds new checks for those vars, splitting checks depending on environment
- Adds tests for those new checks

### DIFF
--- a/lib/run_cucumber.js
+++ b/lib/run_cucumber.js
@@ -23,6 +23,8 @@ const cucumber_api = require('@cucumber/cucumber/api');
   // If they're running it from an interview on their server instead of from GitHub
   if ( session_vars.get_origin() === `playground` ) {
     provided_config.paths.push(`/usr/share/docassemble/files/playgroundsources/${session_vars.get_user_id()}/${session_vars.get_user_project_name()}/*.feature`);
+    // For S3
+    provided_config.paths.push(`/tmp/playgroundsources/${session_vars.get_user_id()}/${session_vars.get_user_project_name()}/*.feature`);
   }
 
   let tags = argv.slice(2);
--- a/lib/run_cucumber.js
+++ b/lib/run_cucumber.js
@@ -9,8 +9,6 @@ const cucumber_api = require('@cucumber/cucumber/api');
 (async function main() {
   let argv = process.argv;
 
-  let just_args = argv.slice(2);
-
   const environment = { cwd: process.cwd() };
   console.log("alkiln-run: environment: %o", environment);
   const provided_config = {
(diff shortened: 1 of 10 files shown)


lib/run_cucumber.js
```diff
@@ -20,21 +18,36 @@ const cucumber_api = require('@cucumber/cucumber/api');
...
+  if ( session_vars.get_origin() === `playground` ) {
+    provided_config.paths.push(`/usr/share/docassemble/files/playgroundsources/${session_vars.get_user_id()}/${session_vars.get_user_project_name()}/*.feature`);
+    // For S3
+    provided_config.paths.push(`/tmp/playgroundsources/${session_vars.get_user_id()}/${session_vars.get_user_project_name()}/*.feature`);
+  }
+
+  let tags = argv.slice(2);
+  if ( session_vars.get_origin() === `playground` ) {
+    tags = session_vars.get_tags();
+  }
```
BryceStevenWilley, line 33: I'm confused by this. Specifically, why we would need to pass tags from the playground through env vars, but everything else still uses command line parameters?

lib/run_cucumber.js
```diff
@@ -20,21 +18,36 @@ const cucumber_api = require('@cucumber/cucumber/api');
...
     "retry": 1,
   }
-  if (just_args.length > 0) {
+
+
+  // If they're running it from an interview on their server instead of from GitHub
+  if ( session_vars.get_origin() === `playground` ) {
+    provided_config.paths.push(`/usr/share/docassemble/files/playgroundsources/${session_vars.get_user_id()}/${session_vars.get_user_project_name()}/*.feature`);
+    // For S3
+    provided_config.paths.push(`/tmp/playgroundsources/${session_vars.get_user_id()}/${session_vars.get_user_project_name()}/*.feature`);
```
BryceStevenWilley, line 27: This is fine for now, but I didn't see an issue anywhere, so I wanted to note; I think the next refactor here should be do add a proper command line argument handler like [minimist](https://github.com/minimistjs/minimist), and then pass these two paths through the command line instead. It has a few benefits:
* removes the necessity of 2 additional environment variables as well, `get_user_id`, and `get_user_project_name`
* puts docassemble specific knowledge (what these specific paths are) in the docassemble specific package (ALKilnInThePlayground)
* makes it more flexible in general; I made a quick little tool that could run 10 different versions of Kiln at the same time by running them in different working directories and pointing them to the source directory elsewhere. 

I did a little preliminary work on this in https://github.com/SuffolkLITLab/ALKiln/commit/f8e9921788abc3c45748a8090e451275081145e9, but I'm trying to shift priorities for a bit, and don't with this to be slowed down by me.

BryceStevenWilley (approved):
LGTM!

lib/run_cucumber.js
plocket: Hmm, I suppose we could change the ALKilnInThePlayground code to pass those params as well in the command. Ok, that should be an issue in [the Playground repo](https://github.com/SuffolkLITLab/docassemble-ALKilnInThePlayground). Part of the issue should be to then make an issue here to clean up this code. It'd be good to get rid of yet another env var. Thanks!

plocket: [Issue made](https://github.com/SuffolkLITLab/docassemble-ALKilnInThePlayground/issues/2). I think we'll leave this as is here for now and make this change as I described above since it's easy to avoid breaking things that way.

lib/run_cucumber.js
plocket: Makes a ton of sense! #687

PR #689, opened 2023-04-08
Address #618 v5, filepaths too long

plocket:
Addresses #618. Approximate example of longest (I think) file path:

alkiln-2023-04-08 at 02h08m58sUTC/alkiln-2023-04-08 at 02h08m58sUTC/Longest filepath should be under 184 characters with scenario descript-1680919740070/random1680919742254/random-25-characters-questi1680919742327.jpg

That's 217 characters.

[Just realized that when we have languages, the language name will also be added. Not sure what to do about that.]

### DIFF
--- a/docassemble/ALKilnTests/data/sources/reports.feature
+++ b/docassemble/ALKilnTests/data/sources/reports.feature
@@ -401,7 +401,7 @@ Scenario: Fail to find var while keeping value secret
   """
   And the Scenario report should include:
   """
-  For security, ALKiln will avoid creating a picture for this error.
+  For security, ALKiln will avoid creating a picture of the page for this error.
   """
   And I start the interview at "test_secret_vars"
   And I set the var "missing_var" to the secret "SECRET_FOR_MISSING_FIELD"
--- a/docassemble/ALKilnTests/data/sources/reports.feature
+++ b/docassemble/ALKilnTests/data/sources/reports.feature
@@ -401,7 +401,7 @@ Scenario: Fail to find var while keeping value secret
   """
   And the Scenario report should include:
   """
-  For security, ALKiln will avoid creating a screenshot for this error.
+  For security, ALKiln will avoid creating a picture for this error.
   """
   And I start the interview at "test_secret_vars"
   And I set the var "missing_var" to the secret "SECRET_FOR_MISSING_FIELD"
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -121,8 +121,7 @@ module.exports = {
     * create it. */
     if ( !scope.report ) { scope.report = new Map(); }
     if ( !scope.scenario_id ) {
-      let date = files.readable_date();
-      scope.scenario_id = `_report_${ date }`;
+      scope.scenario_id = `_report_${ Date.now() }`;
     }
     if ( !scope.report.get( scope.scenario_id )) {
       scope.report.set( scope.scenario_id, {} )
@@ -243,7 +242,7 @@ module.exports = {
     let filename = name_parts.join(`-`);
     let safe_name = safe_filename( filename, { replacement: `_` });
     // Allow room for extensions, date, etc.
-    let safer_name = safe_name.substring(0, (255 - 45));
+    let safer_name = files.below_71( safe_name );
 
     return safer_name;
   },  // Ends scope.getSafeScenarioBaseFilename()
@@ -254,6 +253,14 @@ module.exports = {
     *    a prefix the caller wants, and the base filename. We hope it can
     *    contain non-english characters too. We do not offer a suffix option
     *    because it may be cut off if the filename is too long.
+    * 
+    * @Examples
+    * Return value possible formats
+    * prefix-id-scenario_description-timestamp
+    * id-scenario_description-timestamp
+    * scenario_description-timestamp
+    * 
+    * @returns str
     */
     let name = options.prefix || '';
 
@@ -264,23 +271,20 @@ module.exports = {
 
     if ( scope.page ) {
       let { id } = await scope.examinePageID( scope, 'none to match' );
-      name += `${ id }-`;
+      name += `${ files.below_21( id ) }-`;
     }
 
     name += scope.base_filename;
-    name += `-${ files.readable_date() }`;
+    name += `-${ Date.now() }`;
 
     let safe_name = safe_filename( name, { replacement: `_` });
-    // Allow room for extensions, etc.
-    let safer_name = safe_name.substring(0, (255 - 10));
-    return safer_name;
+    return safe_name;
   },  // Ends scope.getSafeScenarioFilename()
 
   getJSONFilename: async function( scope ) {
     // Get a unique path for a json file
     let { id } = await scope.examinePageID( scope, 'none to match' );
-    let date = files.readable_date();
-    let json_path = path.join( scope.paths.scenario, `json_for-${ id }-${ date }.json` );
+    let json_path = path.join( scope.paths.scenario, `json_for-${ id }-${ Date.now() }.json` );
     return json_path;
   },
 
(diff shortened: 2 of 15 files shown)


lib/utils/files.js
```diff
@@ -38,5 +37,13 @@ files.make_artifacts_folder = function (folder_name) {
   return folder_name;
 };  // Ends files.make_artifacts_folder()
 
+files.below_21 = function (page_id) {
+  return `${ page_id }`.substring(0, 20);
+};
+
+files.below_71 = function (scenario_name) {
+  return `${ scenario_name }`.substring(0, 70);
+};
```
BryceStevenWilley: It took me a bit, but I do understand why these exist (since the inputs might be null, you need to wrap it in a `${ ... }`, which is hard to do nested).  I'd still lean towards just making one function instead of two, and not hard coding the name of the function to exact numbers.

```suggestion
files.below_length = function (length, input) {
  return `${ input }`.substring(0, Math.max(length - 1, 0));
};
```

lib/steps.js
```diff
@@ -1131,7 +1131,7 @@ After(async function(scenario) {
       if ( !!scope.disable_error_screenshot ) {
         scope.addToReport(scope, {
           type: `row`,
-          value: `For security, ALKiln will avoid creating a screenshot for this error. It's possible a secret is being used on this screen.`
+          value: `For security, ALKiln will avoid creating a picture for this error. It's possible a secret is being used on this screen.`
```
BryceStevenWilley: To confirm my understanding: did you change "picture" to "screenshot" because the files now have "picture" in the filename and not "screenshot"?

I'm down with adding an option to say "and I take a pic", but conflicted because "screenshot" really effectively communicated what exactly we're avoiding doing, and saying picture in all of the user-shown text is much more vague. "Picture of the page" is better, but a little wordier, and you haven't used it everywhere (not sure if that's intentional).

I'm hoping that consistency between the user-shown text and the filenames isn't as important in this case, and that folks who run on GitHub will know that the *.jpg files are screenshots, but idk if that's accurate.

BryceStevenWilley (approved):
I'd strong suggest taking `21` and `71` out of the function names, and a minor nit about screenshot vs picture in the user-shown text, but would be fine with things as is too.

plocket:
> minor nit about screenshot vs picture in the user-shown text

I changed it because the filename is going to be 'pic' instead of 'screenshot' to make it shorter. Thoughts?

BryceStevenWilley:
> I changed it because the filename is going to be 'pic' instead of 'screenshot' to make it shorter. Thoughts?

@plocket See https://github.com/SuffolkLITLab/ALKiln/pull/689#discussion_r1167193926

lib/utils/files.js
plocket: This is a good question to bring up and I agree we might need to change the names. I do think it should stay 2 separate functions. Not sure quite how to articulate this atm, but I'll do my best!

There are two lengths we want to limit to. One is the `id` name max length, which is shorter, which we decided was 21 chars but we might change. The other is the scenario name max length. I don't want to decide at each location where these functions are being called how many characters to allow. If we have to update it, we're then stuck updating in multiple places. Also, it lets someone in the future come by and just change it without much thought (which makes me think these need function descriptions too). I think it might also be possible to name these well enough to clarify what they're being used for, though I can't think of the names right now. I decided against `truncate_to_id_max_length` etc. is because I think that when the scenario description isn't in the mix of the path name, maybe we could allow the question id to be longer. Or other such things.

Does that explain a bit? At least about why they're separate and very specific functions.

lib/steps.js
plocket: Sorry, should have read everything before responding to the top post. I'm currently more concerned about a mismatch in language. I've seen folks get really confused about stuff like that. I think elsewhere I changed it to "picture of the page", as you said, and intended to be consistent. Maybe sometimes use one and sometimes another? I dunno. May be worth asking an outside consultant.

lib/utils/files.js
BryceStevenWilley: Thanks, these explanations help a lot.

There are some conflicting arguments:

> I don't want to decide at each location where these functions are being called how many characters to allow. If we have to update it, we're then stuck updating in multiple places. Also, it lets someone in the future come by and just change it without much thought (which makes me think these need function descriptions too).

This argues towards the existing code, just with different names (TBH `truncate_to_id_max_length` is good here).

> I decided against truncate_to_id_max_length etc. is because I think that when the scenario description isn't in the mix of the path name, maybe we could allow the question id to be longer.

This argues towards deciding the length at each call site; if how long we're truncating things to depends on how we're using it, that should be decided at the call site; anything else is gonna be a lot of extra functions and abstractions for limited benefit.

You've done a lot of great work at making choices and shortening things where you can, but IMO we're going to have to tweak the exact numbers, and I'd lean towards making it easier to do that.

lib/steps.js
BryceStevenWilley: > I think elsewhere I changed it to "picture of the page", as you said, and intended to be consistent

That's probably good enough. It is pretty minor, maybe I just have an emotional attachment to screenshot for some reason 📷

docassemble/ALKilnTests/data/sources/reports.feature
```diff
@@ -401,7 +401,7 @@ Scenario: Fail to find var while keeping value secret
   """
   And the Scenario report should include:
   """
-  For security, ALKiln will avoid creating a screenshot for this error.
+  For security, ALKiln will avoid creating a picture for this error.
```
BryceStevenWilley:
```suggestion
  For security, ALKiln will avoid creating a picture of the page for this error.
```

lib/steps.js
```diff
@@ -585,7 +585,7 @@ Then('I will be told an answer is invalid', async () => {  // √
   expect( was_found, no_error_msg ).to.be.true;
 
   let was_user_error = error_handlers.complex_user_error !== undefined || error_handlers.simple_user_error !== undefined;
-  let not_user_error_msg = `The error was a system error, not an error message to the user. Check the error screenshot.`;
+  let not_user_error_msg = `The error was a system error, not an invalid user answer. Check for the error picture if the step did not use any secret variables.`;
```
```suggestion
  let not_user_error_msg = `The error was a system error, not an invalid user answer. Check for the picture of the page's error if the step did not use any secret variables.`;
```

lib/steps.js
```diff
@@ -813,7 +813,7 @@ When(/I answer randomly for at most ([0-9]+) (?:page(?:s)?|screen(?:s)?)/, { tim
 
     let no_content_err = `The interview seemed to run into an error page`
       + ` after the page with the question id "${ id }".`
-      + ` See the screenshot in "${ random_input_path }".`;
+      + ` See the picture in "${ random_input_path }".`;
```
```suggestion
      + ` See the picture of the page in "${ random_input_path }".`;
```

lib/steps.js
```suggestion
          value: `For security, ALKiln will avoid creating a picture of the page for this error. It's possible a secret is being used on this screen.`
```

BryceStevenWilley (commented):
Marked other places we say "picture", and suggested we say "picture of the page" instead.

lib/utils/files.js
plocket: Those are fair points. In that case, I think I should get rid of these functions because it's just recreating `.substring()`. A couple less functions to maintain! Yay!

plocket: Adjusted. If you have a mo, I'd appreciate your eye on the changes.

BryceStevenWilley (approved):
LGTM!

lib/scope.js
```diff
@@ -264,23 +271,21 @@ module.exports = {
...
-    return safer_name;
+    return safe_name;
   },  // Ends scope.getSafeScenarioFilename()
 
   getJSONFilename: async function( scope ) {
     // Get a unique path for a json file
     let { id } = await scope.examinePageID( scope, 'none to match' );
-    let date = files.readable_date();
-    let json_path = path.join( scope.paths.scenario, `json_for-${ id }-${ date }.json` );
+    let json_path = path.join( scope.paths.scenario, `json_for-${ id }-${ Date.now() }.json` );
```
plocket, line 288: Need to shorten the id in the json path name. Gorammit. I think we might update this for v5, but not v4. New issue.

PR #690, opened 2023-04-08
Close #618, filepaths too long

### LINKED ISSUE #618: Make most output file and directory names shorter

plocket:
Merge v5 - #689 - first, then this will close #618.

> Approximate example of longest (I think) file path:
> 
> alkiln-2023-04-08 at 02h08m58sUTC/alkiln-2023-04-08 at 02h08m58sUTC/Longest filepath should be under 184 characters with scenario descript-1680919740070/random1680919742254/random-25-characters-questi1680919742327.jpg
> 
> That's 217 characters.
> 
> [Just realized that when we have languages, the language name will also be added. Not sure what to do about that.]

### DIFF
--- a/.github/workflows/test_tests.yml
+++ b/.github/workflows/test_tests.yml
@@ -44,7 +44,7 @@ jobs:
       - name: Set name of artifact folder with date and UTC time
         # https://www.shell-tips.com/linux/how-to-format-date-and-time-in-linux-macos-and-bash/#how-to-format-a-date-in-bash
         run: |
-          echo "ARTIFACT_NAME=alkiln-$(date +'%Y-%m-%d at %Hh%Mm%Ss' -u) UTC" >> $GITHUB_ENV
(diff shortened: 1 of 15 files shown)


plocket:
Closing until we can cherry-pick from a completed v5

PR #697, opened 2023-04-14
Adds npm-shrinkwrap.json file

### LINKED ISSUE #614: Security: Freeze both our and the dev's package with package-lock.json

BryceStevenWilley:
Ran `npm shrinkwrap`, which renamed my local package-lock.json to npm-shrinkwrap.json. Should work, will have to test publish to npm though.

Resolves #614 and resolves #532

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,7 @@ Format:
   be run in any directory, not just in an npm package.
 - Additional environment variables and their validation to allow for tests that run on a developer's server/Playground instead of through GitHub. Also, other functionality for that purpose. [Issue #661](https://github.com/SuffolkLITLab/ALKiln/issues/661)
 - Tests for new session_vars behavior and improve previous tests.
+- Adds `npm-shrinkwrap.json`, so installs from npm will have fixed version dependencies
 
 ### Changed
 - upgraded cucumber v8.6.0
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -112,6 +112,12 @@ Our tests require the user/developer to have a docassemble server on which they 
 
 When the developer commits code to GitHub, their account triggers our code, passing along variables that our code needs, like their GitHub secret of the API key for their docassemble server developer account.
 
+## Updating dependencies
+
+Sometimes you will have to add dependencies to ALKiln. You can do so by adding the dependency in the `package.json` "dependencies" section, specifying an exact version. Then, run `npm install` to update `npm-shrinkwrap.json` to have the new dependency.
+
+If you have to update an existing dependency, you can change the version in `package.json`, and run `npm install` to update `npm-shrankwrap.json`.
+
 ## Very general architecture of files and folders
 
 An honest look at our current project architecture—some of our files and folders, what they do, and how they interact.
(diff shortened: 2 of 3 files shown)


plocket (approved):
This is awesome! Does it also close #594? For some reason I can't seem to link 614 to close it, and 532 is a discussion. We'll have to do those manually. I'd also like to document the workflow for updating (and not updating) with shrinkwrap.

plocket:
Shall we add an entry to the changelog for this? It'll be getting its own release.

BryceStevenWilley:
I can update the changelog and document usage!

> Does it also close https://github.com/SuffolkLITLab/ALKiln/issues/594?

https://github.com/SuffolkLITLab/ALKiln/pull/624 should have closed that, but since we're working on the v5 branch, and v4 is still marked as the default branch, Github doesn't close things when merged. I'll close that one now.

>  532 is a discussion

that discussion's purpose was to decide on a mechanism to freeze dependencies, and this PR implements the one we decided on.

> It'll be getting its own release

We haven't actually released v5 yet, have we? This one is v5 only too, it won't work with v4.

plocket:
> We haven't actually released v5 yet, have we? This one isna v5 only too, it won't work with v4.

True, but I plan to make a new pre-release with this.

BryceStevenWilley:
Added to the CHANGELOG and CONTRIBUTING!

plocket:
Love the new docs! Instead of the `### Added` section, should this be in a `### Security` section?

I feel like we've covered these before, but I just can't locate the info - if someone just does `npm install` without changing the `package.json`, the shrinkwrap will stay the same? Secondly, do we have to edit the `package.json`? Can we no longer install using `npm install --save some-package`?

BryceStevenWilley:
> Instead of the ### Added section, should this be in a ### Security section

Happy to put it there, but my understanding was that Security was for things that are specific security fixes (i.e. we found this vulnerability and patched it, etc). This change is related to Security, and it's always good to update, but there's not the urgency of fixing a patch; we can't point to a specific library and say "this had a security flaw in it, this change avoids that", right?

> if someone just does npm install without changing the package.json, the shrinkwrap will stay the same?

That's my experience with it. Just to test it, I checked on a different machine, checked out this branch and ran `npm install`, and nothing in `npm-shrinkwrap.json` changed. Not sure all the other corner cases to test, but we will stall have control over all of the changes of `npm-shrinkwrap.json` going into main.

> Secondly, do we have to edit the package.json? Can we no longer install using npm install --save some-package?

I wanted to have one simple instruction for adding dependencies that didn't have any ambiguity, so I only documented that one. Extra benefit that the version in the package.json will more explicit and under our control, and not [the default semver range operator](https://docs.npmjs.com/cli/v9/commands/npm-install?v=true#save-exact) like `--save` does apparently.

 There are still a ton of ways to change `package.json` that were confusing to me, like `npm rm`, `npm update`, `--save`, `--save-exact`, `--save-prod`, and I couldn't tell what the difference was between each, so I just stuck with the simplest one. Those other steps probably still work fine; you can do whatever, and if you didn't expect to change `npm-shrinkwrap.json`, but it shows up in the PR diff (or vice versa), then you can always revert just that file. (I was looking at the `npm install` params that didn't have `--save`, only realized now that it was an older version of the docs, I thought it was a newer version).

plocket:
> This change is related to Security, and it's always good to update, but there's not the urgency of fixing a patch

Makes sense to me! All the rest are great answers too, thanks and looks great! If I could approve again, I would 🚀

PR #702, opened 2023-04-26
#597 json multiline: added a test for multiline json values

### LINKED ISSUE #597: Test the JSON variable text value Step with multi-line text
`(?:the )?text(?: in)?(?: the)?(?: JSON)?(?: var)?(?:iable)?(?: )?"([^"]+)" should be` (https://github.com/SuffolkLITLab/ALKiln/blob/releases/v4/lib/steps.js#L656)

We should test that the above Step can properly test a multi-line string. From testing in the wild, this does seem to work, so it's not urgent, but it would be good to have it under our belts.

jostran14:
specifically for variables set by code not through user input.
closes #597


CHANGELOG.md
```diff
@@ -40,6 +40,7 @@ Format:
 -->
 ## [Unreleased]
 ### Added
+- added a test for multiline json values (only ones set by code not by user input)
```
BryceStevenWilley: The "added" section usually has the specific meaning of "something added to the program interface", i.e. you as the user of this program can do something you couldn't before. This PR probably deserves it's own section, either "Internal" or "Testing", if it needs to be in the changelog at all.  Defering to @plocket, the changelog expert.

docassemble/ALKilnTests/data/sources/observation_steps.feature
```diff
@@ -99,23 +99,26 @@ Scenario: Test "Then I don’t continue" with an apostrophe
   And I will be told an answer is invalid
 
 @fast @o9 @table @json
-Scenario: I can match JSON page var to str
```
BryceStevenWilley, line 102: I'm not sure why this scenario was removed, it should likely still be around?

docassemble/ALKilnTests/data/sources/observation_steps.feature
```diff
@@ -99,23 +99,26 @@ Scenario: Test "Then I don’t continue" with an apostrophe
...
+  """
+  This block is.
+  A multiline block.
+  """
+  Then the text in the JSON variable "single_line_val" should be
   """
-  dropdown_opt_2
+  This is one line.
   """
+  #Then the text in the JSON variable "number_val" should be
```
BryceStevenWilley: From the comments in the issue, it looks like checking the number wasn't working properly. I'd remove the commented lines here and the `number_val` variable from the yaml interview if we don't want to try to make it work in this PR.

BryceStevenWilley (commented):
A good PR! The few suggestions that I'd recommend before merging are in comments

CHANGELOG.md
plocket: I like the idea of adding `### Internal` to our headings! Thanks!

jostran14: Thanks for the input, and we discussed and implemented an internal section to the change log.

docassemble/ALKilnTests/data/sources/observation_steps.feature
jostran14: @plocket and I considered these 2 tests equivalent. The purpose of this test is to make sure ALKiln can match a variable in the json to a value that the test author expects. We concluded that it shouldn't matter how that json variable is set. The previous version of this test used a variable set by filling in a field. The current version uses a variable set by python. We couldn't think of a reason that those 2 methods are testing anything different, and we didn't want duplicate tests. Do you still think we should restore the previous test?

BryceStevenWilley: Eh, yeah, I'd lean towards leaving it. I looked around, and surprisingly, there aren't any tests that seem to actually verify the value of a selection made by the user; we just say to input it and continue on. You're right that it's redundant for testing our JSON stuff, but we also need to test docassemble itself, which that test does.

plocket: We can add it back in, but if we're going to test selections, etc, I think that test belongs in its own set of tests and not buried in a JSON Step test. Again, we can still add this back in here for the moment and just make an issue to move it later and add further tests, it just doesn't seem like this is the appropriate place for it. I'm not sure what you mean by testing docassemble itself and would love to talk more about that if we get a chance. I those tests, too, would belong in a different set of tests.

plocket:
Just need to make the issue 👍

docassemble/ALKilnTests/data/sources/observation_steps.feature
jostran14: https://github.com/SuffolkLITLab/ALKiln/issues/742

plocket (approved):
Looks great! As soon as the tests pass, this can be merged!

PR #705, opened 2023-05-09
Check that inner HTML of error elements isn't blank

### LINKED ISSUE #704: Kiln thinks that Three Parts Date has an error
## Background

The [ALAffidavitOfIndigency](https://github.com/SuffolkLITLab/docassemble-ALAffidavitOfIndigency) uses the ThreePartDate. When running tests for the MotionToStayEviction, which uses the Affidavit, I ran into the following error:

```
The test tried to continue to the next page after qualify-on-face, but the user got an error. Check the picture of this page.
```

However, the screenshot wasn't of that screen (which only has a continue button) at all; it was of the next screen, asking the user to fill in their birthdate, a value that is also in the story table.

Trying to figure out where the error is actually through, I print debugged until I realized that `scope.waitUntilContinued` calls `scope.checkForError` before actually continuing, and `checkForError` was finding an error on the page. Couldn't really figure out much more about the error due to the [fact that ElementHandlers just print out `JSHandle@node`](https://github.com/puppeteer/puppeteer/issues/4976) when turned into strings. So I had to extend the debug window's timeout a bit to search for the same query selectors that we do in `checkForError`, and found that the ThreePartDate has an element that has the `da-has-error` class, it's just blank.

```html
<div class="col-md-8 dafieldpart">
    <div class="al_three_parts_date form-row row">
        <div class="col col-month">
            <label for="_ignore_dXNlcnNbMF0uYmlydGhkYXRl_month">Month</label>
            <select class="form-select al_field month dXNlcnNbMF0uYmlydGhkYXRl" id="dXNlcnNbMF0uYmlydGhkYXRl_month" name="_ignore_dXNlcnNbMF0uYmlydGhkYXRl_month" aria-describedby="dXNlcnNbMF0uYmlydGhkYXRl-error" required="required">
                <option value=""></option>
                <option value="01">January</option>
                <option value="02">February</option>
                <option value="03">March</option>
                <option value="04">April</option>
                <option value="05">May</option>
                <option value="06">June</option>
                <option value="07">July</option>
                <option value="08">August</option>
                <option value="09">September</option>
                <option value="10">October</option>
                <option value="11">November</option>
                <option value="12">December</option>
            </select>
        </div>
        <div class="col col-3 col-day">
            <label for="_ignore_dXNlcnNbMF0uYmlydGhkYXRl_day">Day</label>
            <input
                class="form-control al_field day dXNlcnNbMF0uYmlydGhkYXRl"
                type="number"
                min="1"
                inputmode="numeric"
                id="dXNlcnNbMF0uYmlydGhkYXRl_day"
                name="_ignore_dXNlcnNbMF0uYmlydGhkYXRl_day"
                aria-describedby="dXNlcnNbMF0uYmlydGhkYXRl-error"
                required="required"
            />
        </div>
        <div class="col col-3 col-year">
            <label for="_ignore_dXNlcnNbMF0uYmlydGhkYXRl_year">Year</label>
            <input
                class="form-control al_field year dXNlcnNbMF0uYmlydGhkYXRl"
                type="number"
                min="1"
                inputmode="numeric"
                id="dXNlcnNbMF0uYmlydGhkYXRl_year"
                name="_ignore_dXNlcnNbMF0uYmlydGhkYXRl_year"
                aria-describedby="dXNlcnNbMF0uYmlydGhkYXRl-error"
                required="required"
            />
        </div>
        <div id="al_dXNlcnNbMF0uYmlydGhkYXRl_error" class="da-has-error al_error"></div>
    </div>
    <input alt="Input box" class="form-control daBirthDate" type="hidden" name="dXNlcnNbMF0uYmlydGhkYXRl" id="dXNlcnNbMF0uYmlydGhkYXRl" required="" aria-hidden="true" style="display: none;" />
</div>
```

## Potential fixes:

* ignore things found by the search that have no inner text and not child elements. Seems hacky, but it works.
* fix ALToolbox to not have a div with the class `da-has-error`. This would involve more jquery stuff, and I'm not sure how all the validator works. So while it's a bit cleaner, I'm not sure how to approach it.

BryceStevenWilley:
Fixes #704 in one possible way (the hacky way mentioned at the bottom of the issue, by ignoring empty error elements).

This assumes that all error elements will have some inner HTML children, or text.

### DIFF
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -2269,12 +2269,15 @@ module.exports = {
     let error_handlers = {};
     let was_found = false;
     for ( let error_key in possible_errors ) {
-      if ( possible_errors[ error_key ][0] ) {
+      if ( possible_errors[ error_key ][0]) {
+        let innerHTML = await possible_errors[ error_key][0].evaluate(node => node.innerHTML, possible_errors[error_key][0]);
+        if (innerHTML) {
         // If you're debugging and you close the window early, the process will still
         // take this long to exit after the tests finish. Not sure how to account for that.
         if ( session_vars.get_debug() ) { await scope.page.waitForTimeout( 60 * 1000 ); }
         was_found = true;
         error_handlers[ error_key ] = possible_errors[ error_key ][0];
+        }
       }
     }
 
@@ -2286,7 +2289,7 @@ module.exports = {
     if ( error_handlers.system_error !== undefined ) {
       let sys_err_message = `The test tried to continue to the next page after ${ id }, but there was a system error. Check the picture of this page.`
       await scope.addToReport(scope, { type: `error`, value: sys_err_message });
-      expect( was_system_error, sys_err_message ).to.be.false;
+      expect( true, sys_err_message ).to.be.false;
 
     } else {
       // If it wasn't a system error, it was a user error


lib/scope.js
```diff
@@ -2286,7 +2289,7 @@ module.exports = {
     if ( error_handlers.system_error !== undefined ) {
       let sys_err_message = `The test tried to continue to the next page after ${ id }, but there was a system error. Check the picture of this page.`
       await scope.addToReport(scope, { type: `error`, value: sys_err_message });
-      expect( was_system_error, sys_err_message ).to.be.false;
+      expect( true, sys_err_message ).to.be.false;
```
plocket, line 2292: Was this meant to be changed back?

lib/scope.js
```diff
@@ -2269,12 +2269,15 @@ module.exports = {
...
-      if ( possible_errors[ error_key ][0] ) {
+      if ( possible_errors[ error_key ][0]) {
+        let innerHTML = await possible_errors[ error_key][0].evaluate(node => node.innerHTML, possible_errors[error_key][0]);
+        if (innerHTML) {
         // If you're debugging and you close the window early, the process will still
         // take this long to exit after the tests finish. Not sure how to account for that.
         if ( session_vars.get_debug() ) { await scope.page.waitForTimeout( 60 * 1000 ); }
         was_found = true;
         error_handlers[ error_key ] = possible_errors[ error_key ][0];
+        }
```
plocket, line 2280: Code in new `if` needs indenting.

```suggestion
        if (innerHTML) {
          // If you're debugging and you close the window early, the process will still
          // take this long to exit after the tests finish. Not sure how to account for that.
          if ( session_vars.get_debug() ) { await scope.page.waitForTimeout( 60 * 1000 ); }
          was_found = true;
          error_handlers[ error_key ] = possible_errors[ error_key ][0];
        }
```

plocket (changes requested):
Thanks for catching this and debugging! Just a couple things I'd like a check on before approving, [including rerunning the tests]. [This also makes me think we should up the priority of running scheduled tests.]

[Update: It's not actually the error I was expecting from the `true` value that was still in the code. Two random input tests are failing. Pics might show more.]

It's true "fixing" ThreePartsDate would be better. It's possible to put a totally different class on the container of the errors, but then we do have to make sure that the errors themselves do get that class and there was some kind of barrier to that at the time. The code has changed a lot, though. There's a way with the jQuery validation plugin to give the errors themselves a custom class instead of using the default class. I think if we lose the default class we may have to change the css and bootstrap may not recognize our invalid input class. We'll have to experiment when there's bandwidth.

In the meantime, a good fix for the moment!

lib/scope.js
BryceStevenWilley: ~Yes~  It was meant to be changed; `was_system_error` doesn't exist, and this change matches what we do when there is a user error. I noticed and changed it when I was trying to figure out what the error handler objects were and how they worked.

BryceStevenWilley:
The test errors are caused by the following:

* [We fill in the date input correctly](https://github.com/SuffolkLITLab/ALKiln/blob/f772f7d375c0d6e53abc2f54c865321cd3f92c27/lib/steps.js#L853).
* [We try to continue to the next page](https://github.com/SuffolkLITLab/ALKiln/blob/f772f7d375c0d6e53abc2f54c865321cd3f92c27/lib/steps.js#L864-L873). Something is wrong here, and we don't actually wait until the next page, but we act like we have.
* [we get the HTML content and the fields of the "next" page, which is still the date input page](https://github.com/SuffolkLITLab/ALKiln/blob/f772f7d375c0d6e53abc2f54c865321cd3f92c27/lib/scope.js#L2615-L2617)
* the page finally progresses to the final page; so the HTML content we had is now wrong
* [we try to get the handle from the field information we saved earlier](https://github.com/SuffolkLITLab/ALKiln/blob/f772f7d375c0d6e53abc2f54c865321cd3f92c27/lib/scope.js#L1540-L1544). This fails, since that field is no longer on the page. and that function assumes that we can always find the field.

I'm guessing that it's because the date input javascript validation takes just a bit too long, and what we currently wait for doesn't actually wait for that to finish. Not sure how to fix.

knod:
This is all quite possible. We've tried several different ways of detecting when the next page loads and none of them have been fool-proof so far. The page isn't timing out, so there's something that we actively look for to identify a loaded page that's getting tripped up. We'll have to check again through our "new page navigation" detection stuff to see what's getting detected.

Maybe it'd actually take less time to do the upstream work.

plocket:
My main concern is why this is coming up as a consistent problem now as opposed to before.

Regardless, we've tried several different ways of detecting when the next page loads and none of them have been fool-proof so far. The page isn't timing out, so there's something that we actively look for to identify a loaded page that's getting tripped up. We'll have to check again through our "new page navigation" detection stuff to see what's getting detected.

Maybe it'd actually take less time to do the upstream work for the date datatype.

lib/scope.js
plocket:
```suggestion
      expect( was_system_error, sys_err_message ).to.be.false;
```

```suggestion
      expect( false, sys_err_message ).to.be.false;
```

```suggestion
```

plocket: Or maybe

```suggestion
      expect( was_found, sys_err_message ).to.be.false;
```

plocket: So it's meant to be kept the way it is, right?

plocket: In which case, maybe we can figure out a better way than just true/false, which people seem to find confusing. But that doesn't have to happen this time around. Sorry I'm a little scattered at the moment.

BryceStevenWilley: Sorry, I misread your question as "was this meant to be changed".  It was an intentional change. I don't think it needs to be changed back to the original.  I'm making code at least consistent with itself (the same `expect(` approach is used a few lines down).

BryceStevenWilley:
> My main concern is why this is coming up as a consistent problem now as opposed to before.

That what my guess was, it's the ThreePartDate changes. They weren't tested super closely with the new ThreePartDate type, which now has a lot more jquery validation happening before the page submits and re-loads.

> Maybe it'd actually take less time to do the upstream work for the date datatype.

This problem is (most likely, I think) related to the jquery validation stuff, and not to this specific change. If we fixed the upstream date datatype by placing the `da-has-error` class only when there's an error, it wouldn't necessarily fix the failing random tests.

> We'll have to check again through our "new page navigation" detection stuff to see what's getting detected.

:+1: agreed. I can do that, just might not have the bandwidth to do it soon.

plocket:
I have a guess at a different cause. Usually when there's invalid input, the page won't continue. Thus ALKiln will catch the error and continuing to the next page won't happen. In this situation ALKiln is catching the error, but the page is able to continue and messes everything up.

Maybe you were already saying that and I misunderstood, though.

[Potentially, we could try to detect navigation before detecting an error, but it really shouldn't be necessary I don't think. It'll still misbehave under that circumstance.]

[Hmm, that might be why it misbehaved before the fix. Not sure why it's misbehaving after the fix, though.]

BryceStevenWilley:
Not sure what we should do with this PR; https://github.com/SuffolkLITLab/docassemble-ALToolbox/pull/177 has merged, which I think removed the need for this PR in the first place (though this could fix future errors), and https://github.com/SuffolkLITLab/ALKiln/pull/712 turned off the failing random tests that these comments above looked into.

I'll close and delete the branch for now (we can recreate the branch from the PR diff if need be).

PR #709, opened 2023-06-06
Close #706 improve loops

### LINKED ISSUE #706: Improve using `.target_number` in Story Table
- [x] Allow `.target_number` of 0 to set `.there_are_any` to `False`.
- [x] Allow `.target_number` to set `.there_are_any` in general so the author doesn't need to set `.there_are_any` separately.
- [x] Fix `.target_number` using one fewer loop than it should.
- [x] Create feature tests for those items.

plocket:
This is a bit of a biggie and might actually deal with a bug I wasn't aware of with `.target_number`, which was inconsistent depending on whether the author defined `.there_are_any` or not.

I also wrote docs on how the flow works now.

Closes #706, improve loops.


lib/steps.js
```diff
@@ -1123,8 +1128,6 @@ After(async function(scenario) {
   if (scenario.result.status !== `PASSED`) {
     if ( session_vars.get_debug() ) {
       console.log( `Non-passed status occurred while running test:`, scenario.result.status );
-      // If there's a page visible (locally), give us time to examine it
-      if ( scope.page ) { await scope.page.waitForTimeout( 60 * 1000 ); }
```
BryceStevenWilley, line 1108: Is this intentional? I use it this pause a lot, I understand modifying it when working with a specific issue, but do think it should stay

tests/unit_tests/ensureSpecialRows.test.js
BryceStevenWilley, line 1: I'm kinda torn on removing fixture tests. In general, they just run so much faster, and are something we could use to unit test much for thoroughly. But I do get that they're structured kinda weird, are pretty difficult to read, and are too strict (they can't handle feature additions, only removals).

I don't have any suggestions, just wanted to comment on their removal so it wasn't silent.

docassemble/ALKilnTests/data/questions/test_loops.yml
```diff
@@ -0,0 +1,66 @@
+metadata:
+  title: Loop tests
+  description: Test loops like as with .there_is_another and .target_number
```
BryceStevenWilley:
```suggestion
  description: Test loops with .there_is_another and .target_number
```

docassemble/ALKilnTests/data/questions/test_loops.yml
```diff
@@ -0,0 +1,66 @@
...
+---
+id: person name
+generic object: DAList
+question: Name of the "${ x[ i ].object_name() }" person
+fields:
+  - Name: x[i].name.first
+---
+id: end1
+event: end1
+question: The end of part 1!
```
BryceStevenWilley: Is there a part two? Test content IMO doesn't really matter, but this one was a bit confusing

docs/story_tables.md
```diff
@@ -0,0 +1,187 @@
+# Story Tables overview
+
+The Story Table is a very flexible tool. It lets the author answer interview (form) questions in any order. They can make cosmetic changes to their interview code without having to change their tests. They can edit the order of their interview's pages, move fields from one page to another, and so on.
+
+When the robot arrives on each interview page, it finds all the fields on the page and loops through the whole Story Table trying to match rows with fields.
+
+Here's some [documentation on the use of Story Table rows](https://suffolklitlab.org/docassemble-AssemblyLine-documentation/docs/alkiln/#story-tables).
+
+# Tracking use of rows
```
BryceStevenWilley:
```suggestion
## Tracking use of rows
```

docs/story_tables.md
BryceStevenWilley, line 1: I'm not sure how you want these internal docs to read; I made some edits to make them a bit more straight forward, which removes so of the Question and answer style that was confusing me. Feel free to ignore if you'd prefer to keep the inquisitive / education style though.

docs/story_tables.md
```diff
@@ -0,0 +1,187 @@
...
+
+For example, look for `.there_is_another` in the scenario below:
+
+Page 1: Do you have any moms? (Set `moms.there_are_any` to `True`)
+Page 2: What is your first mom's name? (Set `mom[0].name.first`)
+Page 3: Do you have any other moms? (Set `moms.there_is_another` to `True`)
+Page 4: What is your second mom's name? (Set `mom[1].name.first`)
+Page 5: Do you have any other moms? (Set `moms.there_is_another` to `False`)
+
+See that on page 3, `moms.there_is_another` is set to `True` and on page 5 it's set to `False`. How can we represent that in a Story Table? Pro tip: we can't.
```
BryceStevenWilley:
```suggestion
See that on page 3, `moms.there_is_another` is set to `True` and on page 5 it's set to `False`. We can't represent this in a Story Table.
```

docs/story_tables.md
```diff
@@ -0,0 +1,187 @@
...
+
+# Tracking use of rows
+
+Why do we track how many times each Story Table row was used to set a form's fields?
+
+First, it lets us give the author useful information. We can show the author which rows of their story table were or weren't used. This can help the author tell that some crucial variables aren't being set as they expected. It also outputs a new Story Table for them. If they had a bunch of extra rows adding cruft, like from our [story table generator](https://plocket.github.io/al_story/), it gives the author a table with only the variables that are needed for the test.
+
+Second, in some cases we need to actually throw an error if any row was left unused, so we have to keep track of that.
+
+Third, looping questions that gather multiple items in a list are tricky when the interview answers need the `.there_is_another` attribute. We can't use that attribute. Instead, we handle all gathering with `target_number` rows. What's the problem with that attribute? How do we solve it? And what's the architecture of that code?
```
BryceStevenWilley:
```suggestion
Third, looping questions that gather multiple items in a list are tricky when the interview answers need the `.there_is_another` attribute. We can't use that attribute. Instead, we handle all gathering with `target_number` rows.
```

docs/story_tables.md
```diff
@@ -0,0 +1,187 @@
...
+
+The target number row looks like this:
+
+```
+| x.target_number | 2 | moms.target_number |
+```
+
+That's no problem at all for our code. It has three unique variables. How can that help us with `.there_is_another`, though?
+
+Good question. If we require our users to use `.target_number`[^1] we have one unique variable to tell us how many times to say "Yes" to a `.there_is_another` question. We just have to do a bit of extra work to make that happen.
```
BryceStevenWilley:
```suggestion
That's no problem at all for our code. It has three unique variables. If we require our users to use `.target_number`[^1] we have one unique variable to tell us how many times to say "Yes" to a `.there_is_another` question. We just have to do a bit of extra work to make that happen.
```

docs/story_tables.md
```diff
@@ -0,0 +1,187 @@
...
+    2. Print a warning to the user if they've created a `.there_is_another` row with a value of `'True'` and change that value to `'False'`.
+    3. Add artificial rows - rows we create - which is currently just used for `.target_number`.
+    4. This new list will be used to actually put answers into fields.
+  3. For each field
+    1. Find which Story Table row, if any, matches that field.
+    2. Set the field to the given value.
+    3. Increment the row's accumulators.
+      1. Increment all relevant `.times_used` by 1.
+      2. Increment the relevant `.used_for['foo']` by 1.
+  4. Try to continue.
```
BryceStevenWilley: These need 4 indents to work. 

```suggestion
1. If this isn't a Story Table step, transform the given arguments into a Story Table structure - a list of row objects.
2. For each row object
    1. Create a new object with almost the same properties. It contains the original object in a property called `original` so the original can be used to print stuff for the user's report.
    2. In this new object transform some stuff if needed. For example, substitute environment variables the author gave into actual values. They may do this to keep a value (like a password) secret.
    3. Add ways to accumulate the number of times the row is used.
3. For each page
    1. Identify and store all the fields on the page and all the variables they set.
    2. Shallowly clone the Story Table row list.
        1. Create a new object almost the same as the old object
        2. Print a warning to the user if they've created a `.there_is_another` row with a value of `'True'` and change that value to `'False'`.
        3. Add artificial rows - rows we create - which is currently just used for `.target_number`.
        4. This new list will be used to actually put answers into fields.
    3. For each field
        1. Find which Story Table row, if any, matches that field.
        2. Set the field to the given value.
        3. Increment the row's accumulators.
            1. Increment all relevant `.times_used` by 1.
            2. Increment the relevant `.used_for['foo']` by 1.
    4. Try to continue.
```

docs/story_tables.md
```diff
@@ -0,0 +1,187 @@
...
+Good question. If we require our users to use `.target_number`[^1] we have one unique variable to tell us how many times to say "Yes" to a `.there_is_another` question. We just have to do a bit of extra work to make that happen.
+
+## The loop architecture
+
+First, the very basic flow for setting a variable. Note: every variable that is set uses the same code that handles Story Table rows.
+
+1. If this isn't a Story Table step, transform the given arguments into a Story Table structure - a list of row objects.
+2. For each row object
+  1. Create a new object with almost the same properties. It contains the original object in a property called `original` so the original can be used to print stuff for the user's report.
+  2. In this new object transform some stuff if needed. For example, substitute environment variables the author gave into actual values. They may do this to keep a value (like a password) secret.
```
BryceStevenWilley:
```suggestion
  2. In this new object substitute environment variables the author gave into actual values. They may do this to keep a value (like a password) secret.
```
I checked, and this is the [only thing happening now](https://github.com/SuffolkLITLab/ALKiln/blob/2cb76e0bad834b3f2bc510d200b76063ede23ee2/lib/scope.js#L909-L929). I understand why you added the "some stuff", to cover if the code changes, but reading it, it really doesn't tell me anything.

docs/story_tables.md
```diff
@@ -0,0 +1,187 @@
...
+  3. Add ways to accumulate the number of times the row is used.
+3. For each page
+  1. Identify and store all the fields on the page and all the variables they set.
+  2. Shallowly clone the Story Table row list.
+    1. Create a new object almost the same as the old object
+    2. Print a warning to the user if they've created a `.there_is_another` row with a value of `'True'` and change that value to `'False'`.
+    3. Add artificial rows - rows we create - which is currently just used for `.target_number`.
```
BryceStevenWilley:
```suggestion
    3. If there is a `.target_number` row, use it to create a new row for `.there_is_another`
```

docs/story_tables.md
```diff
@@ -0,0 +1,187 @@
...
+    4. This new list will be used to actually put answers into fields.
+  3. For each field
```
BryceStevenWilley:
```suggestion
  3. For each field on the page
```

docs/story_tables.md
```diff
@@ -0,0 +1,187 @@
...
+  used_for: { `x.there_is_another`: 0 }
+  //`.times_used` is the same
+}
+```
+
+Why is it all so nested? That's just how it evolved. It could probably use some refactoring.
+
+The parts relevant to loops for gathering multiple items in a list are the `.target_number` accumulators and the artificial rows. Whenever a `.there_is_another` row is used to set an answer to a field, we add 1 to its `source.times_used`. That is, we mutate the `.target_number` object's `.times_used` property. We alsoadd 1 to its `source.used_for['x.there_is_another']` value. That mutates the `'x.there_is_another'` property of the `.used_for` of the `.target_number` object. We do the same for `.there_is_another` rows, but that doesn't matter for now.
+
+Why do we increment both the `.times_used` property and the `.used_for` properties? Have patience and read on.
```
BryceStevenWilley:
```suggestion
```

docs/story_tables.md
```diff
@@ -0,0 +1,187 @@
...
+
+We set the value of a `.there_are_any` row using its `source` row's `value`.  If it's greater than `0`, the value is `'True'`, otherwise it's `'False'`.
+
+When this comes from a `.target_number` row (which I think is currently the only way we get to this stage), we set the value of a `.there_is_another` row using the `.there_is_another` accumulator in the `source` row. Three scenarios are possible.
+
```
BryceStevenWilley: I had a lot of confusion reading this section. I think it's the combination of:

* referring the attributes that aren't on the example objects above (`.source.value` doesn't exist anywhere?) 
* trying to refer to a few different rows in the same sentence, without having them labeled.
* trying to be as detailed as the code itself, which makes the prose really verbose and dense with references to specific things and logic. 

I'd suggest linking to the code, and embedding a lot of these explainations (like the "why `value - 1`?" in the code comments) if you can.

lib/scope.js
```diff
@@ -918,15 +918,19 @@ module.exports = {
       }
 
       let result = {
-        original: row,
+        // Unfortunately mutating frozen objects only fails silently.
```
BryceStevenWilley: do we mind this? Trying to figure out why the comment is there, and what someone is supposed to do when they read it

lib/scope.js
```diff
@@ -2106,35 +2108,126 @@ module.exports = {
...
+    if ( row.source ) {
+      // Increment the number times the source row's artificial row trackers
+      // were used. For example, trackers like those for `.there_are_any` and
+      // `.there_is_another` (for `.target_number` rows)
+      if ( row.source.used_for[ row.var ] ) {
+        row.source.used_for[ row.var ] += 1;
+      } else {
+        row.source.used_for[ row.var ] = 1;
+      }
+      // Increment the number of times the source row has been used in total.
```
```suggestion
```

lib/scope.js
```diff
@@ -2106,35 +2108,126 @@ module.exports = {
...
+    if ( row.used_for[ row.var ] ) {
+      row.used_for[ row.var ] += 1;
+    } else {
+      row.used_for[ row.var ] = 1;
+    }
+
+    if ( row.source ) {
+      // Increment the number times the source row's artificial row trackers
+      // were used. For example, trackers like those for `.there_are_any` and
+      // `.there_is_another` (for `.target_number` rows)
```
```suggestion
      // Increment the number of times the source row's artificial row trackers
      // were used. For example, trackers like those for `.there_are_any` and
      // `.there_is_another` (for `.target_number` rows)
```

lib/scope.js
```diff
@@ -2106,35 +2108,126 @@ module.exports = {
...
-    * rows under special circumstances:
-    * - `.target_number` may indicate the need for a `.there_is_another` row.
-    * - `.there_is_another` in a table is invalid
+    *    row objects under special circumstances:
+    * - `.target_number` will create `.there_are_any` and `.there_is_another` rows.
+    *    If they're unused in the end, that's fine.
+    * - `.there_is_another` with a value of True in a Story Table is invalid and
+    *    will be neutralized.
+    * 
+    * To learn the bigger picture, see the docs folder for info about Story Tables.
```
```suggestion
    * To learn the bigger picture, see "docs/story_tables.md"
```

A link would be nice, but idk how to get that working, either in GitHub or locally

lib/scope.js
```diff
@@ -2149,68 +2242,63 @@ module.exports = {
...
+    // target_number: 3 (3 items. press 'yes' on there_is_another 2 times.)
+    // used_for[ <var_name>.target_number ]: 0 -> `True`, 1 -> `True`, 2 -> `False`
+
+    // Why `desired_num - 1`? The first `there_is_another` is actually the second loop
+    // (after they've already filled in the first list item info). If the user always had
+    // to set `there_are_any`, this wouldn't be a problem, but sometimes the author sets
+    // `there_are_any` beforehand instead so we can't depend on this, so we only count the
+    // use of `there_is_another` and make sure to count the 1st appearance as the second
+    // loop. (or the 0'th appearance as the first loop)
+    } else if ( row.source && row.source.used_for[ row.var ] >= ( desired_num - 1 )) {  // Story table loop will end
```
BryceStevenWilley: After thinking about this, and seeing that you've written a lot trying to explain why we're doing `desired_num - 1`, I think it makes more sense to instead of `used_for + 1`; the number of times that you've already said "there is another" plus the first item is how many items are in the list.

```suggestion
    } else if ( row.source && row.source.used_for[ row.var ] + 1 >= desired_num) {  // Story table loop will end
```

BryceStevenWilley (commented):
Splitting this PR into parts to describe my thoughts on each

* the technical change to fix #706: pretty good! I like the additions, and the changes aren't too bad.
* documentation of the current looping stuff: I think this doesn't really work as intended; just from the ordering of the files in the PR, I read though this before reading over the code changes, and I found myself very confused. But after I read the code itself, it made a lot more sense. If it's helpful for you, we can leave it in, but if you wanted it to help other developers, I don't think it's going to do that, and we should improve it.
* test refactors; I noted my thoughts in the comments, but I'm okay with these, even if I think we could do with better unit testing

lib/steps.js
plocket: There's a 1 min. pause happening elsewhere when there's an error. If we want, we can move it from there to here instead, but we shouldn't have it in both spots, I think. This pause was causing issues where the test would "end", but this would still be going on, so it wasn't really ending.

tests/unit_tests/ensureSpecialRows.test.js
plocket: That's fair. It's updating the fixtures that makes me wonder about their utility, though maybe that's actually a pro as opposed to a con - you get to look over what you're producing very carefully. There's stuff in other fixtures that's out of date now too - the HTML ones for object fields. Ugh.

The unit tests are fast, but the integration tests test all that and more, and we haven't really made a thorough set of unit tests. If we did, it would be more thorough, but harder to upkeep. I'm not sure how to balance that. If we do make a decision, it'd be good to document it if we ever need to re-visit it.

docassemble/ALKilnTests/data/questions/test_loops.yml
plocket: I actually intended to split the ones there into 'end1' and 'end2'. I can change that. I keep getting annoyed when we add pages into a test and have to update all related tests. I was hoping some format like this would allow us to add on to a test without having to update old tests. I wanted to think about it a little more. I can change it back for now if we want.

BryceStevenWilley: > I was hoping some format like this would allow us to add on to a test without having to update old tests

Can you explain a bit more how this would avoid that? I can't picture it, as to reach two different endpoints, you'd have to have some choice affect that, and the only way I can picture that is the test making a specific radio button-like choice of "I'm running test 1". At that point, I wouldn't mind just keeping the different questions in different YAML files, even if it means some duplication on that side.

docs/story_tables.md
plocket: Still have to think about this. Just noting for myself that I also want to add a small additional clarification.

```suggestion
That's no problem at all for our code. If we require our users to use `.target_number`[^1] we have one unique variable to tell us how many times to say "Yes" to a `.there_is_another` question, letting us know when to answer "No" and stop the loop. We just also have to do a bit of extra work to make that happen.
```

docs/story_tables.md
plocket: It also creates `.there_are_any`. I was trying to stay fairly broad with this and get into more details later. I was worried the details would drag this part down and make it too much info. Thoughts?

BryceStevenWilley: It seems like you're trying to be both broad but also include specific details; if you want to stay broad, I wouldn't even mention `.there_is_another`, I'd just say that we add artificial rows based on existing rows. But IMO you could just add `there_are_any` to my suggestion to address your concern, and not try to talk about the concept of artificial rows at all. IMO this was a good level of concrete-ness to have.
```suggestion
    3. If there is a `.target_number` row, use it to create new rows for `.there_is_another` and `.there_are_any`
```

docs/story_tables.md
plocket: Good points to bring up!

> referring the attributes that aren't on the example objects above (.source.value doesn't exist anywhere?)

I reference in [a comment](https://github.com/SuffolkLITLab/ALKiln/pull/709/files/afbc42830fe0ee6f70c045d240f5b9da93a2a8ea#diff-ec6b74c585d7f697b1c6951117041474609979b71be984ce73f5082cb0cb160dR128) the fact that it still has all the old values, but it's obviously not sufficient.

> trying to refer to a few different rows in the same sentence, without having them labeled.

Would love to talk about this a bit more, maybe real time if you're ever less busy. I think maybe there's more to that confusion there. I read those and it looks to me like I do label them, so I'd like to explore what's going on there a bit more.

> * trying to be as detailed as the code itself, which makes the prose really verbose and dense with references to specific things and logic.
> 
> I'd suggest linking to the code, and embedding a lot of these explainations (like the "why value - 1?" in the code comments) if you can.

That's one of the things I was unsure of - how to hit a balance between too much detail and too little detail. I tried for a while to put all this as comments in the code, but felt that I made a hash of it and it was too hard to follow the story from one location to another. The bits of explanation didn't make sense alone. Maybe I can give it another shot, though. Maybe, if we keep this document, I can use expandable content to help with this kind of thing. For example, I might be able to put the code examples in collapsed content (using our friend `<details>`).

lib/scope.js
plocket: I imagine this is for me, or someone like me, ages from now who tries to mutate the object and nothing changes and such a person is confused about why. For a long time, despite being relatively competent with js, I didn't know this fact that seems pretty rarely used and obscure. I imagine tracing my way through the code and finding this clue to help. Certainly just reading `.freeze()` wouldn't have been enough even a short while ago. And what they're supposed to do is stop trying to mutate the object.

If/when we build a class for the table, I have plans to make things fail much less silently if devs try to mutate the object.

Does that all make sense, or am I still off base here? Our opinions may just be different about this. I'm somewhat attached to it, but if you feel strongly I can remove it.

lib/scope.js
plocket: I actually think I'm missing a comment further up. I'm trying to differentiate `.times_used` from `.used_for` and explain why I'm keeping both.

lib/scope.js
plocket: I like the more concise description! I do think it still needs some kind of explanation in a comment, though.

docs/story_tables.md
BryceStevenWilley: > I reference in a comment the fact that it still has all the old values

I see. Yeah, I'd just copy the contents in.

> I read those and it looks to me like I do label them, so I'd like to explore what's going on there a bit more.

I phrased that weirdly; my main point was that reading the code was easier that trying to read prose describe single lines of code.

And I really do think it comes down to that; reading the code was easier. I think a great reason to have internal docs like this is to help "following the story from one location to another", to use your words, and with some slight edits, the numbered list above does that well. But after that, when the text descends to this level of  "That mutates the `'x.there_is_another'` property of the `.used_for` of the `.target_number` object" that I just can't read the sentence and understand what's happening without also looking at the code.

lib/scope.js
BryceStevenWilley: That makes sense. Personally, I would see `freeze()` and realize, "oh, I can't mutate it", but that makes sense, and couldn't still be confusing.

I think the "unfortunately" threw me off. Minor nit, but maybe "NOTE" would communicate the purpose better.

```suggestion
        // NOTE: mutating frozen objects will fail silently, so don't mutate this attribute
```

lib/scope.js
BryceStevenWilley: My issue was that the comment doesn't explain anything, just what it's doing, which reads almost identically to the line of code below it. I get the idea, but TBH I don't think you need to explain why you're keeping both attributes in the code itself. Explanations like this are perfect for the internal docs (and I thought that section about this topic in the docs was good).

lib/steps.js
BryceStevenWilley: Gotcha, I made this comment early and forgot about it by the time I reached the other `waitForTimeout` you modified, sorry!

docs/story_tables.md
plocket: Thanks, there are a lot of good points that we'll continue to discuss. Can you pinpoint a particular reason why the Q&A format is particularly confusing. Some of it can be removed, but I think some of it could be useful - inquisitive/educational, as you described. I'm hoping to bring folks along with me in my thinking, let them know what's coming next, etc. A bit like how I'd do a live presentation.

It also seems otherwise pretty dry, though I'm not sure the Q&A format is the solution to that. If you have suggestions, I'd love to hear them.

docassemble/ALKilnTests/data/questions/test_loops.yml
plocket: Those are also alternatives. I'll outline what I'm thinking of anyway for clarity, though I'm not sure it's any better than the other options.

First what we have now:

First test we create:

```
Given I start...
And I do those first things
Then I reach the end screen
```

When we add the next feature, we change the first test:
```
Given I start...
And I do those first things
And I do the second things
Then I reach the end screen
```

With one test, that's fine, but not when we have tests with multiple paths (though granted I'm not sure how many of those we have, though I could envision others that test different inputs). With the new method:

First few tests we create with branching paths:
```
Given I start...
And I do those first things
Then I reach the 1st end screen
```

When we add the next feature, add a second test:
```
Given I start...
And I do those first things
Then I reach the 1st end screen
And I do the second things
Then I reach the 2nd end screen
```

We then don't have to change the first test to make sure it reaches the end.

Again, I'm not sure if this is the ideal way of handling this situation either. Maybe multiple test files is the way to go. Just clarifying what I meant.

lib/scope.js
plocket: Is the comments something you'd like to work on or should I give it a shot? Or is there a compelling reason to leave out comments there?

docassemble/ALKilnTests/data/questions/test_loops.yml
BryceStevenWilley: Hm, yeah, I don't think that's too useful. Test 2 is now running all of test 1, so what's the point of test 1?

* It might be useful to only be able to look at the title of the test that failed and guess what might have happened, but IMO:
    * that's rarely useful, but you have to dive into the code to see what the unexpected failure was anyway
    * if test 1 fails, test 2 will also fail; there's not a guaranteed order that people will see failures in (sometimes they might read from the top, but often I go to the bottom first, because it lists all of the failures, and I don't have to read through the "false failures" that are indistinguishable from real ones earlier in the logs), so you might see test 2 failed before you see test 1 failed, and wonder what could be going wrong with the 2nd thing (nothing, because it's the first thing). 

The main point of this comment is that `end1` doesn't make sense without an `end2`. IMO even if we still want that ending structure, I'd rather it be named `end_gathered` or something, I guess that's the practical takeaway from this discussion for now.

docs/story_tables.md
BryceStevenWilley: > Some of it can be removed, but I think some of it could be useful - inquisitive/educational, as you described

You're right; reading back through, I like some of it; for example transitions between sections, and the examples are really useful (I am biased towards concrete examples though).  I thought that the parts I commented on didn't give enough educational content compared to their length. 

> I'm hoping to bring folks along with me in my thinking, let them know what's coming next, etc. A bit like how I'd do a live presentation.

I think that's the issue; a written document isn't like a live presentation. We can see what's coming next with a glance to the next section title. It's okay to add filler sentences in a 15 or 30 minute presentation, but the reality is that most people will only spend a short amount of time reading docs like these (unless they are stuck, when they might be frustrated, and more likely to be annoyed at sentences that don't add value or are overly long), so conciseness is still important in writing. 

I do think adding personality is good, and you don't have get rid of all of the Q&A. But there were sentences that didn't provide any information (like questions that made me think "idk, you tell me that's why I'm reading the doc"), and I'd lean towards removing those.

lib/scope.js
BryceStevenWilley: Idk, I left it out because any comment here would just be describing the "what" of the code, which isn't going to
be that useful. I don't think we needed the part about trying to use `.there_are_any`; it was distracting and not relevant to what we are currently doing now.

```suggestion
    // check that the number of elements in the list (i.e. the time's we've asked for `.there_is_another`,
   // plus 1 for the first item before asking) hits our desired number. 
    } else if ( row.source && row.source.used_for[ row.var ] + 1 >= desired_num) {  // Story table loop will end
```

docassemble/ALKilnTests/data/questions/test_loops.yml
plocket: > Test 2 is now running all of test 1, so what's the point of test 1?

As I said, this is in the case where test 1 is actually tests 1, 3, and 4. That is, where there are the same tests that test multiple paths on the same interview.

As I said, I was uncertain about this method anyway. I'll remove the `end1` name.

As a separate issue:

> I go to the bottom first, because it lists all of the failures, and I don't have to read through the "false failures" that are indistinguishable from real ones earlier in the logs

That's a problem I have too. Maybe we should add an internal way to tell the difference between internal false failures and true failures. Does that sound at all useful or will we end up just going to the bottom anyway?

docassemble/ALKilnTests/data/questions/test_loops.yml
```diff
@@ -0,0 +1,66 @@
...
+  - there_are_any_people: DAList.using(object_type=Individual)
+  - there_is_another_people: DAList.using(object_type=Individual, there_are_any=True)
+  - target_people: DAList.using(object_type=Individual, ask_number=True)
+---
+mandatory: True
+code: |
+  there_are_any_people.gather()
+  there_is_another_people.gather()
+  target_people.gather()
+  end1
```
plocket:
```suggestion
  end
```

As per https://github.com/SuffolkLITLab/ALKiln/pull/709#discussion_r1228171965.

docassemble/ALKilnTests/data/questions/test_loops.yml
plocket:
```suggestion
id: end
event: end
question: The end!
```

As per https://github.com/SuffolkLITLab/ALKiln/pull/709#discussion_r1228171965.

docassemble/ALKilnTests/data/sources/story_tables.feature
```diff
@@ -89,3 +89,51 @@ Scenario: I upload files with a table
...
+    | x[i].name.first | TargetPerson2 | target_people[1].name.first |
+  And I SHOULD see the phrase "there_are_any_people people: 2"
+  And I SHOULD see the phrase "there_is_another_people people: 2"
+  And I SHOULD see the phrase "target_people people: 2"
+
+@slow @st7 @loops
+Scenario: target_number 1 for all people lists
+  Given I start the interview at "test_loops.yml"
+  And I take a screenshot
+  And I get to "end1" with this data:
```
plocket:
```suggestion
  And I get to "end" with this data:
    | var | value | trigger |
    | x.target_number | 0 | there_are_any_people.target_number |
    | x.target_number | 1 | there_is_another_people.target_number |
    | x[i].name.first | AnotherPerson1 | there_is_another_people[0].name.first |
    | target_people.target_number | 0 |  |
  And I SHOULD see the phrase "there_are_any_people people: 0"
  And I SHOULD see the phrase "there_is_another_people people: 1"
  And I SHOULD see the phrase "target_people people: 0"

@slow @st6 @loops
Scenario: target_number 2 for there_are_any, there_is_another, and target_number lists
  Given I start the interview at "test_loops.yml"
  And I take a screenshot
  And I get to "end" with this data:
    | var | value | trigger |
    | x.target_number | 2 | there_are_any_people.target_number |
    | x[i].name.first | AnyPerson1 | there_are_any_people[0].name.first |
    | x[i].name.first | AnyPerson2 | there_are_any_people[1].name.first |
    | x.target_number | 2 | there_is_another_people.target_number |
    | x[i].name.first | AnotherPerson1 | there_is_another_people[0].name.first |
    | x[i].name.first | AnotherPerson2 | there_is_another_people[1].name.first |
    | target_people.target_number | 2 |  |
    | x[i].name.first | TargetPerson1 | target_people[0].name.first |
    | x[i].name.first | TargetPerson2 | target_people[1].name.first |
  And I SHOULD see the phrase "there_are_any_people people: 2"
  And I SHOULD see the phrase "there_is_another_people people: 2"
  And I SHOULD see the phrase "target_people people: 2"

@slow @st7 @loops
Scenario: target_number 1 for all people lists
  Given I start the interview at "test_loops.yml"
  And I take a screenshot
  And I get to "end" with this data:
```

As per https://github.com/SuffolkLITLab/ALKiln/pull/709#discussion_r1228171965.

docassemble/ALKilnTests/data/questions/test_loops.yml
BryceStevenWilley: > As I said, this is in the case where test 1 is actually tests 1, 3, and 4

I didn't get that from your description, but my main takeaway was that end1 just wasn't a descriptive name.

> Does that sound at all useful or will we end up just going to the bottom anyway?

I would end up going to the bottom anyway.

docs/story_tables.md
plocket: Thanks for clarifying, I'll think more about that. No pressure, but does this change your mind about some of the suggestions you made? If so, it would help if you edit them to reflect that. Thanks for the feedback.

BryceStevenWilley: It doesn't change my mind about the parts that I edited, but I don't think that there needs to be many more edits past that.

docs/story_tables.md
plocket: I'm going to make some alternative suggestions that I hope still prepare the reader for the next steps of the doc. Let me know how they sit with you.

```suggestion
Third, looping questions that gather multiple items in a list are tricky when the interview answers need the `.there_is_another` attribute. We can't use that attribute. Instead, we handle all gathering with `target_number` rows. Next we'll discuss what problem the `.there_is_another` attribute creates and our approach to solving it.
```

docs/story_tables.md
plocket: Do you think some of this helped follow along with what's going on with the code? That is, support reading the code? That is, is it worth keeping some of it? My concern is that the comments would be so interspersed in different sections of the code that the overall flow may be hard to follow. I'll try to reread it in context when I get back to looking at this more deeply (I think a couple of the other issues are more pressing).

lib/scope.js
plocket: I think the key here is "in total" which differentiates it from the `.used_for` properties. Maybe this is signaling that this comment is in the wrong place. Should it be at the instantiation of the property?

lib/scope.js
plocket: How about the below?

```suggestion
	    // Check that the number of elements in the list hits our desired number.
	    // We have to use `+ 1` to represent that the first loop is handled
	    // by the value of the `.there_are_any` interview attribute instead.
	    } else if ( row.source && row.source.used_for[ row.var ] + 1 >= desired_num) {  // Story table loop will end
```

lib/scope.js
BryceStevenWilley: "In total" to me doesn't add anything to the sentence. "Increment the number of times the source row has been used", and "Increment the number of times the source row has been used in total." mean the same thing, as it's implicit in the first sentence.

You differentiate it from the `.used_for` fine when commenting about `used_for`. IMO `.times_used` is self explanatory, and you already have an identical comment about `row.times_used` at the beginning of the function.

docs/story_tables.md
plocket: I do think it's good to address `.times_used` when it first appears. I wanted to just highlight it in people's minds. Also, that's one of the questions I'd have if I were to see that property. Maybe that's a comment that can be added to the code as well. The below is redundant, but maybe it's better than skipping the property altogether. Thoughts?

```suggestion
It would seem like the `.times_used` property is redundant, but it has a different purpose - it lets us know whether the `target_number` row was used at all during the test to let us give appropriate feedback to the user.
```

BryceStevenWilley (approved):

docs/story_tables.md
plocket: For posterity, I've moved this elsewhere.

docs/story_tables.md
plocket: For posterity: I've made changes that take out the "details" section, moving some of it to the comments in the code and some of it to other places in the new docs.

PR #716, opened 2023-06-16
Update Puppeteer to v20.7.2.

### LINKED ISSUE #660: Update puppeteer, ensure all still working

BryceStevenWilley:
A quick PR to fix #660. 

The only thing not updated is that `headless` will eventually change to a new version; we can use `headless: "new"` to use it.

https://developer.chrome.com/articles/new-headless/

### DIFF
--- a/lib/steps.js
+++ b/lib/steps.js
@@ -180,7 +180,7 @@ Given(/I start the interview at "([^"]+)"(?: in lang "([^"]+)")?/, {timeout: -1}
       // I've seen stuff take an extra moment or two. Shame to have it everywhere
       try {
         // Interview files should get downloaded to downloads folder
-        await scope.page._client.send('Page.setDownloadBehavior', {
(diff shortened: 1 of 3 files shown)


plocket:
[Issue made, #718]

> The only thing not updated is that headless will eventually change to a new version; we can use headless: "new" to use it.

From what I understand (just to clarify what I think you're describing already), we can make that temporary change if we want to try out the new behavior (which would be a great idea). If not, our current code will just default to the new headless when the old one gets removed.

If we want to try out the new version sooner, we should make that an issue so we don't lose track of it. Maybe put a date on it or something.

plocket (approved):
Thanks for trying this out and finding out about headless mode! It passes the tests, and I don't really know what else we can ask for [when updating a dependency like this].

[Issue made, #718] ~[Also we should make an issue for the new 'headless'. Have to run, but if I remember it later and don't see something already up for it, I'll try to create the issue.]~

PR #717, opened 2023-06-24
Close #710, fix ALKilnInThePlayground upload Step

### LINKED ISSUE #710: Fix uploading files not working in ALKilnInThePlayground
Closes https://github.com/SuffolkLITLab/docassemble-ALKilnInThePlayground/issues/18.

ALKiln doesn't know the right address to get the projects upload files when it's running in the playground. See the issue above for details.

plocket:
A few quirks here.

1. This doesn't allow passing in command line arguments for source folder paths yet. That may have to be a separate feature. I'm not sure it should be a required command line argument anyway, in which case we'll still need that code for times when the argument isn't handed in. We could ensure ALKilnInThePlayground hands it in, but I'm not yet sure how we can get ALKilnInThePlayground to get the right paths. Maybe it would have better access to the S3 paths, though.
2. [Added a fix in the in-Playground interview that makes this point moot. See https://github.com/SuffolkLITLab/docassemble-ALKilnInThePlayground/pull/24] ~It doesn't quite work with S3. Jonathan gave us the info that the `/tmp` folder for S3 won't keep those files around, so we probably can't count on those files being there for running these tests in an interview~ (ALKilnInThePlayground - now it occurs to me that ALKilnInterview may have been a better name...)
3. I should have probably broken one change into a different PR and I can do that if this is too much for one PR - The variable setting Step for uploading files turned out to be a mismatch for one ethos we're following. We decided that we want to defer to being radically permissive and forgiving with test code in general. In that spirit, I changed it so that if ALKiln can't find a file that the author listed to upload, the file just isn't uploaded and, if not all variables are required to be set, the test can continue. In a Story Table, the fields are not required to be set and there would be no error just because of a missing file. In a linear Step ("I set the var..."), a missing file would cause an error. I'm not sure about being this permissive, but I've been unsure in similar situations in the past and the collective decision has generally erred on the side of flexibility.

I'll fix the version upon publishing.

Closes #710


plocket:
Not super simple, but could use a look (if you have time), @BryceStevenWilley.

CHANGELOG.md
```diff
@@ -58,6 +58,8 @@ Format:
 - Docassemble Project name prefix now includes ALKiln in it for clarity
 
 ### Fixed
+- DOCKER UPDATE (for ALKilnInThePlayground): Allows ALKilnInThePlayground to upload files (though S3 is not consistent yet). See https://github.com/SuffolkLITLab/docassemble-ALKilnInThePlayground/issues/18. This may require some developers to update their system (updating their docker image) as described in [the documentation](https://suffolklitlab.org/legal-tech-class/docs/practical-guide-docassemble/maintaining-docassemble#updates-to-the-docassemble-container). Updating to a new docker images is something that all docassemble users will have to do sometime anyway, sometimes for security.
```
BryceStevenWilley:
```suggestion
- DOCKER UPDATE (for ALKilnInThePlayground): Allows ALKilnInThePlayground to upload files (though S3 is not consistent yet). See https://github.com/SuffolkLITLab/docassemble-ALKilnInThePlayground/issues/18. This may require some developers to update their system (updating their docker image) if they are below [docassemble-os 1.0.8](https://github.com/jhpyle/docassemble-os/releases/tag/v1.0.8). You can tell if you are below docassemble-os 1.0.8 if you run `docker image inspect jhpyle/docassemble | grep Created` on your host server (Lightsail, etc.) and the created date is before 2023-04-09. You can update as described in [the documentation](https://suffolklitlab.org/legal-tech-class/docs/practical-guide-docassemble/maintaining-docassemble#updates-to-the-docassemble-container). Updating to a new docker images is something that all docassemble users will have to do sometime anyway, sometimes for security.
```

lib/scope.js
```diff
@@ -1762,89 +1777,104 @@ module.exports = {
...
+    let dirs = [];
+    if ( session_vars.get_origin() === `playground` ) {
+      let files_playground = fg.sync(
+        [ `/usr/share/docassemble/files/playgroundsources/${session_vars.get_user_id()}/${session_vars.get_user_project_name()}` ],
+        { onlyFiles: false, suppressErrors: true }
+      );
+
+      // TODO: This dir doesn't hold onto files (according to feedback). We
+      // need to find a better way to get these files from S3. Python lib s4cmd
+      // was recommended.
```
```suggestion

```

lib/scope.js
```diff
@@ -1762,89 +1777,104 @@ module.exports = {
...
 
-    if (pkg_paths.length > 1 && pkg_paths.length - err_paths.length > 1) {
-      await scope.addToReport( scope, { type: `warning`, value: `You have multiple folders in the 'docassemble' folder (${ pkg_paths }). These tests will use the first folder with a "data/sources" subfolder (${chosen_dir}).` });
+    // Make sure sources directory exists and has files in it
+    if ( dirs.length === 0 ) {
+      await scope.addToReport( scope, {
+        type: `warning`,
+        value: `The test files need to be in the "sources" directory, `
+          + `but ALKiln couldn't find the directory or it was empty.`
+          // Should this say the full directory glob used to try to find the files?
```
BryceStevenWilley: Yes we should say the full directory! We can add it to the dev log if we're concerned that it'll confuse folks when seeing it

lib/scope.js
```diff
@@ -1762,89 +1777,104 @@ module.exports = {
...
-      // start of the string. Remove it. If they forgot the space, that's fine too.
-      name = name.replace(/^\s/, '');
-      let file_path = path.join( chosen_dir, name );
+    let missing_files = [];  // For any user warnings
+    for ( let name of uploads_names ) {
+      let trimmed_name = name.trim();
+      let matching_paths = [];
+      if ( dir === null ) {
+        for ( let one_dir of dirs ) {
+          // console.log( one_dir );
```
```suggestion
        for ( let one_dir of dirs ) {
```

lib/scope.js
```diff
@@ -1762,89 +1777,104 @@ module.exports = {
...
+          matching_paths = fg.sync([ `${ one_dir }/${ trimmed_name }` ]);
+          if ( matching_paths.length > 0 ) {
+            // Once we've found a matching file, we choose to use that directory
+            dir = one_dir;
+            break;
+          }
```
BryceStevenWilley: I _think_ we should be able to get rid of this part, i.e. only looking in one directory. As long as there aren't 100's of directories, we should be able to just look through all of them each time, and catch any weird cases where things are in `/tmp/` and not in `/usr/local/...`.

lib/scope.js
BryceStevenWilley: We should also just log all of the dirs in the dev log, regardless of erroring.

BryceStevenWilley (approved):
LGTM, a few small thoughts mentioned in the comments.

PR #721, opened 2023-06-30
#713, test multiple languages

### LINKED ISSUE #360: Languages: Add accessing languages by adding to the url params instead of by tapping a button
e.g. for Spanish, add `?lang=es` to the end of the url. The developer will list the language codes to test.

Will we need some _different_ way for the developer to list the languages to test? Use the same env var, but with different values or with a special step?

note: At request of users.

[Possibly we could have some languages we test by default, but I'm not sure about doing that because it would double, triple, etc. the length of the tests by default, possibly without the dev being aware of it. Also, how would we detect that those languages actually existed for this interview? So that we wouldn't just be testing English repeatedly if the language url just defaulted to English.]

### LINKED ISSUE #713: Restore multi-language tests set by env vars
[Edit: Nothing really change in ALKiln itself, this is just authors using cucumber features that we didn't know about before.]

As per user request, restore running language tests and write internal tests for such (though I'm not sure how we check that the right tests have been run). These should primarily be triggered manually through the workflow dispatch, though allowing an env var for them might still be useful to carry over from previous behavior. Include:

- [x] Access language page url as per #360
- [x] Add language to Scenario tags
- [x] Internal tests of some kind, maybe using tags in Scenario name section. Check that certain text appears multiple times won't be sufficient as errors can also cause that behavior. [Maybe create our own translation file with different text that we can then look for. Use [MADE](https://github.com/GBLS/docassemble-MAEvictionDefense)'s methods as a model for that as that is our primary user need at the moment?]
- [x] ~[If we're creating the language files each time] Ensure InThePlayground deletes the language files during post-test cleanup.~

Should close #360.

Edit:
- [x] Add [new issue to document this in the documentation repo](https://github.com/SuffolkLITLab/docassemble-AssemblyLine-documentation/issues/new)

plocket:
This has minimal invasiveness to just get this working again. It's mostly user initiated. Languages don't get tested by default. If the user wants to test languages, they can use the `@al_language` tag, at which point, alkiln will step aside and let language tests run.

Getting this into circulation can get people started if they need it. If this interface is acceptable for folks to start with (see the new tests), then that's good. I think it'd still be great to have a deeper discussion about what interfaces we want to provide for languages.

Closes #713
Closes #360


docassemble/ALKilnTests/data/sources/languages.feature
```diff
@@ -0,0 +1,52 @@
...
+
+# Usually only run default test and skip or shorten others
+# When commanded, run all non-default language tests
+
+# We'll need to test these manually each time, as our users would,
+# and check that the automatic tests didn't trigger the language
+# tests
+
+# Not sure how to test urls like this: http://es.<other stuff>interview.yml
+# That's how wikipedia does it.
```
BryceStevenWilley: Wikipedia has a very different infrastructure than docassemble; I doubt that if some one is running a multi-lingual interview, they would have different copies of the interview on different servers. which they would need to do with `es.<other stuff>interview.yml`, so I don't think that needs to be noted in the file.

The other way that I would recommend testing or supporting (eventually at least) is the `Accept-Language` header that [docassemble uses if there isn't a `lang` in the URL](https://docassemble.org/docs/language.html#configuration):

> If there is no lang parameter in the URL, docassemble will look for an Accept-Language header in the request from the user’s browser. This is typically set by the user’s web browser to whatever language is defined in the web browser settings. So if the user is a German speaker who is using a web browser that is set up to use the language de, then the language de will be used for translations. In interviews, you can read this language by calling language_from_browser().

docassemble/ALKilnTests/data/questions/test_languages.yml
```diff
@@ -0,0 +1,31 @@
+metadata:
+  title: Language url
+---
+comment: |
+  https://apps-dev.suffolklitlab.org/interview?i=docassemble.playgroundIDproject:lang.yml&new_session=1&lang=en
+  https://apps-dev.suffolklitlab.org/interview?i=docassemble.playgroundIDproject:lang.yml&new_session=1&lang=es
+  https://apps-dev.suffolklitlab.org/interview?i=docassemble.playgroundIDproject:lang.yml&new_session=1&lang=fr
```
BryceStevenWilley:
```suggestion
  https://apps-dev.suffolklitlab.org/interview?i=docassemble.playgroundIDproject:lang.yml&new_session=1&lang=ar
```

Otherwise, I'm not sure what this comment is for, if not just links to test the languages that are translated in this interview

docassemble/ALKilnTests/data/sources/languages.feature
```diff
@@ -0,0 +1,52 @@
...
+#  Examples: english
+#    | lang | arg | text |
+#    | en | &new_session=1&lang=en | English page |
+#
+#  @al_language @es
+#  Examples: spanish
+#    | lang | arg | text |
+#    | es | &new_session=1&lang=es | Español page |
+
+
```
BryceStevenWilley:
```suggestion
```

docassemble/ALKilnTests/data/sources/languages.feature
```diff
@@ -0,0 +1,52 @@
...
+# and check that the automatic tests didn't trigger the language
+# tests
+
+# Not sure how to test urls like this: http://es.<other stuff>interview.yml
+# That's how wikipedia does it.
+
+# Tag use examples:
+# # No tags for default language only
+# @al_language # for all languages
+# @al_language @es # to isolate one language
```
BryceStevenWilley: I think this would be confusing behavior; From my reading of things, `@es` wouldn't work (since `and not @al_language` would be added), and IMO it should just work.

lib/run_cucumber.js
BryceStevenWilley, line 1: I read through the other issues (#713 and #360), and I don't really agree with the idea that we should only run english tests by default, and ignore everything else. Mostly, the idea that we only need to run the English tests by default and not other languages will lead to cases where other language tests aren't run, and eventually break because of that. The argument that things will automatically double in runtime when adding a language doesn't hold here, as with this interface, the tests need to be manually added for each language, for each scenario. Also, as I commented before, it makes for confusing behavior when trying to use the language specific tags.

BryceStevenWilley (commented):
I really like the `Examples` API of running things; it'll need to be documented, but once I understood what was going on, it's really nice, and would make it easy to compare and edit different languages in the interview. I don't think we need to treat `al_language` as special, which is why I didn't approve. Feel free to skip over the necessary approval though if you think this is the right way

docassemble/ALKilnTests/data/questions/test_languages.yml
plocket: I'll remove this.

lib/run_cucumber.js
plocket: To record the user conversation I had, they didn't recall having expressed the desire for the need to only run language tests manually. They then said it seemed like they must not have been that attached to the idea, so I'm fine leaving it out.

I do think the tests will be made longer, especially if they do want to thoroughly test all languages, but if the user isn't concerned about it, I'm not either. I will say I'm surprised, since the user's interview has 5 languages to test, which seems like it would really expand the amount of time it takes, but again, I'm fine if that's at their discretion. I'll post about this in the issue too to outline the current plan.

plocket:
I'm going to go ahead and merge. Anyone can feel free to make an issue or comment if something needs adjusting.

plocket:
Also, nothing really change in ALKiln itself, this is just authors using cucumber features that we didn't know about before. We should edit the CHANGELOG to reflect this.

PR #726, opened 2023-07-14
Enter time input correctly

BryceStevenWilley:
The default text doesn't enter time inputs correct for a few reasons:

* The text input doesn't expected you to type ":", spaces, or the "M" in "AM" or "PM". The space specifically seems to mess up inputs.
* You need to key press "Enter" after entering the time, otherwise moving focus will erase the current input.

This change filters the time value given to remove ":", " ", and "M", and also adds an Enter keypress to make sure the value is entered correctly.

I added a new YML, because the input needs to be required for us to actually test it, and I can't make it required with 5 different tests all trying to use `all_tests.yml` for the same thing, so I split it out instead.

### DIFF
--- a/docassemble/ALKilnTests/data/questions/test_date_and_time.yml
+++ b/docassemble/ALKilnTests/data/questions/test_date_and_time.yml
@@ -1,0 +1,34 @@
+metadata:
+  title: Test testing library - date and time
+  short title: Test ALKiln
+---
+features:
+  css:
+    - styles.css
+---
+# Necessary to tell us what the sought var is on each page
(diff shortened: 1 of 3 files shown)


plocket (approved):
This is great! And I love the new test file and that it passes! We should make an issue to add time to random inputs too as we work on that Step.

plocket:
Reminder to myself to edit the CHANGELOG when publishing.

BryceStevenWilley:
> We should make an issue to add time to random inputs too as we work on that Step.

#729 is made!

PR #733, opened 2023-07-17
Freeze deps for v5

plocket:
Sorry a previous commit for a bump snuck in here. Wasn't sure if there's a reason we were avoiding freezing these.


BryceStevenWilley (approved):
LGTM. This would be to prevent someone from running `npm update` and changing things, and `npm install` will now only use what's in shrinkwrap I believe. Still useful though

PR #735, opened 2023-07-17
BREAKING: Remove EXTRA_LANGUAGES, now using cucumber `Examples`

### LINKED ISSUE #59: Specify the format of EXTRA_LANGUAGES in various places
The format is a comma separated list of words that are in the links or buttons that switch languages. Haven't figured out another way yet.

This might be specified in notes in the code, as indicated, but should also be in the README.

_Originally issue posted by @niharikasingh in https://github.com/plocket/docassemble-cucumber/pull/51#discussion_r532981626_

### LINKED ISSUE #337: Features: Allow dev to limit some tests to only be tested in certain languages
[Not sure exactly how we'd pair this with the `EXTRA_LANGUAGES` env/config variable functionality. Would they end up conflicting? I think the individual test parameters would take over, but the same language test would then run multiple times.

e.g, `EXTRA_LANGUAGES=Español,官话` and `Given I start the interview... in Español` would end up creating two tests because that's how `EXTRA_LANGUAGES` currently works. In the step, though, they'd both be switched to Español. That would mean Español would get run twice, which could be confusing for the developer, especially if the test got an error. Partly because the tags would show the EXTRA_LANGUAGES value.
]

### LINKED ISSUE #444: Bug: An page that in fact doesn't have language buttons even though languages are specified in EXTRA_LANGUAGES does not fail
See https://github.com/plocket/docassemble-CompositeActionsTest/runs/4621804598?check_suite_focus=true

[We can add a test for a failure report for this.]

plocket:
Closes #444, closes #337, closes #59 (EXTRA_LANGUAGES issues). This also removes all the places where languages are set and referenced. It's not completely necessary because the code works just fine when `EXTRA_LANGUAGES` is `undefined`, but it is a bit more cleanup.

[I'm not sure it's absolutely necessary, but it is good to clean up.]


.gitignore
BryceStevenWilley, line 1: This is fine, but IMO we should leave them in, in case you or I have some lang_*.feature files hanging around still

BryceStevenWilley (approved):
LGTM! Thanks for taking everything out, the simpler the code the better

PR #739, opened 2023-07-28
PLAYGROUND_* inputs aren't actually required

BryceStevenWilley:
GitHub will happily still run if you don't give them, (and works just fine: see https://github.com/SuffolkLITLab/docassemble-AssemblyLine/blob/f70ded5c029efd82e8d354d650fa02066422db8f/.github/workflows/run_form_tests.yml#L34-L36 ) but it does give erraneous warnings in VSCode, if you have the GitHub actions extension.

Haven't changed the CHANGELOG, since this doesn't actually change any functionality.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -7,13 +7,13 @@ inputs:
     required: true
   PLAYGROUND_EMAIL:
     description: 'The email of your docassemble testing account. This should be in your GitHub SECRETS or the SECRETS of your org.'
-    required: true
+    required: false
   PLAYGROUND_PASSWORD:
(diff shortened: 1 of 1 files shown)


plocket (approved):
This looks good. These could actually be removed - they were replaced by the API key, and that's when this was actually a breaking change. We can do that another time, though.

BryceStevenWilley:
> This looks good. These could actually be removed - they were replaced by the API key, and that's when this was actually a breaking change

They're removed now! Didn't realize we don't use them at all anymore.

plocket:
Looks good, thanks!

PR #740, opened 2023-07-31
GitHub Server action, to run DA on GitHub

### LINKED ISSUE #756: Run tests in a GitHub-created docker container to isolate tests and decrease chances of spurious failures from server timeouts
Timeouts for things like someone pulling a package with a python module, editing the config, and other such things.

BryceStevenWilley:

In this PR, I have:

### Reason for this PR

Get things running without external issues.

Closes #756

TODO:
* [x] CHANGELOG
* [x] while loop when the server is starting (no 8 minute sleep)
* [x] research how to call ALKiln action from the same branch.


lib/utils/session_vars.js
```diff
@@ -197,6 +198,36 @@ session_vars.save_artifacts_path_name = function ( path_name ) {
   return file;
 };  // Ends save_artifacts_path_name()
 
+
+session_vars.save_install_type = function (install_type) {
```
plocket: From what I can see, this can possibly be taken care of with the `_ORIGIN` env var. Is there a reason it needs to be saved into the runtime config instead that I'm not seeing?

github_server/action.yml
```diff
@@ -0,0 +1,53 @@
...
+  # strange considering we're controlling everything through the action, but
+  # that actually doesn't touch the npm version that's pulled in.
+  ALKILN_VERSION:
+    description: 'Version of ALKiln to get from npm.'
+    required: false
+    default: '^5.0.0'
+  MAX_SECONDS_FOR_SETUP:
+    description: "Amount of time to give the Project to upload your package's GitHub code."
+    required: false
+    default: '120'
```
plocket: Note to self: Remove this default, like we did for the v5 normal action.

lib/utils/session_vars.js
plocket: I see now that `install_type` is different in definition than `_ORIGIN` conceptually. `install_type` is if it was installed in the playground vs. on the server. That is a more clear description. I'm wondering if it's still enough to just infer it from `_ORIGIN` or if we really want to introduce a new action step-crossing config variable. I'll restore the previous code, but I think it's worth discussing.

github_server/action.yml
```diff
@@ -0,0 +1,53 @@
+name: "ALKiln/github_server: Isolated docassemble tests"
+description: "Automatically test docassemble interviews on GitHub's servers, in an isolated environment. Avoids docassemble reload errors and other external factors"
+
+
+inputs:
```
plocket: TODO: Do the regular ALKiln `inputs` of the package authors need to be handled in any special way to pass them on to the default ALKiln action? [This should become obvious if we test it with a package that needs some inputs.]

lib/utils/session_vars.js
BryceStevenWilley: 1) I think of `_ORIGIN` and `install_type` as separate things: `_ORIGIN` should only deal with where Kiln JS is running, and `install_type` is independent of that. I can easily see GitHub actions or local runs using either installed interviews or playground interviews. I agree it's less likely to run KilnInThePlayground on an installed interview, but...

2) I still hold that `_ORIGIN` is a temporary solution, and the problem it solves is better and more flexibly solved by command args for parameters, so I've been avoiding building more functionality on top of it.

github_server/action.yml
BryceStevenWilley: Secrets should _not_ be affected, as they're all read from env vars, but we do need to manually pass these args to the `uses: SuffolkLITLab/ALKiln` step below

plocket:
Note: need to know what different errors look like in this process and document them (like missing env vars, missing config stuff, missing package dependencies, etc.)

[Conclusion: For missing config/packages/general docker container issues, we'll just need to tell people that they'll have to look at the docker log or use tmate in order to troubleshoot the problem. Made https://github.com/SuffolkLITLab/ALKiln/issues/817]

github_server/action.yml
```diff
@@ -0,0 +1,53 @@
...
+        shell: bash
+      - name: Download and start docker
+        run: |
+          mkdir /tmp/config
+          echo "$CONFIG_CONTENTS" > /tmp/config/myconfig.yml 
+          docker pull brycewilley/smoldocassemble
+          docker run --env DA_ADMIN_EMAIL="admin@example.com" --env DA_ADMIN_PASSWORD="password" --env DA_ADMIN_API_KEY="${{ env.DA_ADMIN_API_KEY }}" --env DA_CONFIG=/tmp/config/myconfig.yml --volume /tmp/config:/tmp/config --cap-add SYS_PTRACE --memory="4gb" -d -p 80:80 jhpyle/docassemble
+          # TODO before merge: do a while loop here until `localhost` shows "It's ready"
+          sleep 480
+          curl localhost
```
BryceStevenWilley: A thing that should change before this gets merged; we shouldn't need a huge sleep. It should do some loop with a few seconds of sleep (3-5 probably), and then `curl localhost` to see if we get served the "It's ready" page, or the "docassemble is loading...".

github_server/action.yml
```diff
@@ -0,0 +1,53 @@
...
+        # TODO: Research how to get this to always be the same branch we're on
+        uses: SuffolkLITLab/ALKiln@v5
+        with:
+          SERVER_URL: "http://localhost"
+          DOCASSEMBLE_DEVELOPER_API_KEY: "${{ env.DA_ADMIN_API_KEY }}"
+      - name: Docker logs
+        # TODO: create an arifact for this instead
+        run: |
+          docker logs $(docker ps --filter "ancestor=jhpyle/docassemble" --format {{.ID}})
+        shell: bash
```
BryceStevenWilley: Another refactor: the logs could have a few issues:

* this won't print if the ALKiln tests or the previous step fails for any reason. it should print in those cases too
* there is the chance that it prints secrets from the config file. The way that we set the config file is through the entire thing, which means I think that GitHub will only censor things if it prints the config in it's entirety, and won't censor specific lines. Some ways around that include:
  * making the config file a mako template or something similar, and replacing the actual secret values with specific github secrets. That would mean that github would still censor the actual secret parts in case they're printed in the log.
  * seeing if GitHub lets us restrict who can download artifact files, and making this a downloadable file instead of printing straight to the GH log?

lib/docassemble/get_base_interview_url.js
```diff
@@ -11,9 +11,18 @@ let get_base_interview_url = async function () {
...
-  return `${ server_url }/interview?new_session=1&i=docassemble.playground${ dev_id }${ project_name }%3A`;
+  console.log(session_vars.get_install_type());
+  if (session_vars.get_install_type() === "server") {
+    let x = `${ server_url }/interview?new_session=1&i=${session_vars.get_package_name().replace("-", ".")}%3A`;
+    console.log(`base_interview_url: ${x}`)
+    return x;
+  } else {
+    let dev_id = await docassemble_api_interface.get_dev_id();
+    let project_name = session_vars.get_project_name();
+    let x = `${ server_url }/interview?new_session=1&i=docassemble.playground${ dev_id }${ project_name }%3A`;
```
BryceStevenWilley: Discussed on deep dive, but another refactor I would consider if there should be a different way of passing along this URL. Not sure the best way though.

lib/docassemble/startserver.js
```diff
@@ -0,0 +1,69 @@
...
+  console.log(`submitted`);
+  const api_text = await page.$eval('body p code', elem => elem.innerText);
+  process.env.DOCASSEMBLE_DEVELOPER_API_KEY = api_text;
+  console.log(`api key: ${api_text}`);
+  let file = fs.writeFileSync( ".tmpenv", `export DOCASSEMBLE_DEVELOPER_API_KEY=${api_text}`);
+
+  await page.goto(`${ session_vars.get_da_server_url() }/user/sign-out`, {waitUntil: `domcontentloaded`});  
+  console.log(`signed out`);
+  await browser.close();
+  */
```
BryceStevenWilley: IDK if any of the above is useful, but I did have a version that could make an API key for the user, but it still needs a user name and password.

tmp.bash
BryceStevenWilley, line 1: can remove this file, I don't remember what it existed for

lib/docassemble/startserver.js
BryceStevenWilley, line 1: worth thinking of a different name for this file; startserver isn't great, likely `setup-install` or something closer to the other setup

lib/utils/session_vars.js
plocket: Deep dive discussion: Since the two vars could be used in combination someday, we'll keep them separate for now. We'll also add the `install_type` interview url to the arguments that will be passed in through the command line when we do the same for the file names (avoiding `_ORIGIN`, though we do have to check if that env var is used for something else too), so for now we'll keep things as they are since they'll be changed anyway.

github_server/action.yml
plocket: Can confirm. [For future discussion:] This duplicates the work we do in ALKiln. I wonder if there's another way to do this. Also, I'm concerned about the custom env vars that users are setting - will they come through. I suspect they'll be fine, but we'll be able to test that later. Given that, would it be better to have authors set the ALKiln as env vars instead of passing them as inputs. It would be a breaking change, but it would save a lot of duplication and maintenance cost.

lib/docassemble/get_base_interview_url.js
plocket: Thoughts: We're always going to be the ones deriving the URL. I think we were just describing separating the concerns. We could make a script that just determines the url of an interview installed on the server and a script that determines the url of a playground interview. Each action could use the appropriate one and output the result to the action, which would then use it in the command line arguments.

lib/docassemble/startserver.js
plocket: If we do want to create unique API keys and we do want to keep them secret, we'd need to work with the GitHub API to create those secrets.

Stream of consciousness: Not sure why we need a custom API key - the server gets killed after the tests. Are we concerned people will interfere with the process in the middle of the action? I think only people with permissions could use something like tmate with a test run. It seems like that could be a security measure the authors have to take - only allow trusted folks to be collaborators. Maybe that's too restrictive? If so, does GitHub have a way to have finer-grained permissions - allow only admins to mess with runningtests? Even that might be too restrictive since collaborators do need to look at tests, though I guess they don't need to look at running tests...

lib/docassemble/startserver.js
plocket: Maybe `server_install`.

plocket:
More notes in #753 (the PR made from that branch). ~I'll put the notes of my action experiments on https://github.com/SuffolkLITLab/ALKiln/tree/start_server_workflow in here so everything's in the same place.~

The biggest issue I see as I work through this is that we run the github_server setup to get the interview on the server, but then we also run the regular setup and takedown, which will install the repo on the Playground. Do we want to double up that way? Is there a point to doing both? Where do we want to control whether those happen or not?

Testing bugs so far: In an [08/14 test run](https://github.com/SuffolkLITLab/ALKiln/actions/runs/5863230862/job/15896328024):
- A wait of 700 still [didn't give the server enough time to start](https://github.com/SuffolkLITLab/ALKiln/actions/runs/5863230862/job/15896328024#step:3:164) (the curl happens after the sleep, so it should be showing that the server has already started, which it isn't). Bryce mentioned that we may not be able to look at what's going wrong very easily, so we'll have to figure out the next steps.
- As suspected, `SERVER_URL` and friends were not defined in the ALKiln action itself: https://github.com/SuffolkLITLab/ALKiln/actions/runs/5863230862/job/15896328024#step:3:198. (at alkiln-startserver, where `session_vars` is first used).
- `REPO_URL` also wasn't defined, which makes sense, but means we need to duplicate some of the work we're doing in the regular `action.yml`. It is needed in `da_i.install_on_server`. I'm not loving the amount of duplication we're having to do.

Thoughts:

Maybe we need to somehow break out into further actions to avoid duplication - an action for accepting input and creating env vars that happens before the other actions. It, or something else, would then need to manage calling the right follow-up actions. If we do that, the github server action can avoid the regular `action.yml` setup and takedown by just calling `test` ourselves.

This seems complicated and would require conditional logic with another env var or input, which we were trying to avoid by splitting these actions into separate scripts (as we discussed). We should discuss which is worse - the duplication or the additional input/env var. (I'm reluctant to use another input as we're running into duplication problems with our current ones.)

lib/docassemble/startserver.js
BryceStevenWilley: Whatever API key we make, we shouldn't keep around, so don't need to worry about that.

And I only brought this up in case it might be useful for other Kiln features; it would not be useful for running docassemble on docker on GitHub.

BryceStevenWilley:
Seeing the state of things and the fact that we do still need to export / duplicate the env vars in both `actions.yml`s for things to be fully separate , I think the extra input to `ALKiln/action.yml` isn't that bad of an option; it can be optional, so it won't change any one's existing usage, and can default to the normal usage.

> we're running into duplication problems with our current ones

If you're talking about v5 vs v4 duplication issues, this is v5 only.

github_server/action.yml
```diff
@@ -0,0 +1,53 @@
...
+  steps:
+      - name: "github_server: Set environment variables"
+        run: |
+          echo "DA_ADMIN_API_KEY=abcd1234abcd1234abcd5678abdc5678" >> $GITHUB_ENV
+        shell: bash
+      - name: Download and start docker
+        run: |
+          mkdir /tmp/config
+          echo "$CONFIG_CONTENTS" > /tmp/config/myconfig.yml 
+          docker pull brycewilley/smoldocassemble
```
BryceStevenWilley:
```suggestion
          docker pull jhpyle/docassemble
```

This will avoid an extra 1-2 minutes of downloading a docker image that we aren't using anymore (though it's worth considering trying that again if it keeps taking that long for everything to start up).

github_server/action.yml
plocket: Possible argument for keeping inputs: It makes those vars distinct from other vars - they clearly belong to ALKiln and we are able to make some required. Maybe they're also (only a little) easier to look up in the documentation. I'm not as sure about the ones that aren't required.

github_server/action.yml
plocket: I've at least confirmed that a user who isn't logged in can't download the artifacts or see the details of the console log. I assume you mean restricting on a more fine-grained basis. E.g. only allow admins to see those.

plocket: This is what I'm told about config logs:

> The configuration isn't echoed to output, but error messages might contain references to configuration values.

plocket: Another alternative (not that I'm a huge fan) is to encrypt the output file. It's a lot more work for the user, but so far I haven't found a way to restrict permissions.

As for Mako + secrets, just to state it explicitly, that's up to the user to do in their own repository. I'm not sure what those files would look like and how you'd run those scripts. That, too, would be a lot of work for the user. Maybe more work than an encrypted file.

lib/docassemble/get_base_interview_url.js
plocket: We'll have to have a real-time discussion to refresh my memory on this. Right now my confused brain is wondering why we need to create the whole url ourselves. For example, couldn't we send the base url of the interview through the command line [- for the server `serverurl/interview?i=`, playground same plus `docasemble.playground`]. Maybe that's still us - ALKilnInThePlayground is us - but there's still a distinction, right? Allow that repo to send the base url through the command line? Is that the flexibility we're looking for? So that we, as developers [running the command manually], can pass in the base url of the interview through the command line?

Maybe we should write some user stories to help clarify the exact paths we're thinking of.

lib/docassemble/startserver.js
plocket: It's true it's another way to make an api key. I wish there were a place to store code like this without keeping it in the repo itself. I'll never find it in here. Maybe we can make issues for this kind of stuff with a special label or something.

lib/utils/session_vars.js
plocket: Conclusion: We used `install_method` and kept `_ORIGIN` separate.

lib/docassemble/get_base_interview_url.js
plocket: Conclusion: There were only 2 url types - server install and playground install, so we'll handle it by using an npm script to install a runtime config var we can use as a flag.

lib/docassemble/startserver.js
plocket: Conclusion: Made and closed https://github.com/SuffolkLITLab/ALKiln/issues/815

github_server/action.yml
plocket: Conclusion: (not about inputs in general) We made a separate workflow script that authors can make to use the server action. In their workflow, the set up the server with one action and then run ALKiln with the other action, with special arguments.

plocket:
Docassemble core is giving the wrong error [and that's causing our tests to fail]. This is blocked till the next release when that should be fixed.

[Fixed!]

plocket (approved):
I'm a bit too close to it, but from what I can tell we're good to put it out there.

PR #741, opened 2023-07-31
Remove MAX_SECONDS_FOR_SETUP default in action.yml

BryceStevenWilley:
The default value is already present in the JS code, so it shouldn't be in two individual places.

Not listed in the changelog, since it doesn't change any behavior.


plocket (approved):
Great change, thanks! I do wonder if we should list these things in the changelog anyway. We may think they don't affect things and then find out that they did more than we thought.

Issue #687, opened 2023-04-06
Pass feature/sources tests path through the command line


plocket:
This is fine for now, but I didn't see an issue anywhere, so I wanted to note; I think the next refactor here should be do add a proper command line argument handler like [minimist](https://github.com/minimistjs/minimist), and then pass these two paths through the command line instead. It has a few benefits:
* ~removes the necessity of 2 additional environment variables as well, `get_user_id`, and `get_user_project_name`~
* puts docassemble specific knowledge (what these specific paths are) in the docassemble specific package (ALKilnInThePlayground)
* makes it more flexible in general; I made a quick little tool that could run 10 different versions of Kiln at the same time by running them in different working directories and pointing them to the source directory elsewhere. 

I did a little preliminary work on this in https://github.com/SuffolkLITLab/ALKiln/commit/f8e9921788abc3c45748a8090e451275081145e9, but I'm trying to shift priorities for a bit, and don't want this to be slowed down by me.

_Originally posted by @BryceStevenWilley in https://github.com/SuffolkLITLab/ALKiln/pull/686#discussion_r1160031854_

In response: Great points!

Edit:
~We may add to this the base url of the interview. The urls are different depending on whether the package was installed in a server's Playground, or on the server itself. This base url might also be passed into the command line instead of `install_type` config/env var. See comments/discussions about `install_type` in https://github.com/SuffolkLITLab/ALKiln/pull/753 and https://github.com/SuffolkLITLab/ALKiln/pull/740.~

Also, if we don't need `get_user_id` for checking whether the server is up, we might use `server_url.com/health_check?ready=1`. Still have to check the validity of the API key somehow.

plocket:
We might make this the `source` path and add `*.feature` ourselves later. Helpful for getting to the upload files too.

plocket:
Does this need to be done before v5 is released? Are we going to make this a breaking change? [i.e. will this be an optional argument?]

Update: No, this argument will be optional and we'll have default behavior, so this doesn't need to be part of this milestone.

plocket:
What do we envision as some examples of using the command line in this way. We're passing in:

- Tags
- The location of the feature files
- ~The base url of the interviews of the project~

Tags currently: `npm run test @tag1`. Will the other ~two~ arguments have flags?

[...]

PR #753, opened 2023-08-21
Implement GitHub server workflow and action

### LINKED ISSUE #756: Run tests in a GitHub-created docker container to isolate tests and decrease chances of spurious failures from server timeouts

plocket:

It looks intimidating, but most changes are a dir name change that we may be able to change back.

In this PR, I have:

* [ ] Added my changes to the CHANGELOG.md at the top, under the "Unreleased" section

### Reason for this PR

Getting GitHub server workflow and action to run and pass tests. Wanted this to be a separate PR so we could review changes and possibly decide to make a new branch with cleaner commits since this needed a lot of tiny changes for experimentation.

We probably need a section in the PR template for what more specific things we've changed that may not go in the changelog. Things that may need discussing (with numbers for ease of reference):
1. ✔ [Conclusion: remove code, but make an issue/document the link to the commits. Maybe we add it and remove it. May not notice dangerous screenshots for a few weeks. Suggestion: Can we just change the session vars function to always return `null` so we don't have to remove the infrastructure. Response: The infrastructure is the concern. An `if` could get mistyped in a future change. Question: Can we provide a link to the code we're going to remove so it's easier to redo. Response: yes, or make a branch that always has this, but the commit might be better. Response: somewhat defeats the purpose of secret stuff, or at least makes them less safe.] Added _ALKILN_DANGEROUS_DEBUG for taking pictures of passwords and getting values of secret variables as it became very difficult to debug our tests without that.
2. ✔ [Conclusion: if for someone specific (plocket), then that sounds fine. `action_github_server`] We still need to decide on the name of the folder that contains the action. I kind of want to design this with potential future actions in mind. Maybe some kind of naming convention. 'action' at the end of the name makes english sense, but making putting it at the start of the name would make sure the actions would be proximal to each other, and I like that idea. Other ideas?
3. ✔ [Conclusion: Keep it as it is. Response: don't love that format, but it does allow people to work with their docker container before running the tests themselves. Those outputs are good.] Added outputs from the docker run so devs don't have to guess what values to use for the variables we're creating, like API key, email, and password. One of them is the server url. I'm not sure about that one. It does avoid a hard-coded string, though, that may confuse some people otherwise. Is it better to have a complex, but consistent, way of giving and getting info, or would it be better to have a recognizable hard-coded string that would be slightly different for the developer?
4. ✔ [Conclusion: `install_method` ~location may be too confusing. Keep 'type'.~] Should we use the name INSTALL_TYPE or INSTALL_LOCATION? I like the second, but am afraid it could be confusing when we start using command-line arguments. That said, I can see the values being 'server', 'playground', _or_ some url path that we can pass to the command line. Maybe that's trying to do too much. Maybe that last one could be a totally separate `BASE_INTERVIEW_URL` at some future point. Maybe `WHERE_INSTALL`.
5. **X** [Conclusion: ~Make an issue.~ We haven't run into this again, so we may wait to make an issue if it comes up again. Note: Might be supervisord.] Ran into some frustrating possibly timeout/race condition bugs that forced me to re-run github server actions and makes me question some of the utility of this (though it would still help with longer test runs).
6. ✔ [different issue] Do we need to prefix all our `echo`s with "ALKiln"? Maybe we do, because even I got confused sometimes.
7. ✔ [Change it back] Changed ALKilnTests name to ALKiln. It turns out it probably wasn't necessary...? Probably?
8. ✔ [Conclusion: false positive (GitHub may, fortunately or unfortunately, remember this choice we made)] Git Guardian didn't like me hard-coding the api key. Should we mark it a false positive, or is it going to "learn" from that and possibly miss future security issues? We could otherwise just skip it without marking it as a false positive.

Other things:
- Now installs the package on the server with `dainstall` in the CI instead of in js
- Made a new artifact file
- Required AssemblyLine properly
- Called ALKiln action from the same branch

Close #756

Lots of GitHub workflow runs that can be seen in 'Actions'. Includes testing inputs to and outputs of the new action.


BryceStevenWilley:
> Changed ALKilnTests name to ALKiln. It turns out it probably wasn't necessary...? Probably?

Definitely not necessary, I do suggest that you change it back.

> Git Guardian didn't like me hard-coding the api key. 

I couldn't mark it as a false positive / test credential until this morning, but just did. It shouldn't show up again.


I'll have time for an actual review later, but wanted to mention those two things first, which should unblock you for a bit

github_server/action.yml
```diff
@@ -1,53 +1,143 @@
...
-          # TODO before merge: do a while loop here until `localhost` shows "It's ready"
-          sleep 480
-          curl localhost
-        shell: bash
-      - name: Run Kiln
-        # TODO: Research how to get this to always be the same branch we're on
-        uses: SuffolkLITLab/ALKiln@v5
+          echo "${{ inputs.CONFIG_CONTENTS }}" > /tmp/config/myconfig.yml
+          docker pull jhpyle/docassemble
+          docker run --env DA_ADMIN_EMAIL="${{ env.DA_ADMIN_EMAIL }}" --env DA_ADMIN_PASSWORD="${{ env.DA_ADMIN_PASSWORD }}" --env DA_ADMIN_API_KEY="${{ env.DA_ADMIN_API_KEY }}" --env DA_CONFIG=/tmp/config/myconfig.yml --volume /tmp/config:/tmp/config --cap-add SYS_PTRACE --memory="4gb" -d -p 80:80 jhpyle/docassemble
```
BryceStevenWilley:
```suggestion
          docker run --env DA_ADMIN_EMAIL="${{ env.DA_ADMIN_EMAIL }}" --env DA_ADMIN_PASSWORD="${{ env.DA_ADMIN_PASSWORD }}" --env DA_ADMIN_API_KEY="${{ env.DA_ADMIN_API_KEY }}" --env DA_CONFIG=/tmp/config/myconfig.yml --volume /tmp/config:/tmp/config --cap-add SYS_PTRACE --memory="4gb" -d -p 127.0.0.1:80:80 --name docassemble_container jhpyle/docassemble
```

This will make sure that the docassemble server is only visible from the GitHub action machine, a bit more secure.

plocket: Note: We set `--memory` to give a more understandable error if docker runs out of memory, but we may in future need to add an input for that value if someone has a problem.

github_server/action.yml
```diff
@@ -1,53 +1,143 @@
...
+# Bug? Race condition? Exit with 52 when waiting for container to finish starting
+# https://github.com/SuffolkLITLab/ALKiln/actions/runs/5912900256/job/16036751264#step:4:188
+# Had no completion messages, but it exited with an error anyway
+# Took 121 seconds
+# The rerun of the test completed that step just fine, though it took less time
+# Took 120 and was fine: https://github.com/SuffolkLITLab/ALKiln/actions/runs/5913167036/job/16037288791
+# 150 secs with exit code 56:
+# https://github.com/SuffolkLITLab/ALKiln/actions/runs/5914677262/job/16040233419#step:4:196
+# 120 sec with exit code 56:
+# https://github.com/SuffolkLITLab/ALKiln/actions/runs/5914963614/job/16040773624#step:4:206
```
BryceStevenWilley: Looked into this a bit more and I was incorrect on our call; exit code [56 comes from `curl`](https://everything.curl.dev/usingcurl/returns), and is a "Failure when receiving data from the peer". Error 52 is apparently [empty reply from the server](https://stackoverflow.com/a/42837622/11416267). Both of these are probably expected, because DA is starting servers and everything, so if it starts the nginx server in the middle of our call, `curl` would get this error.

The confusing thing here was that by default, GitHub runs `bash` steps as `shell: /usr/bin/bash --noprofile --norc -e -o pipefail {0}`. `-e` means that if [any command in the bash script fails, then the entire step fails](https://stackoverflow.com/a/9952249/11416267). Since `curl` returns lots of non-zero exit codes, we don't want the step to fail anytime it happens, we just want to ignore it (if it fails, we will still print out the curl response in the logs for debugging, so that should be fine).

GitHub's [suggested way of avoiding these extra args to bash](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#exit-codes-and-error-action-preference) is to use `shell: bash {0}`, which I recommend we do in this case.

lib/docassemble/docassemble_api_interface.js
```diff
@@ -329,29 +329,6 @@ da_i.pull = async function ( pull_options ) {
   }
 };  // Ends da_i.pull()
 
-
-da_i.install_on_server = async function ( install_options ) {
```
BryceStevenWilley, line 333: I think it's still useful to have these APIs as a part of ALKiln. I know we aren't using them directly in this PR, but at some point you had asked me how to make things more portable so others could run it, and this is how; the only reason that we switched to `dainstall` is because DA has a bug with weird workarounds for handling packages where the repo name is different than the package contents.

.github/workflows/github_server.yml
```diff
@@ -0,0 +1,83 @@
...
+      # Place the root directory in this branch
+      - uses: actions/checkout@v3
+
+      # Debugging. Maybe add a long sleep after whatever it is we want to check
+      # to give us time to explore.
+      - name: Setup tmate session
+        if: ${{ inputs.enable_tmate }}
+        uses: mxschmitt/action-tmate@v3
+        with:
+          detached: true
```
BryceStevenWilley: Is this something that we need to leave in?

.github/workflows/github_server.yml
```diff
@@ -0,0 +1,83 @@
+name: GitHub server tests
+
+on:
+  push:
+  # why is workflow dispatch not showing up in github?
+  workflow_dispatch:
+    inputs:
+      tags:
```
BryceStevenWilley, line 7: This passes all the way through to the `ALKiln/action.yml`? Interesting, not sure how I feel about that though. Now there are 3 different mechanisms to pass inputs to ALKiln; this, inputs, and env vars. Doesn't look like we can easy refactor though.

.github/workflows/github_server.yml
```diff
@@ -0,0 +1,83 @@
...
+      #    echo "container_name is $container_name"
+
+      - name: "Run ALKiln tests"
+        if: ${{ success() }}
+        uses: ./
+        with:
+          SERVER_URL: "${{ steps.github_server.outputs.SERVER_URL }}"
+          DOCASSEMBLE_DEVELOPER_API_KEY: "${{ steps.github_server.outputs.DOCASSEMBLE_DEVELOPER_API_KEY }}"
+          INSTALL_METHOD: "server"
+          # TODO: See what happens if this is an empty string
```
BryceStevenWilley: It should default to `^5.0.0`, the default value of `ALKILN_VERSION` in `/action.yml`. Do we need the TODO?

action.yml
```diff
@@ -23,6 +23,12 @@ inputs:
     description: "Max amount of time to give the server to reload if it needs to"
     required: false
     # no default to let the custom Scenario timeouts create the reload timeout
+  # INSTALL_LOCATION? Will that get confused with the location passed into the command line?
+  # Maybe `INSTALL_LOCATION` could be "playground", "server", or a custom path and we could handle all those.
+  INSTALL_METHOD:
```
BryceStevenWilley, line 27: `INSTALL_METHOD` was an idea that didn't come up when we were discussing today; I like it! I'd be fine with that or `INSTALL_TYPE`.

lib/docassemble/server_install.js
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env node
+
+const session_vars = require('../utils/session_vars');
+
+const server_install = async () => {
+  session_vars.save_install_type("server");
+};  // Ends server_install();
```
BryceStevenWilley: This script is even weirder now that it doesn't do anything besides setting the install type for the rest of Kiln to use. Like, you can't look at the script and understand what's happening at all.

I'm not sure what's more hacky; leaving this script and the `session_vars.save_install_type` APIs, or just echoing a `{ "da_install_type": "server" }` to `runtime_config.json` directly in the GitHub action. But I'm not happy with the state of these APIs now. Now that `dainstall` is doing all of the work, maybe it is better to pass `install_type` through on env var, idk.

github_server/action.yml
```diff
@@ -1,53 +1,143 @@
...
+      - run: dainstall . --apiurl http://localhost --apikey $DA_ADMIN_API_KEY
+        shell: bash
+
+      # Upload artifacts that subscribers can download on the Actions summary page
+      - name: "ALKiln github_server: Docker logs"
+        if: ${{ always() }}
+        run: |
+          docker_logs=$(docker logs $(docker ps --filter "ancestor=jhpyle/docassemble" --format {{.ID}}))
+          # Should this name have a date/time attached? The logs themselves have timestamps
+          echo "$docker_logs" > /tmp/alkiln_docker_logs.txt
```
BryceStevenWilley:
```suggestion
          # Should this name have a date/time attached? The logs themselves have timestamps
          echo "$(docker logs docassemble_container)" > /tmp/alkiln_docker_logs.txt
```
We can set the `--name` of the docassemble container earlier, which simplifies this command a lot

github_server/action.yml
```diff
@@ -1,53 +1,143 @@
...
+        shell: bash
+    
+      # Should this date/time be created at the start of this action?
+      - name: "ALKiln github_server: Get folder name with date and time"
+        if: ${{ always() }}
```
BryceStevenWilley: Is `if: ${{ always() }}` necessary? IMO it's easier to understand without it.

github_server/action.yml
```diff
@@ -1,53 +1,143 @@
...
+          SECONDS=0
+          # Wait 1 minute before pinging the server constantly
+          sleep 60
+
+          # Use a variable instead of checking $response so we can control to avoid race
+          # conditions from preventing the success log message.
+          ready="false"
+
+          while [[ $ready == "false" ]]; do
+            echo "Time ellapsed: $SECONDS seconds"
```
BryceStevenWilley:
```suggestion
            echo "Time elapsed: $SECONDS seconds"
```

github_server/action.yml
```diff
@@ -1,53 +1,143 @@
...
+          # ALKiln github_server: Wait for server to start
+
+          # Reset the SECONDS variable
+          SECONDS=0
+          # Wait 1 minute before pinging the server constantly
+          sleep 60
+
+          # Use a variable instead of checking $response so we can control to avoid race
+          # conditions from preventing the success log message.
+          ready="false"
```
BryceStevenWilley: I'm okay with this variable, but I don't think the comment is correct; I don't think this prevents any race conditions, does it? It lets us have a `while` loop instead of a `do-while`.

github_server/action.yml
```diff
@@ -1,53 +1,143 @@
...
+
+          while [[ $ready == "false" ]]; do
+            echo "Time ellapsed: $SECONDS seconds"
+
+            # Check if max seconds till loading has passed
+            if [[ $SECONDS -ge $MAX_SECONDS_FOR_DOCKER ]]; then
+              echo "ALKiln GitHub Server Error: Timed out waiting for docker container to serve the site. If you want to give your docker container more timem to install, set the MAX_SECONDS_FOR_DOCKER input for this action."
+              # Exit with failure, or allow this to keep going?
+              break
```
BryceStevenWilley: IMO exiting here is a good idea.

```suggestion
              exit 1
```

github_server/action.yml
```diff
@@ -1,53 +1,143 @@
...
+      # Upload artifacts that subscribers can download on the Actions summary page
+      - name: "ALKiln github_server: Docker logs"
+        if: ${{ always() }}
+        run: |
+          docker_logs=$(docker logs $(docker ps --filter "ancestor=jhpyle/docassemble" --format {{.ID}}))
+          # Should this name have a date/time attached? The logs themselves have timestamps
+          echo "$docker_logs" > /tmp/alkiln_docker_logs.txt
+        shell: bash
+    
+      # Should this date/time be created at the start of this action?
```
BryceStevenWilley: IMO it's not worth it.

BryceStevenWilley (commented):
A quick review of this

github_server/action.yml
plocket: If we give it `--name docassemble_container`, would that be a useful output so devs don't go have to go looking for the process name? Wish we could get it's id too.

.github/workflows/github_server.yml
plocket: Need is a strong word, but if we leave it in it's easy to trigger it with a manual workflow to debug GitHub server type stuff. Maybe this another thing where we can make an issue and point to this code, though.

.github/workflows/github_server.yml
plocket: This is so manual workflows can get flexible arguments, like just run one or two tests. It means folks don't have to edit their files just to run a test for a tag or two.

plocket: In this group of arguments, we could most easily remove the need for `with:` inputs by encouraging people to pass in those values as env vars. I do see reasons to at least keep required inputs as inputs, though.

.github/workflows/github_server.yml
plocket: We don't need to keep the todo if we're sure of that. I wanted to test it out. I was unclear if an empty string in bash is the same as undefined.

lib/docassemble/server_install.js
plocket: We can't just `echo` because there's other stuff that needs to be in the `.json`. I know what you mean, though. I don't see a great solution either, though I do admit I like `dainstall`. Just one line, so much simpler.

Since `install_type` is going to be temporary and eventually taken over by passing a base url in for the command line, how much does it matter? Partly depends on how long we keep this stuff around I suppose. [For command line discussions, see #687.]

It occurs to me that if we're later going to require the base url, that'll be another breaking change. Maybe we need to take care of that before we go implementing an intermediate measure like `install_type`. I think that may make my head explode. If it's the better way, though...

github_server/action.yml
plocket: Without it, the variable won't be created on failure and that'll mess up the creation of the artifact. [I think we could create it at the start of the action, though, before any failures can happen.]

github_server/action.yml
plocket: I think "race conditions" might be the wrong wording. Can't figure out how to explain it well. Before I implemented this, I saw one occasion where no success message was logged, but everything continued successfully. I believe that was because the `while` checked the result of the curl and caught the change, preventing the code from ever getting to the success message. Maybe I missed something else, though.

github_server/action.yml
plocket: See [above comment about `always()`](https://github.com/SuffolkLITLab/ALKiln/pull/753#discussion_r1300580554).

.github/workflows/github_server.yml
BryceStevenWilley: I don't have a problem with users giving us tags to run! Definitely needed. 

It's just that it's a not-formally-listed input to our `action.yml`; it's not listed in the `inputs` section, it's [only referred to within the action itself](https://github.com/SuffolkLITLab/ALKiln/blob/4625d58a58565c96edfe0fc6be8ad586a0156a2d/action.yml#L65), kinda like a global. I just only realized that now, it's not really related to this PR.

.github/workflows/github_server.yml
BryceStevenWilley: They are the same in bash, and on GH actions I believe; https://github.com/actions/toolkit/issues/940

github_server/action.yml
BryceStevenWilley: Didn't realize that, the placement is fine where it is, I just thought `if: ${{ always() }}` was the default for some reason.

github_server/action.yml
BryceStevenWilley: I think we should suggest people use `docker ps` command

github_server/action.yml
BryceStevenWilley: if the code looked like this:

```bash
while [[ result =="*200*" ]];
    ...
    result=$(curl ...)
    if [[ result == "*200*" ]];
        ...
```

then I don't think there could be a race condition like that. I looked through some older commits, and it looks like it it used to check for "docassemble is starting" which might have caused the issue.

.github/workflows/github_server.yml
plocket: It looks like someone from the team did mark that as a bug. Is it something we should count on?

plocket:
> > Changed ALKilnTests name to ALKiln. It turns out it probably wasn't necessary...? Probably?
> 
> Definitely not necessary, I do suggest that you change it back.

Not sure if I did something differently than you were envisioning, but it seems that changing ALKiln back to ALKilnTests [does break the tests](https://github.com/SuffolkLITLab/ALKiln/actions/runs/5933904641/job/16090001831?pr=753#step:8:287): "DANotFoundError: Interview docassemble.ALKiln:data/questions/AL_tests.yml not found". That might show a flaw in the ALKiln code itself - I think it's using the name of the repo rather than the name of the package in the repo. [That's why I previously assumed the names had to align.] Is that something we should account for in user's tests, or is it just in our own? Or have I missed something else?

github_server/action.yml
plocket: Just for Future People: https://docs.github.com/en/actions/learn-github-actions/expressions#status-check-functions

plocket:
> using the #egg= syntax in GitHub URLs to indicate the package name

Should we go back to using the js install scripts and try to use `#egg=`? If so, we need to figure out how to read the directory of the GitHub repo. That'll probably be easiest in the GitHub action. `dainstall` is still so much simpler. ~Are there downsides to using it?~ [Sorry, I saw your message above later. Not sure where we're landing on this now as our discussions have continued. Portability makes sense, I suppose, if we want to install on servers outside of github actions.]

BryceStevenWilley:
Sorry, I hadn't realized it, but the issue is because the package name from `session_vars.get_package_name` is based on the repo url, which is wrong; we would need to have the correct package name at runtime.

The 2 approaches I see:

* what you mentioned, try to look inside the repo we're given to fine the actual package name / the folder name inside docassemble. This would be reliable but complicated, and might also couple us unnecessarily to the contents of the repo and make https://github.com/SuffolkLITLab/ALKiln/issues/687 more complicated as well.

* If we improve our URL parsing a bit, we could support people giving us URLs with the `#egg=` sytax as well, and then we'd know the proper package name without having to introspect the package. I think this is close to what pip does, and why the egg thing exists in the first place. We could stick with dainstall, or go back to the JS install in this case.

I'm still in support of keeping the name ALKilnTests; I think the name change worked through a loop hole that only works in our code; our code still assumes that repo names are docassemble-XYZ, and it just happens to work when something is just XYZ. DA would still have issues with our repo name by default either way.

.github/workflows/github_server.yml
BryceStevenWilley: Yes; that's the behavior now, and changing that would be a breaking change. If they add some ability to detect it, it'd have to be some other parameter.

lib/docassemble/get_base_interview_url.js
```diff
@@ -11,17 +11,12 @@ let get_base_interview_url = async function () {
   *    when the user is not logged in.) We can't use `reset=2` because it caused an
   *    'Unable to locate interview session' error. */
   let server_url = session_vars.get_da_server_url();
-  console.log(session_vars.get_install_type());
   if (session_vars.get_install_type() === "server") {
-    let x = `${ server_url }/interview?new_session=1&i=${session_vars.get_package_name().replace("-", ".")}%3A`;
-    console.log(`base_interview_url: ${x}`)
-    return x;
+    return `${ server_url }/interview?new_session=1&i=docassemble.${session_vars.get_package_name().replace("-", ".")}%3Adata/questions/`;
```
plocket:
```suggestion
    return `${ server_url }/interview?new_session=1&i=docassemble.${session_vars.get_package_name()}%3Adata/questions/`;
```

github_server/action.yml
plocket: What about our own failure that we're adding in? Won't that be eaten too? Maybe I can just do a try/catch on the curl.

lib/docassemble/server_install.js
plocket: Would love your thoughts on this, @BryceStevenWilley (sorry it's buried in all these comments). I've started looking at the command line code.

github_server/action.yml
BryceStevenWilley: Running without `-e` doesn't eat all exits, just those that come from commands; `exit` will always exit as normal. Running without `-e` is closer the C standard library, where a call to a function can fail, as long as it's not a null error or other segfault, the program will continue. See [the bash docs](https://www.gnu.org/software/bash/manual/bash.html#The-Set-Builtin) for more info there, but I'm happy to explain more if needed.

Bash doesn't have a try-catch in the way you're thinking of. Thinking about it that way, `-e` causes failing commands to throw exceptions, and so removing it just stops that behavior.

lib/docassemble/server_install.js
BryceStevenWilley: TBH I didn't comment because I don't have more thoughts about it, and am a bit wishy-washy on it and didn't want to continue to contradict myself.

Pretty sure we can still just `echo`; if something is installed on the server, the only thing that needs to be in there is `da_install_type`. `da_project_name` is entirely unused, and `artifacts_path` will be added when ALKiln runs the tests, it doesn't have to be there before.

`dainstall` is fewer lines, but it's working differently (by zipping up an on-machine directory instead of from a repository), and does add a new python dependency to things. It's practical, but not clean IMO. `dainstall` is the thing that makes it "not portable", [like we've talked about in other places](https://github.com/SuffolkLITLab/ALKiln/pull/753#discussion_r1300557592). Idk, not arguing for anything in particular here.

I'm starting to think that passing the base URL (or what I'd call the URL path, from the end of the domain to the beginning of the interview/) directly isn't a good idea; it's much more complicated and doesn't add any new possibilities for people, and that sticking with `install_method` has benefits here; it only gets rid of like, 5 lines of JS?

plocket: You have some good points. If we're going to `echo` in the action then we still need to have the js handle a situation that isn't a GitHub action. Do you have preferences for how to handle that? We can just use session vars defaults with a non-env var. That's part of why we called it `session_vars` as it does more than just env vars.

Just a note for Future People - ~`da_project_name` is used locally, but that's not relevant in the GitHub action.~ I'm not sure this is true, but it wouldn't be used on the GitHub server at least. It's used to store the project name in the runtime config, which is created during setup and then used to get the interview's url later. The GitHub server installs on the server, so it doesn't have a playground project name.

plocket:
~For now, for the sake of shared vocabulary, I'm calling the `docassemble-whatever` the package name, as that's what's created when you create a package on the docassemble "Packages" page and is the name of the project on pypi (right?). The inner folder name is the name of the folder inside the package. [I realize the repo name and package name are now the same, but if they're both appropriate, I'd rather not misuse package name. Unless pypi does that differently.] We can discuss further if needed.~

repo name = "docassemble-repo_name_second_part"
~package name~ folder name = name of folder in repo /docassemble
What name does pypi have for the package? [docassemble.repo_name_second_part]

> If we improve our URL parsing a bit, we could support people giving us URLs with the #egg= syntax as well, and then we'd know the proper package name without having to introspect the package.

Users aren't currently giving us URLs, we're the ones generating them with the info we get from the repo, so I'm not sure what you anticipate users providing. We can have our current behavior as default, but I'm not understanding what the user's action input would look like. Could you give an example? Would they provide the [package] name?

[I'm also not sure if it's a huge problem to look at the repo contents, but you may know more about this than me. Are there multiple formats for what that would look like? Aren't we already deeply coupled with the contents of the repo? How can we be less coupled? This is all for default behavior, btw, which we have to keep anyway. Whatever we pass in through the command line will take precedence.]

[When you talk about being coupled to the repo, are you referring to a future in which we could install from a local repo as opposed to from GitHub? That's an interesting thought. Maybe that's not what you meant - we still have a repo in that case.]

[Pypi package name is `docassemble.PackageName`, so we really can't call anything the package name I suppose. The vars we have access to directly are the inner folder and the repo name.]

plocket:
Would it be more helpful or less helpful to provide a default config? [So people aren't required to provide their own if it's not needed and it's faster to set up.]

[Conclusion: da makes its own, so we don't need to, though we do now avoid making an empty config if the dev hasn't given a config.]

plocket:
~`da install` needs folder name to add `#egg=`~ [It doesn't need any of that stuff. It installs the folder that it's in]. Do it all on the action and assume local runs will take care of their own installation, either with manually installing on the server or using the usual ALKiln `setup` script.

ALKiln will not do GitHub server installs from the js, which is a little sad. ~Maybe implement methods just for local tests where you can hand in an `egg` command line arg var and run a local server install that's just separate from any other way of doing it?~ Conclusion: Not worth keeping around a bunch of extra code just for internal local tests and suggest others use that for local server installation too. They might also want to install from a local file too.

Notes on package name uses:
- Name of Project for Playground installation (not breaking for our devs)
- Name of package folder for server installation

github_server/action.yml
plocket: Note for People of The Future: made #801 to follow up

.github/workflows/github_server.yml
plocket: Future People: Made #803

.github/workflows/github_server.yml
plocket: Future People: Made and closed #802

lib/docassemble/get_base_interview_url.js
plocket: Think I did this

lib/docassemble/server_install.js
plocket: Possibly a conclusion: I think now that we're going to start using `dainstall` for GitHub server installation and _also_ use it for remote server Playground installation (to be able to install PRs and such), either we need to keep the old setup scripts for local developers or we need to do a good job documenting how local devs should use the command line to install stuff on remote servers.

.github/workflows/github_server.yml
```diff
@@ -0,0 +1,77 @@
...
+        if: ${{ inputs.enable_tmate }}
+        uses: mxschmitt/action-tmate@v3
+        with:
+          detached: true
+
+      - name: Start the temporary GitHub server
+        id: github_server
+        uses: ./github_server
+        with:
+          CONFIG_CONTENTS: "${{ secrets.CONFIG_CONTENTS }}"
```
plocket, line 37: Would it be useful to add a default config that's the simplest config ever? I think it would only need 2 or 3 props. That way this can be run immediately if there aren't special needs.

BryceStevenWilley: DA has a default config in the docker inage. If we just don't set the config when starting the DA container, we get https://github.com/jhpyle/docassemble/blob/master/Docker/config/config.yml.dist.

github_server/action.yml
```diff
@@ -0,0 +1,126 @@
...
+        shell: bash
+
+      # Don't rely on the environment's version of python
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.10'
+      - run: pip install docassemblecli
+        shell: bash
+      # Install the current directory onto the docassemble server
+      - run: dainstall . --apiurl http://localhost --apikey $DA_ADMIN_API_KEY
```
plocket: Possibly a conclusion about `dainstall`: I think now that we're going to start using `dainstall` for GitHub server installation and also use it for remote server Playground installation (to be able to install PRs and such), either we need to keep the old setup scripts for local developers or we need to do a good job documenting how local devs should use the command line to install stuff on remote servers.

[Or we could write a new setup script that uses bash]

.github/workflows/github_server.yml
plocket: Makes sense! Then we just need to make sure to only conditionally create the config file if the config secret exists. [I think I remember that being a problem when, in the past, I hadn't yet defined the secret.]

Edit:

Trying to confirm this, I'm running the tests over again with no config secret.

[Test 1](https://github.com/SuffolkLITLab/ALKiln/actions/runs/6591655069/job/17927202437): got the error "ALKiln GitHub Server Error: Timed out waiting for docker container to serve the site." The da server log was:

> 2023-10-21 13:37:12,376 CRIT Supervisor is running as root.  Privileges were not dropped because no user is specified in the config file.  If you intend to run as root, you can set user=root in the config file to avoid this message.
> 2023-10-21 13:37:12,376 INFO Included extra file "/etc/supervisor/conf.d/docassemble.conf" during parsing
> 2023-10-21 13:37:12,384 INFO RPC interface 'supervisor' initialized
> 2023-10-21 13:37:12,384 CRIT Server 'inet_http_server' running without any HTTP authentication checking
> 2023-10-21 13:37:12,385 INFO RPC interface 'supervisor' initialized
> 2023-10-21 13:37:12,385 CRIT Server 'unix_http_server' running without any HTTP authentication checking
> 2023-10-21 13:37:12,385 INFO supervisord started with pid 1
> 2023-10-21 13:37:13,388 INFO spawned: 'initialize' with pid 7
> 2023-10-21 13:37:13,391 INFO spawned: 'nascent' with pid 8
> 2023-10-21 13:37:13,393 INFO spawned: 'watchdog' with pid 9
> 2023-10-21 13:37:13,404 INFO success: initialize entered RUNNING state, process has stayed up for > than 0 seconds (startsecs)
> 2023-10-21 13:37:13,405 INFO success: nascent entered RUNNING state, process has stayed up for > than 0 seconds (startsecs)
> 2023-10-21 13:37:14,755 INFO success: watchdog entered RUNNING state, process has stayed up for > than 1 seconds (startsecs)
> 2023-10-21 13:37:48,717 INFO exited: initialize (exit status 1; not expected)

That doesn't really give us info about what's going on, so maybe it wasn't the problem I thought. I'm running it again. It does say "no user is specified in the config file", so maybe it is the problem, not sure.

[Test 2](https://github.com/SuffolkLITLab/ALKiln/actions/runs/6591655069/job/17927064591): Again "ALKiln GitHub Server Error: Timed out waiting for docker container to serve the site." Also, the same kind of debug log

> 2023-10-21 13:55:33,509 CRIT Supervisor is running as root.  Privileges were not dropped because no user is specified in the config file.  If you intend to run as root, you can set user=root in the config file to avoid this message.
> 2023-10-21 13:55:33,509 INFO Included extra file "/etc/supervisor/conf.d/docassemble.conf" during parsing
> 2023-10-21 13:55:33,556 INFO RPC interface 'supervisor' initialized
> 2023-10-21 13:55:33,556 CRIT Server 'inet_http_server' running without any HTTP authentication checking
> 2023-10-21 13:55:33,556 INFO RPC interface 'supervisor' initialized
> 2023-10-21 13:55:33,557 CRIT Server 'unix_http_server' running without any HTTP authentication checking
> 2023-10-21 13:55:33,557 INFO supervisord started with pid 1
> 2023-10-21 13:55:34,561 INFO spawned: 'initialize' with pid 7
> 2023-10-21 13:55:34,565 INFO spawned: 'nascent' with pid 8
> 2023-10-21 13:55:34,569 INFO spawned: 'watchdog' with pid 9
> 2023-10-21 13:55:34,874 INFO success: initialize entered RUNNING state, process has stayed up for > than 0 seconds (startsecs)
> 2023-10-21 13:55:34,874 INFO success: nascent entered RUNNING state, process has stayed up for > than 0 seconds (startsecs)
> 2023-10-21 13:55:36,141 INFO success: watchdog entered RUNNING state, process has stayed up for > than 1 seconds (startsecs)
> 2023-10-21 13:56:47,940 INFO exited: initialize (exit status 1; not expected)

I'm going to assume this is because of the missing config.

github_server/action.yml
```diff
@@ -0,0 +1,126 @@
...
+  DOCASSEMBLE_DEVELOPER_API_KEY:
+    description: "Use this inside `with:` for the ALKiln step. Temporary admin api key. It will be destroyed after the tests are done."
+    value: ${{ steps.api_key_output_step.outputs.DOCASSEMBLE_DEVELOPER_API_KEY }}
+  SERVER_URL:
+    description: "Use this inside `with:` for the ALKiln step. URL of the temporary docassemble server."
+    value: "http://localhost"
+  DA_ADMIN_EMAIL:
+    value: ${{ steps.api_key_output_step.outputs.DA_ADMIN_EMAIL }}
+  DA_ADMIN_PASSWORD:
+    value: ${{ steps.api_key_output_step.outputs.DA_ADMIN_PASSWORD }}
```
plocket: Probably should make a README for this folder. Also probably include a README for the regular action, though I'm not sure where that would go. That would probably be in a different issue. We could also just put it in the suffolk documentation and just provide a link in the README.

[Actually, not sure we _can_ add another file into a github action folder, though research isn't turning anything up. Easy enough to test. Regardless, I think we should make this a documentation issue and link to the documentation when it's ready, so maybe this is a matter for the Future.]

Tangentially, should some kind of default github server workflow be added to the setup interview?

.github/workflows/github_server.yml
plocket: This now handles no config and, supposedly, handles a config with a value and passes. It also logs the message I created for a non-existent config: "The developer did not provide a custom docassemble config file as an input into this action. Docassemble will use its default config.".

Not sure how to confirm that we use the right config when we do have a config, but I restored our config and saw the console message I created for an existing config: "The developer did provided a custom docassemble config file as an input into this action."

lib/utils/session_vars.js
```diff
@@ -174,66 +178,156 @@ session_vars.delete_project_name = function () {
...
-    console.log( `ALKiln debug: Stored name of artifacts directory "${ path_name }" locally.` );
-  }
-  return file;
+  return session_vars.save_runtime_config_var({ key: artifacts_path_key, value: path_name });
+
+  // // Ensure artifacts path file exists
+  // if ( !fs.existsSync( runtime_config_path ) ) {
+  //   // If the file doesn't exist, make one because next we need to read it
+  //   fs.writeFileSync( runtime_config_path, JSON.stringify( {}, null, 2 ) );
+  // }
```
plocket: Regarding this commented-out code and the rest of it too [my brain's fried and I just need a second pair of eyes] - just want a sanity check on what I've done and then I'll clean it up. There might be another one I can abstract too, I'm not sure.

Also, should all the spots using these specific methods just use `session_vars.save_runtime_config_var()` directly?

plocket:
Were we discussing renaming the `github_server` action folder to something like `action_github_server`? Do you know where we landed on that? [Comment at top says to rename it. Made a comment down here for it that can be resolved.]

lib/utils/session_vars.js
```diff
@@ -24,6 +24,11 @@ module.exports = session_vars;
 // These default to null
 // TODO: How to isolate debug elsewhere so this file can use the logger?
 session_vars.get_debug = function () { return process.env.DEBUG || null; };
+session_vars.get_dangerous_debug = function () {
+  /** Note this will cause failure of tests that reassure the user about
+  *   security precautions we took, like not taking a screenshot. */
+  return process.env._ALKILN_DANGEROUS_DEBUG || null;
+};
```
plocket: Remove dangerous debug, but make commits that add and then remove this so that it can be referenced in an issue describing how to do this.

Discussion:
> Added _ALKILN_DANGEROUS_DEBUG for taking pictures of passwords and getting values of secret variables as it became very difficult to debug our tests without that.
> 
> Response: somewhat defeats the purpose of secret stuff, or at least makes them less safe.
> 
> Conclusion: remove code, but make an issue/document the link to the commits. Maybe we add it and remove it. May not notice dangerous screenshots for a few weeks.
> 
> Suggestion: Can we just change the session vars function to always return `null` so we don't have to remove the infrastructure.
> 
> Response: The infrastructure is the concern. An `if` could get mistyped in a future change.
> 
> Question: Can we provide a link to the code we're going to remove so it's easier to redo.
> 
> Response: yes, or make a branch that always has this, but the commit might be better.

github_server/action.yml
```diff
@@ -1,53 +1,143 @@
 name: "ALKiln/github_server: Isolated docassemble tests"
-description: "Automatically test docassemble interviews on GitHub's servers, in an isolated environment. Avoids docassemble reload errors and other external factors"
+description: "Download and start up a temporary docassemble server on GitHub's servers, in an isolated environment. Avoids docassemble reload errors and other external factors."
```
plocket: Change the name of this folder to `action_github_server`

(As per main comment item 2)

.github/workflows/github_server.yml
```diff
@@ -0,0 +1,68 @@
...
+    steps:
+      # Place the root directory in this branch
+      - uses: actions/checkout@v3
+
+      - name: Start the temporary GitHub server
+        id: github_server
+        uses: ./github_server
+        with:
+          CONFIG_CONTENTS: "${{ secrets.CONFIG_CONTENTS }}"
+      - run: echo "Finished starting the GitHub server"
```
plocket: Find all the console messages [and names] meant for users and preface with "ALKiln" or "ALKiln github_server"

lib/utils/session_vars.js
plocket: I'll clean up and leave a comment to discuss later or something. Or maybe make a 'question' issue.

plocket: Decision: We really shouldn't get rid of the other functions. They keep track of the right runtime config key and we don't want each location to need to know that. [We should get rid of the commented code, though.]

github_server/action.yml
plocket: Docs: Made https://github.com/SuffolkLITLab/ALKiln/issues/811
Setup: Made https://github.com/SuffolkLITLab/ALKiln/issues/812

lib/utils/session_vars.js
plocket: Issue for reference: https://github.com/SuffolkLITLab/ALKiln/issues/813

github_server/action.yml
plocket: Made https://github.com/SuffolkLITLab/ALKiln/issues/814

PR #755, opened 2023-08-23
Close #687, pass sources path in through the command line

### LINKED ISSUE #687: Pass feature/sources tests path through the command line

plocket:

~Will have to update~ Have updated ALKilnInThePlayground.

Is this format what we want?

``` bash
npm run test @tag --sources=path/to/sources/dir
```

In this PR, I have:

* ~Added tests for any new features or bug fixes (if relevant)~

### Reason for this PR

This allows the caller to specify the path to the sources folder, something that's very environment specific and hard for ALKiln itself to work out. See the issue.

Closes #687 

Tested locally with:

```
npm run cucumber @none -- --sources=./none
npm run cucumber @none -- --sources=./none --sources=./docassemble/*/data/sources
npm run cucumber @none -- --sources=./docassemble/*/data/sources
```

Tested from ALKilnInThe Playground as well.


plocket:
Conclusion: That is indeed the desired format for the command line args.

plocket:
This will have some merge conflicts with #774. I'd like to prioritize that one, so I'm holding off until that one's merged.

plocket:
I think this may be unblocked now, and ready for review.

plocket:
I think this may be unblocked now, and ready for review. If you have bandwidth, @BryceStevenWilley, would love your input.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@suffolklitlab/alkiln",
-  "version": "5.4.0",
+  "version": "5.1.0-sources-7",
```
BryceStevenWilley, line 3: Should probably change this:

```suggestion
  "version": "5.4.1-sources-8",
```

lib/run_cucumber.js
```diff
@@ -6,64 +6,91 @@ const session_vars = require('./utils/session_vars');
...
+  /** Create the cucumber config, run the tests, create and
+  *   manipluate the output. */
+  const environment = { cwd: process.cwd() };
+  console.log("alkiln-run: environment: %o", environment);
+
+  // process.argv is a list of strings of commands and args.
+  // If using `npm run cucumber`, use `--` before `--sources=./foo`
+  let argv = require(`minimist`)( process.argv.slice(2), {
+    default: {
+      // Testing from GitHub. TODO: Do we want to allow multiple sources paths?
```
BryceStevenWilley: Given that we're using `fg.sync( {suppressErrors:True...})`, I think including all of these as defaults wouldn't hurt.

lib/scope.js
```diff
@@ -1681,75 +1681,52 @@ module.exports = {
...
+    * @returns { [str] } sources folder paths
+    */
     // First make sure the list of files was defined. Don't remove spaces yet -
     // allow a typo where someone forgot to include a space after a comma.
+    // TODO: wish we had the var name here that folks are trying to set
     if ( to_find_names.length === 0 ) {
       reports.addToReport( scope, {
         type: `error`,
-        value: `Your list of files to find is empty.`
+        value: `Your list of files is empty.`
```
BryceStevenWilley: Not sure I get this change. Why would you remove this extra information? Without context, the former report msg is clearer IMO.

BryceStevenWilley (commented):
LGTM, tiny nit comments

package.json
plocket: I'm never sure how to handle this. `v5.1.0-sources-7` is the one that has been published. `v5.4.1-sources-8` doesn't exist, though on a surface level it more accurately describes where in commit history this now sits. I'll be publishing it under a new version anyway, so I think I'm going to keep what I have and resolve this.

lib/scope.js
plocket: To get into this nit, if you're interested, this is just my gut feeling: I feel it muddies the waters because it doesn't necessarily line up with the step they wrote. Getting inaccurately specific can sometimes be more confusing than being more general. They either gave file(s) to "upload" or they gave file(s) to "compare". They didn't mention "finding" anything and they don't know what the code is doing in the background. Again, just my gut feeling. We could try asking a user, though, either now or at some future point.

lib/run_cucumber.js
```diff
@@ -6,64 +6,91 @@ const session_vars = require('./utils/session_vars');
...
+      sources: [ `./docassemble/*/data/sources` ],
+      // From playground: `/usr/share/docassemble/files/playgroundsources/<user_id>/<project>/*.feature`
+      // From playground with S3: `/tmp/playgroundsources/<user_id>/<project>/*.feature`
+    }
+  });
```
plocket: Do you mean like this?

```suggestion

  // We'll use the user info to construct the possible paths to /sources
  let id = session_vars.get_user_id();
  // `project` is the name in the server docassemble "Project" page
  let project = session_vars.get_user_project_name();

  // process.argv is a list of strings of commands and args.
  // If using `npm run cucumber`, use `--` before `--sources=./foo`
  let argv = require(`minimist`)( process.argv.slice(2), {
    default: {
      sources: [
        `./docassemble/*/data/sources`,  // In GitHub folder
        `/usr/share/docassemble/files/playgroundsources/${id}/${project}/*.feature`, // From playground without S3
        `/tmp/playgroundsources/${id}/${project}/*.feature` // From playground with S3
      ],
    }
  });
```

lib/run_cucumber.js
plocket: I made an example lower down of what I _think_ you mean

lib/scope.js
BryceStevenWilley: I get that, but I don't think this make it any clearer to the user. If I was a user, the first thing I'd think if I saw this error message, I'd think "what list?". If they only intended to pass one file to upload (even if they didn't pass any)? In their mind, they didn't pass a list, so this message is still confusing. It should be followed by a specific error message from the upload or compare steps anyway.

I guess this gets into who the messages are for. When they contain relatively little information like this, IMO they should at least be useful to developers. Several months out, the former message would be still be useful to me, whereas the latter wouldn't, and depending on how the user is communicating their issue with them, it either adds a simple search, or a lot more back and forth than necessary.

lib/run_cucumber.js
BryceStevenWilley: Yeah! I'd make sure the tests run fine, we can't really test this because it's not gonna be used in ALKilnInThePlayground, but it can't hurt, I like it a lot better than commented code.

lib/scope.js
plocket: Good point. What do you think of `There are no file(s) to find written in your Step.`? I'll push that for now, dependent on your input. [Or maybe I'll try to think of something with less passive voice.]

plocket: `Your Step is missing the names of which file(s) to find.`

BryceStevenWilley: I like the `Your step is missing...` one!

BryceStevenWilley (approved):

PR #769, opened 2023-10-02
input type -> datatype for number

BryceStevenWilley:
`input type: number` doesn't exist and doesn't actually validate anything. This is a similar issue to https://github.com/SuffolkLITLab/docassemble-ALToolbox/pull/193.

I'm not sure that we check for DA validation of non numbers, but figured I'd make the change anyway, to prevent any confusing down the line. Manually tested to make sure that it works, and it does.


plocket (approved):
Awesome, thank you!

Issue #748, opened 2023-08-13
Discuss: Screenshot for accessibility failures are confusing, but removing them may be confusing too.


plocket:
A problem with accessibility errors is that, even when they're the only error, the test still takes a screenshot of the final page of the test, even though that page isn't relevant. Maybe we should avoid taking a screenshot at the end if the only failures are accessibility failures.

Additionally, should we take a screenshot of the page/field where the accessibility issue arose? [Conclusion: We should take the a11y error pic just after adding the a11y failure info to the report.]

BryceStevenWilley:
If we start saving the DOM as HTML as well as screenshoting (see #763), that would be very useful for accessibility issues.

plocket:
That's true, but the screenshot itself is still not relevant and may be confusing/misleading to the dev. We'd still need to make sure to take screenshots if there were other errors, of course.

BryceStevenWilley:
> the screenshot itself is still not relevant and may be confusing/misleading to the dev

Should have clarified; I agree that the current situation of taking a screenshot of the final page isn't useful. However, a screenshot of the page that the accessibility error is on would still be useful. Sure, it doesn't fit in with our current UX of encouraging people to look through the screenshots to see the errors, but the screenshot is still useful to have when looking for certain accessibility issues.

plocket:
I think we're actually on the same page and I mistook what you were referring to in your comment. All sounds good 👍

plocket:
We should take the a11y error pic just after adding the a11y failure info to the report.

PR #774, opened 2023-10-11
Pdfjs

### LINKED ISSUE #25: Develop the ability to compare pdfs to each other
The eventual goal would be a step to compare an existing pdf to a downloaded pdf. First we have to figure out how to compare two PDFs, though.

BryceStevenWilley:
I added the ability to compare PDFs! It's currently one step, `I expect the baseline PDF "something-in-sources.pdf" and the new PDF "downloaded.pdf" to be the same`, but happy to add more, depending on what we think would be useful for folks.

Making this a draft to make sure the API is as we want it.

In this PR, I have:

* [x] Added my changes to the CHANGELOG.md at the top, under the "Unreleased" section (still unsure what the final API will look like)

Fix #25. 

Just added the automated tests.


lib/scope.js
```diff
@@ -2503,6 +2511,26 @@ module.exports = {
...
     },  // Ends scope.steps.download()
 
+    compare_pdfs: async function (scope, {existing_pdf_path, new_pdf_path}) {
+      let full_existing_path = await scope.findFiles(scope, {uploads_names: [existing_pdf_path]});
+      let full_new_pdf_path = `${scope.paths.scenario}/${new_pdf_path}`;
+      let diffs = await pdf_reader.compare_pdfs(full_existing_path[0], full_new_pdf_path);
+      if (diffs) {
+        let diff_str = diffs.reduce((err_str, part) => {
+          if (part.added) {
+            return err_str + `added "${ part.value }", `
```
BryceStevenWilley: Consider adding newlines?

Test on long PDFs that are completely different.

Consider saving screenshots of the two different PDFs.

Consider adding a line to the log of "If this new PDF is what you want it to be, just replace it". Way in the future, consider a tool that copies the new PDFs into a separate artifact folder, for easy copy-replacing into the sources folder.

lib/utils/pdf_reader.js
```diff
@@ -0,0 +1,111 @@
+// All functions dealing with getting information from PDFs
+
+const pdfjsLib = require("pdfjs-dist/legacy/build/pdf.js");
+const Diff = require("diff");
```
BryceStevenWilley:
```suggestion
const diffLib = require("diff");
```

lib/utils/pdf_reader.js
```diff
@@ -0,0 +1,111 @@
...
+
+pdf_reader.compare_pdfs = async function(existing_pdf_path, new_pdf_path) {
+  /**
+   * Returns the change objects between the PDF fields, if there's any diff
+   */
+  let existing_doc = await _load_pdf(existing_pdf_path);
+  let new_doc = await _load_pdf(new_pdf_path);
+  let existing_content = await _get_content(existing_doc);
+  let new_content = await _get_content(new_doc);
+
```
BryceStevenWilley, line 94:
```suggestion

  // Assumption: authors won't be changing their PDF templates programmatically;
  // so if the text of the PDFs are different, then it's actually the fields that are different,
  // but they aren't editable / are flattened into the PDF. 
```

lib/scope.js
BryceStevenWilley: Follow up on this: New line separated with markdown-like bullets looks good enough IMO. It's not great, but it handles really long PDFs better, and is at least structured a bit more.

Screenshot from 2023-10-13 14-27-45

Making an issue for the future work now.

BryceStevenWilley: PDF fields diff pretty well too, it only shows the different value.

```
ERROR: The PDFs were not the same.
The new PDF added: 
-       "value": "Proxyname1"


The new PDF removed: 
-       "value": "Proxyname12"
```

BryceStevenWilley:
Ready for review, addressed all of the comments, and made slight cleanups to the `findFiles` function, but everything else remained the same!

lib/scope.js
```diff
@@ -2503,6 +2514,29 @@ module.exports = {
       await scope.detectDownloadComplete( scope );
     },  // Ends scope.steps.download()
 
+    compare_pdfs: async function (scope, {existing_pdf_path, new_pdf_path}) {
+      let existing_paths = await scope.findFiles(scope, {to_find_names: [existing_pdf_path]});
+      if (existing_paths.length == 0) {
+        // The row wasn't used
+        let msg = `Could not find the existing PDF at ${ existing_pdf_path }`;
+        expect(existing_paths.length, msg).to.be.above(0);
```
plocket: Good message. We should probably add that to the report too, though. Untested:

```suggestion
        let msg = `Could not find the existing PDF at ${ existing_pdf_path }`;
        reports.addToReport(scope, { type: `error`, value: msg });
        expect(existing_paths.length, msg).to.be.above(0);
```

Just thinking over this - In the past, we've decided to err towards being permissive and warning users as opposed to erroring, saying something like "so we couldn't compare PDFs". That said, you might be right to error here. If we just warn and the tests pass, that seems misleading.

For far in the future: I wonder if there's a way to have both - remember this, but also finish the test, and error later, like the a11y stuff.

plocket: Just as a note, #687 is about to add [a warning](https://github.com/SuffolkLITLab/ALKiln/pull/755/files#diff-4b632aedc4c91285de358dd7cbcc91a52724745c5c759b81b3a6d80e8d6d7abcR1824) about not finding files. When that's split into `find_files`, that might be a better place to put a warning. We can then error here. We can probably keep this and choose later where is a more appropriate to put it and what it should look like.

lib/steps.js
```diff
@@ -705,6 +705,13 @@ Then(/I get the(?: page's)?(?: JSON)? var(?:iable)?s and val(?:ue)?s/i, { timeou
   );
 });
 
+Then(/I expect the baseline PDF "([^"]+)" and the new PDF "([^"]+)" to be the same/i, async (existing_pdf_path, new_pdf_path) => {
```
plocket, line 709: Is it worth making some of this optional so there can be a shorthand? I'm open to either. It would be something like the below (tested at regex101.com):

```suggestion
Then(/I expect(?: the baseline PDF)? "([^"]+)" and(?: the new PDF)? "([^"]+)" to be the same/i, async (existing_pdf_path, new_pdf_path) => {
```

The shortened form is `I expect "simple-doc-Baseline.pdf" and "simple-doc.pdf" to be the same`

lib/scope.js
```diff
@@ -2503,6 +2514,29 @@ module.exports = {
...
+      let diffs = await pdf_reader.compare_pdfs(full_existing_path, full_new_pdf_path);
+      if (diffs) {
+        let added_str = diffs.filter(part => part.added).reduce((err_str, part) => {
+          return err_str + `- ${ part.value }\n`
+        }, 'The new PDF added: \n');
+        let removed_str = diffs.filter(part => part.removed).reduce((err_str, part) => {
+          return err_str + `- ${ part.value }\n`
+        }, 'The new PDF removed: \n');
+        let msg = `The PDFs were not the same.\n${ added_str }\n${removed_str}\n\n You can see the full PDFs at ${ full_existing_path } and ${ full_new_pdf_path}`;
+        await reports.addToReport(scope, { type: `error`, value: msg });
```
plocket: Reports are sync now

```suggestion
        reports.addToReport(scope, { type: `error`, value: msg });
```

lib/steps.js
```diff
@@ -705,6 +705,13 @@ Then(/I get the(?: page's)?(?: JSON)? var(?:iable)?s and val(?:ue)?s/i, { timeou
...
+  return await wrapPromiseWithTimeout(
```
plocket, line 716: Maybe a description here, take it or leave it

```suggestion
  /** Compare the text or fields of a baseline PDF and the newly downloaded PDF.
  * 
  * Variations:
  * Then I expect the baseline PDF "simple-doc-Baseline.pdf" and the new PDF "simple-doc.pdf" to be the same
  * Then I expect "simple-doc-Baseline.pdf" and "simple-doc.pdf" to be the same
  */
  
  return await wrapPromiseWithTimeout(
```

or

```suggestion
  /** Compare the text or fields of a baseline PDF and the newly downloaded PDF.
  * 
  * Variations:
  * Then I expect the baseline PDF "simple-doc-Baseline.pdf" and the new PDF "simple-doc.pdf" to be the same
  */
  
  return await wrapPromiseWithTimeout(
```

plocket (changes requested):
So psyched for this! A couple things to maybe change and a couple notes. I think the report changes are the main things and the others are dealer's choice. Sorry I didn't catch some of this before.

lib/steps.js
BryceStevenWilley: That is what I had at first, but since we treat the first argument differently from the second (the first is always searched for in the `sources` folder, and the second is always from the downloads), I wanted it to be unambiguous. Adding that shorthand makes it too easy for  people to write code that will confuse themselves and other readers later IMO.

If we did want this short hand, we'd need to let either argument be source or download, and the complexity of letting people use either argument for either PDF (but ensuring that one one from the source PDFs and the other is from the downloads, i.e. people aren't accidentally comparing two source PDFs in a test) isn't worth it IMO.

lib/scope.js
BryceStevenWilley: Sorry, forgot about adding to the report. IMO, yeah, a warning in `findFiles`, and but potentially error in the function that calls `findFiles` is a good option.

TBH doing what we do with the accessibility stuff is a good idea, and not hard to implement. It will add to #748, but whatever solution we have for that would also fix this as well. I'll try it out, if it's taking too long, I'll make an issue for it.

BryceStevenWilley: Found a good solution for this, I think it works out really well, except for the above mentioned extra screenshot with #748, but this and the accessibility code is very closely related, and the solution to one should fix the other.

plocket (approved):
Great! Thanks for digging into this, it's a fantastic addition 🚀

plocket:
Add to the changelog if you have a mo

Issue #492, opened 2022-01-15
Bug: Downloads take the max time, they don't end the step when they're done downloading (I think)


plocket:
This is happening in MADE local tests at least. Whatever max seconds is set to, the step takes that long to complete, no matter when the files are done downloading. Not super high priority, just annoying.

BryceStevenWilley:
We mentioned this on the Deep dive call today: it seems to be a deficiency in Puppeteer. These few comments in the middle of a long thread on the issue have some solutions:
* [longer solution](https://github.com/puppeteer/puppeteer/issues/299#issuecomment-435609879)
* [shorter, but idk if it works](https://github.com/puppeteer/puppeteer/issues/299#issuecomment-435662328)
* [one using axios, that I think we're using](https://github.com/puppeteer/puppeteer/issues/299#issuecomment-455637075), I think it manually transfers over the cookies to axios so permissions should work fine
* [a self-contained one](https://github.com/puppeteer/puppeteer/issues/299#issuecomment-474435547)
* [one that specifically ends up with two downloads](https://github.com/puppeteer/puppeteer/issues/299#issuecomment-474435547), we could manually delete the first one (where we can't control the location of the download

plocket:
This is out of left field, but the Playwright package would make this much easier. [An article about using it](https://scrapingant.com/blog/playwright-download-file). At some point I'll make a discussion about moving to Playwright, pros and cons.

Edit: migrating from puppeteer to playwright: https://playwright.dev/docs/puppeteer

BryceStevenWilley:
There's an older issue about switching to playwright: https://github.com/SuffolkLITLab/ALKiln/issues/212.  Could switch it to a discussion, but I'm okay with sticking to that issue.

plocket:
Not sure if this was contained in one of the previous link, but [one solution on stackoverflow suggests using `get`](https://stackoverflow.com/a/74476592) because download behavior has a history of being inconsistent. That might be able to be timed differently. Not sure what that would involve for us considering possible encryption, etc. Maybe it's not possible.

plocket:
#798 partially addressed this. If we can get rid of the old download behavior fallback, we can close this. We need to research under what circumstances `fetch` would fail to download an interview.

PR #798, opened 2023-10-18
Faster downloads

BryceStevenWilley:
Speeds up the downloads by directly downloading the file instead of trying to rely on `response` callbacks, which almost always timeout.

I made things a bit resilient; I think things work fine without all of the `failed_to_download` stuff (see https://github.com/SuffolkLITLab/ALKiln/actions/runs/6556325525/job/17806135932, an earlier test run without that code), but figured I'd keep the existing code in, in case there are situations we couldn't foresee. Happy to rip out the old code if we want to though.

I tried a lot of different options, none of them seemed to work particularly well:

* `https.get` doesn't work with http://localhost (might be a me problem, but IMO, it's frustrating), the workaround is hacky and uses both `https.get` and `http.get`.
    * I also had issues getting the file to download still? Just kept getting `File Not Found`. I _think_ it's because DA by default checks the cookies / logged in status of who's downloading a file (that's why you need to use `url_for(private=False)`). So in general, trying to access the file from outside of the browser session is going to be tricky, if not impossible.
* You can't pass back a blob of buffer object directly from the browser (https://github.com/puppeteer/puppeteer/issues/3722). I took the suggested code from that issue and used it here. 

### In this PR, I have:

#492

Running a test that downloads a PDF (`@o17`) runs in 3.4 seconds now, as opposed to 33 seconds on `v5`.

### DIFF
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -2484,11 +2484,28 @@ module.exports = {
       let msg = `"${ filename }" seems to be missing. Cannot find a link to that document.`;
       if ( !elem ) { reports.addToReport(scope, { type: `error`, value: msg }); }
       expect( elem, msg ).to.exist;
+      
+      const res = await scope.page.evaluate(el => {
+        const target = el.getAttribute("target");
+        const url = el.getAttribute("href");
+        if (target === "_blank") {
+          return fetch(url, {
+            method: "GET"
+          }).then(r => r.text()); 
+        } else {
+          return new Promise(() => '');
+        }
+      }, elem);
+      if (res !== '') {
+        fs.writeFileSync(filename, res); 
+      } else {
+        console.log('could not download file using fetch');
+        scope.toDownload = filename;
+        // Should this be using `scope.tapElement()`?
+        await elem.evaluate( elem => { return elem.click(); });
+        await scope.detectDownloadComplete( scope );
+      }
 
-      scope.toDownload = filename;
-      // Should this be using `scope.tapElement()`?
-      await elem.evaluate( elem => { return elem.click(); });
-      await scope.detectDownloadComplete( scope );
     },  // Ends scope.steps.download()
 
     compare_pdfs: async function (scope, {existing_pdf_path, new_pdf_path}) {
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -2485,30 +2485,35 @@ module.exports = {
       if ( !elem ) { reports.addToReport(scope, { type: `error`, value: msg }); }
       expect( elem, msg ).to.exist;
 
-      const pdfString = await scope.page.evaluate(el => {
-        const target = el.getAttribute("target");
-        const url = el.getAttribute("href");
-        return new Promise(async (resolve, reject) => {
-          const response = await fetch(url, {method: "GET"}); //type: 'application/pdf'
-          let blob = await response.blob()
-          const reader = new FileReader();
-          reader.readAsBinaryString(blob);
-          reader.onload = () => resolve(reader.result);
-          reader.onerror = () => reject('Error occurred while reading binary string');
-        });
-      }, elem);
-      const pdfData = Buffer.from(pdfString, 'binary');
-      fs.writeFileSync(`${ scope.paths.scenario }/${ filename }`, pdfData);
-      //if (res) {
-      //  const buffer = Buffer.from(res);
-      //  console.log("Done writing");
-      //} else {
-      //  console.log('could not download file using fetch, using click download method');
-      //  scope.toDownload = filename;
+      failed_to_download = false;
+      try {
+        const binaryStr = await scope.page.evaluate(el => {
+          const url = el.getAttribute("href");
+          return new Promise(async (resolve, reject) => {
+            const response = await fetch(url, {method: "GET"});
+            const reader = new FileReader();
+            reader.readAsBinaryString(await response.blob());
+            reader.onload = () => resolve(reader.result);
+            reader.onerror = () => reject('Error occurred while reading binary string');
+          });
+        }, elem);
+        if (binaryStr !== '') {
+          const fileData = Buffer.from(binaryStr, 'binary');
+          fs.writeFileSync(`${ scope.paths.scenario }/${ filename }`, fileData);
+        } else {
+          failed_to_download = true;
+        }
+      } catch {
+        failed_to_download = true;
+      }
+
+      if (failed_to_download) {
+        console.log('could not download file using fetch, using click download method');
+        scope.toDownload = filename;
         // Should this be using `scope.tapElement()`?
-      //  await elem.evaluate( elem => { return elem.click(); });
-      //  await scope.detectDownloadComplete( scope );
-     // }
+        await elem.evaluate( elem => { return elem.click(); });
+        await scope.detectDownloadComplete( scope );
+      }
     },  // Ends scope.steps.download()
 
     compare_pdfs: async function (scope, {existing_pdf_path, new_pdf_path}) {
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -2512,7 +2512,7 @@ module.exports = {
       }
 
       if (failed_to_download) {
-        reports.addToReport(scope, { type: `warning`, value: `Could not download file using fetch (${ err_msg }). Using click download method` });
+        reports.addToReport(scope, { type: `warning`, value: `Could not download file using fetch (${ err_msg }). ALKiln will now fallback to the click download method` });
         scope.toDownload = filename;
         // Should this be using `scope.tapElement()`?
         await elem.evaluate( elem => { return elem.click(); });
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -2484,28 +2484,31 @@ module.exports = {
       let msg = `"${ filename }" seems to be missing. Cannot find a link to that document.`;
       if ( !elem ) { reports.addToReport(scope, { type: `error`, value: msg }); }
       expect( elem, msg ).to.exist;
-      
-      const res = await scope.page.evaluate(el => {
+
+      const pdfString = await scope.page.evaluate(el => {
         const target = el.getAttribute("target");
         const url = el.getAttribute("href");
-        if (target === "_blank") {
-          return fetch(url, {
-            method: "GET"
-          }).then(r => r.text()); 
-        } else {
-          return new Promise(() => '');
-        }
+        return new Promise(async (resolve, reject) => {
+          const response = await fetch(url, {method: "GET"}); //type: 'application/pdf'
+          let blob = await response.blob()
+          const reader = new FileReader();
+          reader.readAsBinaryString(blob);
+          reader.onload = () => resolve(reader.result);
+          reader.onerror = () => reject('Error occurred while reading binary string');
+        });
       }, elem);
-      if (res !== '') {
-        fs.writeFileSync(filename, res); 
-      } else {
-        console.log('could not download file using fetch');
-        scope.toDownload = filename;
+      const pdfData = Buffer.from(pdfString, 'binary');
+      fs.writeFileSync(`${ scope.paths.scenario }/${ filename }`, pdfData);
+      //if (res) {
+      //  const buffer = Buffer.from(res);
+      //  console.log("Done writing");
+      //} else {
+      //  console.log('could not download file using fetch, using click download method');
+      //  scope.toDownload = filename;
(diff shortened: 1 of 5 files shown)


lib/scope.js
```diff
@@ -2485,10 +2485,35 @@ module.exports = {
...
+      failed_to_download = false;
+      try {
+        const binaryStr = await scope.page.evaluate(el => {
+          const url = el.getAttribute("href");
+          return new Promise(async (resolve, reject) => {
+            const response = await fetch(url, {method: "GET"});
+            const reader = new FileReader();
+            reader.readAsBinaryString(await response.blob());
+            reader.onload = () => resolve(reader.result);
+            reader.onerror = () => reject('Error occurred while reading binary string');
```
plocket: A chance of making the error message more clear. That might also be useful to add to the report on failure when we catch the rejection. Have to check the rest of the code to see what's going on in the download flow.

```suggestion
            reader.onerror = () => reject('Error occurred while reading binary string while trying to download a file');
```

Would be great to add the name of the file that we were attempting to download.

lib/scope.js
```diff
@@ -2485,10 +2485,35 @@ module.exports = {
...
+          fs.writeFileSync(`${ scope.paths.scenario }/${ filename }`, fileData);
+        } else {
+          failed_to_download = true;
+        }
+      } catch {
+        failed_to_download = true;
+      }
+
+      if (failed_to_download) {
+        console.log('could not download file using fetch, using click download method');
```
plocket: Did you mean to add this to the report? And/or to console log on debug (or use `log.debug()` which we should start doing a bit more often)?

It's good that it hasn't failed for us, but I do want to keep the failure condition around and add to the report. If something's that wrong with the user's file, that's good for them to know. I do think that even if we can't think of it now, someone can find a way to make a download fail.

lib/scope.js
```diff
@@ -2485,10 +2485,35 @@ module.exports = {
...
+        scope.toDownload = filename;
+        // Should this be using `scope.tapElement()`?
+        await elem.evaluate( elem => { return elem.click(); });
+        await scope.detectDownloadComplete( scope );
```
plocket, line 2519: Why still also trigger the old method? I'm actually not completely sure on how it's currently running faster without the ability to set `scope.downloadComplete = True`, so maybe there's something I'm not understanding about the changes.

If we don't need these lines, though, we can get rid of `detectDownloadComplete` and I think we can get rid of [the section in `steps.js` that listens for `response`](https://github.com/SuffolkLITLab/ALKiln/blob/faster_downloads/lib/steps.js#L203).

lib/scope.js
```diff
@@ -2485,10 +2485,35 @@ module.exports = {
...
+            const response = await fetch(url, {method: "GET"});
+            const reader = new FileReader();
+            reader.readAsBinaryString(await response.blob());
+            reader.onload = () => resolve(reader.result);
+            reader.onerror = () => reject('Error occurred while reading binary string');
+          });
+        }, elem);
+        if (binaryStr !== '') {
+          const fileData = Buffer.from(binaryStr, 'binary');
+          fs.writeFileSync(`${ scope.paths.scenario }/${ filename }`, fileData);
```
plocket, line 2503: This is great - we can let the author rename the file now, which might be good if they want to do things like generate baseline files with the right names. [Made #800]

plocket (commented):
Speeding this up will be a huge win. I think we may need to check on some stuff first as this may be a bit more complex than it would first appear, but maybe I'm off - I feel like I'm missing some of this picture. Anyway, some notes here and some in the review comments. At least some of them are relevant.

Numbered for easier referencing:

1. Do we really need to update shrinkwrap?
2. Added some comments about adding messages to let the author know what's going on.
3. Should we add something in the report when the download has succeeded? Before we didn't know if it had succeeded, but now we should.
4. I think this would be especially good to add to the changelog as an internal change.
5. Does this change handle the timeout running out? Previously, we'd go till the timeout and just continue the tests regardless of what happened (since we couldn't tell what happened). If the files didn't download, there wasn't anything to do about it. Now, though, we may need to handle the timeout if our await goes too long. Not sure if we'd want to error or warn at that point. We may want to add a test for whatever behavior we choose - make a huge image file or something to download.
6. When comparing downloads, did we make a provision for if/when the download doesn't exist (since it may not complete in time)? ~I'll go check when I have a mo.~ It looks like we didn't, so we might need to take care of that in another PR fairly soon.

lib/scope.js
BryceStevenWilley: I could definitely add the same of the file itself to the message, I'll do that.

lib/scope.js
BryceStevenWilley: Yeah, console is a shameful hack I still use, I'll change that.

lib/scope.js
BryceStevenWilley: We only trigger the old method if the download didn't work the first time. It only happens something goes wrong, which I anticipate being something specific to the blob/fetch stuff we're doing.

I'm a bit confused; do you want to remove `detectDownloadComplete` or not? You mentioned above keeping this failure condition/fallback around.

BryceStevenWilley:
>  Do we really need to update shrinkwrap?

That happens when I run `npm install`, which I needed to do after #755 merged.

> Should we add something in the report when the download has succeeded? Before we didn't know if it had succeeded, but now we should.
> I think this would be especially good to add to the changelog as an internal change.

:+1: (I've been waiting for changelog stuff when I put out multiple PRs b/c I don't know which will go out first)

> Does this change handle the timeout running out? Previously, we'd go till the timeout and just continue the tests regardless of what happened (since we couldn't tell what happened). If the files didn't download, there wasn't anything to do about it. Now, though, we may need to handle the timeout if our await goes too long. Not sure if we'd want to error or warn at that point. We may want to add a test for whatever behavior we choose - make a huge image file or something to download.

Good points. IMO I'd rather error, since if we tried to download something and didn't download it, something's likely wrong with the server (testing might be hard; even really big stuff will download quickly on the same machine).

> When comparing downloads, did we make a provision for if/when the download doesn't exist (since it may not complete in time)?

This will exist though? It completes in this step (unless there's a timeout, and I that case I think we should error, as mentioned above).

lib/scope.js
plocket: My bad, I totally missed the `if`. It's good this way for now, though I'm not sure how we'll ever determine if it's useful or not.

With "failure condition" I thought you meant get rid of failure handling in general and just let the errors speak for themselves, so I wasn't a fan of that. Glad that it's not a problem!

Resolve if you want 👍

plocket:
> 1. shrinkwrap

I think we could start using [`npm ci`](https://docs.npmjs.com/cli/v6/commands/npm-ci) for this. It's meant for pipelines, but I think it would serve the purpose we're looking for - installing a consistent version based on the lockfile each time. When we need to update for security or for new packages we would, of course, just `npm install`.

Another option is to [use `npm install --no-save`](https://github.com/npm/npm/issues/17761#issuecomment-334200107) in future to avoid updating shrinkwrap, but that unfortunately doesn't really make things stable - the shrinkwrap file doesn't change, but your local environment is still different than what the shrinkwrap uses. It looks like we all end up using different shrinkwraps.

I wonder if [item 4 here](https://github.com/npm/npm/issues/17979#issuecomment-326712196) indicates that what happened to your shrinkwrap is a bug and may be worth reporting.

> IMO I'd rather error

Sounds good to me 👍 

> testing might be hard

Yeah, we probably won't be able to write tests that will consistently pass (by failing) until we do some research about implementing throttling.

> [The download] completes in this step [unless it errors]

Good point. I can still see a situation where someone would forget to include a download step before comparing. We might handle that in a different issue, though.

plocket:
Not sure if you saw that note about a success message, so just repeating that here.

lib/scope.js
BryceStevenWilley: > though I'm not sure how we'll ever determine if it's useful or not

I'm hoping it'll show up in our logs or someone else's and we'll notice. Might be wishful thinking, but we don't have any better way, and I don't want this to break people at all.

BryceStevenWilley:
I've been a bit lenient about shrinkwrap changes. I'll avoid them more in the future (though this time it's because the version in `package.json` on main is still `5.1.7-sources-7`, like I pointed out in https://github.com/SuffolkLITLab/ALKiln/pull/755#discussion_r1360808636 👀 , not a bug).

lib/scope.js
```diff
@@ -2494,21 +2495,24 @@ module.exports = {
...
         }
-      } catch {
+      } catch (error) {
         failed_to_download = true;
+        err_msg = error;
       }
 
       if (failed_to_download) {
-        console.log('could not download file using fetch, using click download method');
+        reports.addToReport(scope, { type: `warning`, value: `Could not download file using fetch (${ err_msg }). Using click download method` });
```
```suggestion
        reports.addToReport(scope, { type: `warning`, value: `Could not download file using fetch (${ err_msg }). ALKiln will now fallback to the click download method` });
```

plocket:
> testing might be hard; even really big stuff will download quickly on the same machine

It occurred to me that we can decrease the max timeout for the Step to a fraction of a second and thus trigger a timeout. Would that work for our purposes? It's only one type of error, but that may still be worth testing.

BryceStevenWilley:
> It occurred to me that we can decrease the max timeout for the Step to a fraction of a second and thus trigger a timeout. 

Good point, I can try that! IMO, we can try to manually and see how it does. I don't think we need to add it to the test suite though.

lib/scope.js
```diff
@@ -2485,10 +2485,39 @@ module.exports = {
...
+            reader.onerror = () => reject(`Error occurred on page when downloading ${ url }: ${ reader.error }`);
+          });
+        }, elem);
+        if (binaryStr !== '') {
+          const fileData = Buffer.from(binaryStr, 'binary');
+          fs.writeFileSync(`${ scope.paths.scenario }/${ filename }`, fileData);
+          reports.addToReport(scope, { type: `info`, value: `Downloaded ${ filename } (${ fileData.length } bytes) to ${ scope.paths.scenario }`});
+        } else {
+          failed_to_download = true;
+          err_msg = `Couldn't download ${ url }, dinary data for download was empty`;
```
plocket: ~err_msg = `Couldn't download ${ url }, binary data for download was empty`;~

plocket: Or

```suggestion
          err_msg = `Couldn't download ${ filename }, binary data for download was empty`;
```

(I might have missed it, but I don't think `url` exists in this scope)

plocket (approved):
This is looking pretty good to me! I think once you feel comfortable with it, we can send it out into the wild!

BryceStevenWilley:
> It occurred to me that we can decrease the max timeout for the Step to a fraction of a second and thus trigger a timeout.

I did this, and got the following printout in the cucumber report:

```
1) Scenario: I compare the same PDFs (attempt 2) # docassemble/ALKilnTests/data/sources/observation_steps.feature:220
   ✔ Before # lib/steps.js:104
   ✔ Given I start the interview at "test_pdf" # lib/steps.js:165
   ✔ Then the question id should be "proxy vars" # lib/steps.js:453
   ✔ When I set the var "x[i].name.first" to "Proxyname1" # lib/steps.js:939
   ✔ And I tap to continue # lib/steps.js:740
   ✔ Then I sign # lib/steps.js:984
   ✔ And I tap to continue # lib/steps.js:740
   ✔ Then the question id should be "simple doc" # lib/steps.js:453
   ✔ Given the max seconds for each step in this scenario is 0.1 # lib/steps.js:279
   ✖ When I download "simple-doc.pdf" # lib/steps.js:999
       Error: Action did not complete within 0 milliseconds
           at Timeout.<anonymous> (/home/brycew/Developer/LITLab/code/ALKiln/lib/cucumber_8_shim.js:18:14)
           at listOnTimeout (node:internal/timers:569:17)
           at process.processTimers (node:internal/timers:512:7)
   - And I expect the baseline PDF "simple-doc-Baseline.pdf" and the new PDF "simple-doc.pdf" to be the same # lib/steps.js:708
   ✔ After # lib/steps.js:1113
```

The PDF file did actually download fine, because I guess the download had enough time to start. The conflicting states isn't great, but I'm assuming that the presence of a clear "Action did not complete within <time limit>" message is good enough for people. I tried uploading a larger PDF to do this with, but ran into issues (DA doesn't like too big of files, and it doesn't like PDFs without fields, all of the big PDFs I have don't have fields and it'd be difficult to add them to it). 

I'll add to the changelog now and merge!

PR #799, opened 2023-10-18
Missing docx

BryceStevenWilley:
Was going through old branches, and found this fairly simple one that seems to have been left behind. Adds a test for a missing DOCX template on the DA side. The test fails as expected, so it seems like an easy merge.


plocket (approved):
I'm not sure it'll get the same error everywhere because sometimes errors are different when the interview is run in the playground vs. installed on a server, but it'll be good to know that we need to deal with it if/when it comes up, so merge or not as you wish!

Issue #637, opened 2023-01-09
Random Answering: Radio buttons are skewed 


BryceStevenWilley:
Only the last radio button is ever pressed. 

In https://github.com/SuffolkLITLab/ALKiln/pull/633,  the behavior was adjusted to each radio button is considered separately, heavily weighting the answers towards the latter radio buttons; the last button has a 50% chance of being clicked, and the 2nd to last has a 25% chance, etc.

BryceStevenWilley:
Idea to fix: use the radio button grouping in docassemble to get all radio buttons for a field in one collection.

plocket:
Also this to keep in mind:

> ...this [fix] probably shouldn't need to exist, I just know that `get_custom_datatype` and `is_custom_datatype` would say that radio buttons were custom datatypes when they weren't actually. I would guess that `get_custom_datatype` isn't correct, but I don't know how...

_Originally posted by @BryceStevenWilley in https://github.com/SuffolkLITLab/ALKiln/pull/633#discussion_r1067413876_

PR #806, opened 2023-10-22
Remove radio button changes

plocket:
I think we should also just get rid of the back button presses completely for the original PR, but this is one step forward.

[Updated] ~Note: This is failing because of that one error message test. When we rebase onto master on the original branch, we can easily fix this. I might actually go rebase that now and then rebase this again depending on how snarled it gets.~

In this PR, I have:

* N/A [ ] Added my changes to the CHANGELOG.md at the top, under the "Unreleased" section
* N/A [ ] Ensured issues that this PR closes will be [automatically closed](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue)

### Reason for this PR

Skip radio button fixes for random input tests for now. Pass tests.

~### Links to any solved or related issues~

~### Any manual testing I have done to ensure my PR is working~


BryceStevenWilley:
#637 is the tracking issue for the removed code btw.

> I think we should also just get rid of the back button presses completely for the original PR

I don't think so. Back buttons don't undo external network calls or saving things to Redis, so making sure programs can handle (or least don't crash) when users press back buttons is still important.

BryceStevenWilley (approved):

PR #820, opened 2023-10-31
Close 819, hide docker logs by default

### LINKED ISSUE #819: GitHub server: improve security of sensitive information
We won't print any sensitive info, but we don't know what docassemble will print in its logs.

1. [By default], skip creating a docker log artifact. That way, users can be sure docassemble doesn't print any of their sensitive information in the GitHub artifact. Also `echo` that that artifacts are turned off and how to turn them on.
2. Instruct users to put their sensitive config strings in their GitHub secrets so that the information doesn't show in the job console either.
3. [By default], silence the job console docker output completely. `echo` how they can turn those logs on.

Edit:

Docs: because your config is a github secret, you won't be able to see it's value at all as GitHub will replace it with "\*\*\*" even if you choose to show the logs. We'd need to implement this: ~You can use [an action "variable"](https://docs.github.com/en/actions/learn-github-actions/variables) for your organization or your repository if you don't want the config to be hidden by github.~

Maybe a question for a new issue: Instead of making an artifact, return the text of the docker log as an output? Still, avoid doing it by default, at least before turning off the logs, as that will show it in the console - whenever any bash command is run, all the output values are shown as env vars for some reason. I am concerned, though, that this output will show in the dev's own code unintentionally.

plocket:
In this PR, I have:

### Reason for this PR

Avoid logging docker output for to be thorough with privacy/security. Only people with admin/write/etc permissions can see the action's console output anyway, but we have users that still have concerns about log outputs.

[Don't bother looking at individual commits - developing actions is messy - but I hope the commit as a whole stands on its own.]

Closes #819

Manual workflow runs in actions

### DIFF
--- a/.github/workflows/github_server.yml
+++ b/.github/workflows/github_server.yml
@@ -12,6 +12,11 @@ on:
         description: 'Run the build with tmate debugging enabled (https://github.com/marketplace/actions/debugging-with-tmate)'
         required: false
         default: false
+      show_docker_logs:
+        required: false
+        default: false
+        type: boolean
+        description: 'Show the docker logs while building the GitHub server container. It will also save the docker log artifact. This might show sensitive config information.'
 
 jobs:
 
--- a/.github/workflows/github_server.yml
+++ b/.github/workflows/github_server.yml
@@ -36,21 +36,23 @@ jobs:
       # Place the root directory in this branch
       - uses: actions/checkout@v3
 
-      - name: Run ALKiln - Start the temporary GitHub server
+      - name: ALKiln - Start the isolated temporary docassemble server on GitHub
         id: github_server
         #### Developer note: you'll need to replace `.` with the path to
-        #### the repo and branch. For example
+        #### the repo and branch. For example:
         #### suffolkLITLab/ALKiln@v5/action_for_github_server
         uses: ./action_for_github_server
         with:
           CONFIG_CONTENTS: "${{ secrets.CONFIG_CONTENTS }}"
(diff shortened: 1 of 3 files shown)


.github/workflows/github_server.yml
```diff
@@ -7,6 +7,12 @@ on:
       tags:
         required: False
         description: 'Optional. Use a "tag expression" specify which tagged tests to run (https://cucumber.io/docs/cucumber/api/#tag-expressions)'
+      show_docker_output:
+        required: false
+        default: false
+        type: boolean
+        description: 'Show the docker logs while building the GitHub server container. It will also save the docker log artifact. This might show sensitive config information.'
+      #### Developer note: You probably don't need this
```
plocket, line 15: In our documentation we're considering advising our users to look at this file as an example of how to make their own workflow file. They'll need to customize some things. It may also help them to understand some details.

If they don't look at it, these notes may help us advise them on how to create their own file.

action_for_github_server/action.yml
```diff
@@ -53,20 +57,72 @@ runs:
           echo "DA_ADMIN_PASSWORD=$DA_ADMIN_PASSWORD" >> "$GITHUB_OUTPUT"
         shell: bash
 
-      - name: "ALKiln github_server: Download and start docker"
+      - name: "ALKiln github_server - Get config"
         run: |
+          # Info: config values
```
plocket, line 62: It turns out that this is what shows up in the job's console logs, not the name. Go figure.

Issue #827, opened 2023-12-05
dainstall doesn't install dependencies anymore


plocket:
docassemblecli changed dainstall a couple weeks ago and that's why our 3-parts dates tests are failing in the isolated GitHub server tests. The other AL tests are probably failing too. Maybe it's just getting buried in the noise.

Version 0.0.14 [changes when it restarts the server](https://github.com/jhpyle/docassemblecli/compare/v0.0.13...v0.0.14#diff-b29de36dc7d11398ab979dcf8683952f00ca0e53846c2e1545bff7fafc2cddd4). [I'm not sure what about it caused this issue, but changing my version of docassemblecli is what recreated the issue on my local server.] If there's no python file in the package being installed, it doesn't restart the server. It doesn't check for dependencies, so if there are dependencies, they don't get installed till the next restart/reload.

---

Old issue for posterity:

Topic: Running too fast?

[Sometimes our tests fail](https://github.com/SuffolkLITLab/ALKiln/actions/runs/7091525179) because the ThreePartsDate custom fields aren't done adding themselves to the page (I think): the error says that the element is not a `<select>` element and the pic shows a regular text input field. I checked manually on the testing and dev server, and the fields look fine. [This is happening in our GitHub server tests, which might be running a faster server than we're used to.]

I'm reluctant to slow the tests down more, especially since these lag issues are so unpredictable and we're already waiting the for a bit over the length of animation on at least `show if` fields. Maybe we can just slow down tests for custom datatypes? Maybe we just add more info to the ThreePartsDate error message, but I'm not sure how much that would help.

plocket:
~Maybe we'll just slow down date steps... [at least for now]~ Didn't work because that wasn't the problem. Read above.

BryceStevenWilley:
> Version 0.0.14 changes when it restarts the server. [I'm not sure what about it caused this issue, but changing my version of docassemblecli is what recreated the issue on my local server.] If there's no python file in the package being installed, it doesn't restart the server. It doesn't check for dependencies, so if there are dependencies, they don't get installed till the next restart/reload.

That's fairly frustrating. IMO, that's an upstream bug; we don't know which dependencies would get installed from `dainstall`, so it should usually restart, or at least should let the server tell us if we should restart. Happy to help raise this issue upstream or try to fix it. Until then, we should pin to version `0.0.13` to avoid that behavior.

plocket:
[...]

PR #838, opened 2023-12-18
Add Step to click without navigating. Use puppeteer click again as

### LINKED ISSUE #834: Step: I tap the element using selector (without navigating)
Currently that Step always expects navigation. [We've opted not to making a breaking change, but instead tweak the Steps a bit.]

~What I really want to do is change that Step to not navigate and make a new step that ends with "to navigate" or "to go to another/the next page".~

~This would be a breaking change, but I want to check with folks whether they use that step yet at all. If not, we can possibly change that step so it doesn't navigate without changing major versions. Otherwise we have to make a new Step with extra text that I don't appreciate.~

This would be a good opportunity to move "waiting to navigate" into its own function in scope.js. Right now it's in the `AfterStep` function.

- [x] ~check with all the users we know whether they're currently using that Step to navigate.~

plocket:
regular .click() supposedly only works with input elements, according to stackoverflow, and it seems to be working. We'll see if inconsistent behavior comes back. I can't find it mentioned online in the context it used to be mentioned.

Close #834

**[Note: The tests are breaking because of dainstall. See #827]**

In this PR, I have:

### Reason for this PR

See #834

See #834

N/A

### DIFF
--- a/.github/workflows/playground.yml
+++ b/.github/workflows/playground.yml
@@ -54,3 +54,5 @@ jobs:
           #### Developer note: Useful for temporarily shortening ALKiln tests
           #### to debug the docker install step
           # ALKILN_TAG_EXPRESSION: @a_tag_expression
+          #### Developer note: you probalby don't need this
+          ALKILN_VERSION: 5.7.0-feature-tap1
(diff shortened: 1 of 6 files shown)


BryceStevenWilley (approved):
`tapTab` stuff looks okay to me!

PR #841, opened 2023-12-21
Close #827 Freeze docassemblecli version at 0.0.17

### LINKED ISSUE #827: dainstall doesn't install dependencies anymore

plocket:
Use custom docassemblecli repo (rejected)
Use custom docassemblecli in the old action as well (rejected) Try new version of docassemblecli
Use pip instead of pip3
Close #827 Freeze docassemblecli version at 0.0.17

In this PR, I have:

* [x] ~Added tests for any new features or bug fixes (if relevant)~

### Reason for this PR

docassemblecli wasn't installing dependencies on the given server. It got fixed upstream and we're freezing at this version to have better control over what changes we pull in.

Closes #827

Tested `dainstall` locally, but most of the testing was in GitHub as it is the environment we're concerned about.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -31,6 +31,10 @@ inputs:
   ALKILN_TAG_EXPRESSION:
     description: "The tag expression to use to limit which tests to run. Default is no tag expression."
     required: false
+  DOCASSEMBLECLI_VERSION:
+    description: 'Where to get the docassemblecli package. E.g. a GitHub zip or a pypi version.'
+    required: false
+    default: "0.0.17"
 
 runs:
   using: "composite"
@@ -52,6 +56,7 @@ runs:
         echo "ALKILN_VERSION=${{ inputs.ALKILN_VERSION }}" >> $GITHUB_ENV
         echo "MAX_SECONDS_FOR_SETUP=${{ inputs.MAX_SECONDS_FOR_SETUP }}" >> $GITHUB_ENV
         echo "SERVER_RELOAD_TIMEOUT_SECONDS=${{ inputs.SERVER_RELOAD_TIMEOUT_SECONDS }}" >> $GITHUB_ENV
+        echo "DOCASSEMBLECLI_VERSION=${{ inputs.DOCASSEMBLECLI_VERSION }}" >> $GITHUB_ENV
         echo "_ORIGIN=github" >> $GITHUB_ENV
       shell: bash
     - name: "ALKiln setup info: confirm environment variables"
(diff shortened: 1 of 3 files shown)


action.yml
```diff
@@ -31,6 +31,10 @@ inputs:
   ALKILN_TAG_EXPRESSION:
     description: "The tag expression to use to limit which tests to run. Default is no tag expression."
     required: false
+  DOCASSEMBLECLI_VERSION:
+    description: 'Where to get the docassemblecli package. E.g. a GitHub zip or a pypi version.'
```
BryceStevenWilley, line 35: Fairly certain this syntax `pip install docassemblecli==$DOCASSEMBLCLI_VERSION` will only work with specific pypi versions. You'd to wrap the whole thing in quotes for it to work with `@ https://github.com/.../archive/main.zip`.  https://pip.pypa.io/en/stable/reference/requirement-specifiers/ should have the different things `pip` accepts.

To be clear, I think restricting to pypi versions of the package is fine, but I just think that should be reflected in the description

PR #850, opened 2024-02-04
Close #582, implement 2-column Story Tables with no trigger var

### LINKED ISSUE #452: Only show the missing trigger variable warning for Story Table Steps
The trigger variable only relevant for Story Table steps trying to identify unique pages for fields that use `index variables` or `generic objects` (see #256). In those cases, the developer needs to add some special lines in their interview to give us that clue. We've written a warning for whenever the developer leaves the trigger variable out of their interview, but it shows up all the time, not just when it would matter.

Ideas for when the warning of a missing trigger variable should be added to the report:
1. Whenever there's a Story Table Step.
2. Whenever a Story Table has any rows that use index variables or generic objects.
2. Whenever a Story Table row uses an index variable or generic object.
3. Whenever a Story Table row includes something in its "trigger" column.
4. ~Whenever a trigger variable is present in the page.~ Rejected because a trigger variable would be present on every page of an interview, especially if Assembly Line is included, not just those that really need them.
4. A combination of the above.

Regardless, it shouldn't show up for other types of Steps, even if they set variables. The page where the value is meant to be set is already clear in that situation.

I'm reluctant to pass another parameter into `scope.setFields()`, but that may be the only way to do it.

Edit: See docs for trigger var: https://suffolklitlab.org/docassemble-AssemblyLine-documentation/docs/automated_integrated_testing/#a-missing-trigger-variable

### LINKED ISSUE #573: Simplify warning about missing trigger var
> WARNING: You are missing an element containing the data for the page's trigger variable. Some fields, like loops with index variables, may not work correctly in these tests. See https://suffolklitlab.org/docassemble-AssemblyLine-documentation/docs/automated_integrated_testing/#a-missing-trigger-variable.

We can get reduce the full explanation now that we have a link to the documentation

[Different from #452 - this one is about simplifying, not removing.]

### LINKED ISSUE #582: Remove the need for `x` in the variable name column
Realized today that we can simplify the story table a lot by providing us with more information from docassemble.

We need to trigger value because generic objects don't show all of their information when being sought, just the `x.sought_attribute`. However, if we are on a generic object screen, we can just directly include the `x.instanceName` in the DOM, the same way we include the `user_info().variable`:

```html
    <div data-variable="${ encode_name(str( x.instanceName if defined('x') else '' )) }" 
        id="al_generic_object" aria-hidden="true" style="display: none;"></div>
    <div data-variable="${ encode_name(str( i if defined('i') else '' )) }" 
        id="al_generic_index" aria-hidden="true" style="display: none;"></div>
```

This gives us the exact name of the `x` object (`x.instanceName` must be set in order for DA to seek it). This means that Kiln can take this value and substitute it with the `x.sought_attribute` on the page, giving the correct value of `my_var.sought_attribute`.

Much of the same above goes for `i` and `j` as well.

The only potential historical issue that relates to this that I can find is https://github.com/SuffolkLITLab/ALKiln/issues/158#issuecomment-801442823, but I don't think that affects this?

Need to think through all of the combinations of things as well, but this does seem promising.

plocket:
In this PR, I have:

### Reason for this PR

Allow authors to avoid proxy variables and their complications in Story Tables. This means a 2-column layout where the `var` column never contains `x`, `i`, etc, just the name of the actual variable that needs to be set.

It does require the author to put new HTML in their interview:

```html
<div id="alkiln_proxy_var_values"
      data-generic_object="${ encode_name(str( x.instanceName if defined('x') else '' )) }"
      % for letter in [ 'i', 'j', 'k', 'l', 'm', 'n' ]:
      data-index_var_${ letter }="${ encode_name(str( value( letter ) if defined( letter ) else '' )) }"
      % endfor
      aria-hidden="true" style="display: none;"></div>
```

Closes #582 - 2 column table
Closes #573 for now, shortening the message a little. We'll work more on error messages another time.
Closes #452 for now. Reduces the number of times we show this error message.

Also:
- Add name of Scenario to progress dots. May help with logging progress on GitHub
- Add internal test should fail even when the report is correct (a previously undiscovered bug)
- Allow "expected status" tests to accept more than 'passed' and 'failed'
- Added doc explaining what results we expect
- Fix failures because [upstream Assembly Line change](https://github.com/SuffolkLITLab/docassemble-AssemblyLine/blob/f9bd864990f8bc2c0c0096700ba6ad647772cb83/docassemble/AssemblyLine/al_general.py#L1008) added punctuation

None on the final product


lib/scope.js
```diff
@@ -308,14 +311,11 @@ module.exports = {
     } catch ( error ) {
 
       if ( error.name === `TimeoutError` ) {
-        let non_reload_report_msg = `ALKiln tapped "${ tapperText }", but the interview took to long to finish.`;
+        let non_reload_report_msg = `After ALKiln tapped "${ tapperText }" it took too long to load the next page.`;
```
plocket, line 314: Just an error message tweak

lib/scope.js
```diff
@@ -292,6 +292,9 @@ module.exports = {
     if ( !elem ) { reports.addToReport(scope, { type: `error`, value: msg }); }
     expect( elem, msg ).to.exist;
 
+    // Get the text on the button or link for a potential error message later
+    let tapperText = await (await elem.getProperty('textContent')).jsonValue();
+
```
plocket, line 297: Basically unrelated to the new feature

lib/scope.js
```diff
@@ -308,14 +311,11 @@ module.exports = {
...
-        let non_reload_report_msg = `ALKiln tapped "${ tapperText }", but the interview took to long to finish.`;
+        let non_reload_report_msg = `After ALKiln tapped "${ tapperText }" it took too long to load the next page.`;
         await scope.handle_page_timeout_error( scope, { non_reload_report_msg, error });
       } else {
-
-        // Get the text on the button or link for a potential error message later
-        let tapperText = await (await elem.getProperty('textContent')).jsonValue();
         // Throw any non-timeout error
-        let err_msg = `Error occurred when ALKiln tried to tap "${ tapperText }".`
+        let err_msg = `An error occurred when ALKiln tried to tap "${ tapperText }".`
```
plocket, line 318: Unrelated

lib/scope.js
```diff
@@ -815,17 +871,22 @@ module.exports = {
...
-    if ( !trigger_elem.length ) { trigger_elem = $( `#sought_variable` ); }  // Backwards compatibility
-    let encoded_trigger_var = $( trigger_elem ).data( `variable` );  // This should come out to 'None' at the very least
-    if ( !encoded_trigger_var ) {
-      // TODO: Add documentation and then link to it in this message.
-      reports.addToReport(scope, { type: `warning`, value: `You are missing an element containing the data for the page's trigger variable. Some fields, like loops with index variables, may not work correctly in these tests. See https://suffolklitlab.org/docassemble-AssemblyLine-documentation/docs/automated_integrated_testing/#a-missing-trigger-variable.` });
-      encoded_trigger_var = toBase64( `None` );
-    }
+    let $trigger_elem = $( `#trigger` );
+    if ( !$trigger_elem.length ) { $trigger_elem = $( `#sought_variable` ); }  // Backwards compatibility
+    let encoded_trigger_var = $( $trigger_elem ).data( `variable` ) || ``;
```
plocket, line 831: New 2-column feature actually starts.

lib/scope.js
```diff
@@ -724,13 +721,58 @@ module.exports = {
...
+  //           var_data.push({var: row_raw[0], value: row_raw[1], trigger: row_raw[2]});
+  //         }
+  //       }
+  //     }
+  //   }
+
+
+
+  // },
+
```
plocket: Delete this

lib/scope.js
```diff
@@ -1040,41 +1113,69 @@ module.exports = {
...
+        field_like_names,
+        is_none_of_the_above,
+        proxies
+      })
+    );
+
+    return guesses;
+  },  // Ends scope.getPossibleDecodedNames()
+
+  decodeFully: function ( scope, {
```
plocket, line 1117: So many arguments! Some of which are drilling so far down!

lib/scope.js
```diff
@@ -1118,30 +1234,77 @@ module.exports = {
...
-      possible.push(`${ expr_str }`);
+      possible = possible.concat( scope.with_and_without_proxies( scope, {
+        var_name: `${ expr_str }`,
+        proxies,
+      }) );
     }
     return possible;
   },
 
+  with_and_without_proxies: function ( scope, { var_name, proxies={} }) {
```
plocket, line 1200: This is the meat of the new feature that I've started calling "proxy substitution" - it replaces `x`, `i`, `j`, etc. (proxies) in the field names and values (if the proxies exist there) with the actual values of the proxies (if the author included those values in the HTML).

lib/scope.js
```diff
@@ -845,62 +906,62 @@ module.exports = {
...
-        });
+        not_decoded_name = first_sib.attribs.name;
+        // TODO: Do we need daaota as well? ("all of the above")
       }
 
+      let var_names = await scope.getPossibleDecodedNames( scope, {
+        not_decoded: not_decoded_name,
+        field_like_names,
+        is_none_of_the_above: is_none_of_the_above,
+        proxies,
```
plocket, line 917: This is where the drilling starts, and it drills so far down!

lib/scope.js
plocket: Pointed to the start of the drilling in a comment below

plocket:
The Playground test is failing because of the feature test with "Sr", but that test no longer exists on this repo. Wat.

PR #856, opened 2024-02-22
854 arbitrary url

### LINKED ISSUE #854: Allow authors to go to completely arbitrary urls
Our current recommendation to go to a completely arbitrary url is to use the "I start the interview at" Step. Unfortunately, it checks that the page is a da interview page. Some authors have change their pages to customize them and they don't contain things like `.daquestion` (`#daquestion?`) that we use to check that it's a da page. For example, the log in page at https://dev.pleadingpower.com/user/sign-in.

We could look for other docassemble elements, but it feels like it's just pushing off the inevitable. Tests won't be able to do very much on those pages if they don't use some da conventions, like the kinds of input field DOM structures we're expecting, but they can do some things, like check for phrases.

plocket:
In this PR, I have:

### Reason for this PR

Initially it was to make the `I go to "<url>"` Step to go to an arbitrary url. It morphed into just expanding some behavior to allow PleadingPower to pass tests. Probably stuff we needed anyway, but probably not stuff that should be in a PR all together. The new Step isn't really necessary any more for this particular purpose, but I've been meaning to add it.

All changes should be listed in the changelog.

N/A

The only way I was able to test the more robust addition to the trigger variable name flow was with PleadingPower. We don't have a server that has the necessary restriction (config `restrict input variables=True`), so we don't have a way to test handling that restriction.


plocket:
[Stackoverflow question from today about chrome failing to download half the time](https://stackoverflow.com/questions/78042000/how-to-fix-puppeteer-intermittently-failing-when-downloading-chrome-headless-she). No answers yet.

PR #874, opened 2024-05-03
Close #873, add log codes for better tests, docs, and communication

### LINKED ISSUE #873: Add log codes for errors, warnings, info, successes, and debug messages
### Reasons

Basically, each log will have something that is consistent and distinct about it and that gives advantages:

1. It can make tests more robust and maintainable. Instead of checking for the text of the log (which we change over time to increase clarity), our internal tests can check for the code. We decrease the amount of times we need to update tests.
2. It can make logs easier to document for the same reason.
3. It can make communication with users easier. They have a piece of information that is very clear and specific to look up in the documentation or to pass on to us.

### Details

- Each code will start with `ALK` and be followed by 4 digits. For example, `ALK0023`.
- The numbers will simply be sequential, regardless of the type of log. The research we've done indicates that categories are too fuzzy and change too much over time.
- The reports should show numbers for errors or warnings, but we'll have to discuss more about the other types of report items. For example, variable row descriptions shouldn't have codes in the report.
- The logs might start with emojis just before the code.
   - A code like that can look alarming with no context and users might assume that every log with a code is an error (so, all of them). If we include a clear visual hinting at the type of log, that can immediately show what's going on.
   - The emojis should be as universal as we can make them. Avoid emojis like 👍 because those are very different between cultures and countries. Facial expressions are better, but if we use faces for them all, it will be hard to tell them apart from each other with a glance.
   - The emojis should avoid increasing blood pressure. An ❌ is stressful to see.
   - [The emojis should be supported by the systems that are going to display these messages. We'll see how that goes.]
- Include the text description of the log - ERROR, WARNING, INFO, SUCCESS, or DEBUG (English for now).
- Include any additional info about where the message came from. E.g. "setup", "run", "takedown".
- Then comes the message and possibly additional data.

### Supplementary tools

Additionally, it's hard to keep track of what new code to add next, so include a way to:
   - Get the highest current code number
   - Test that we haven't duplicated any codes
   - Test that we haven't skipped any codes

Add these to our automated internal tests to catch mistakes.

plocket:
In this PR, I have:

### Reason for this PR

Add log codes. Also:
- Add automatic log code validation and information script.
- Update action and workflow dependencies.
- Fix inaccurate log on failure for docker container setup.
- Neutralize internal a11y tests, which are failing again because of changes to docassemble (that ironically actually are increasing accessibility!). We will trigger those tests manually from now on. This does make testing manually tricky now because I think we have to use a hack like `not @none` if we want to test all of the tests at once, but I didn't have time to troubleshoot that one.

Closes #873

Ran the new code-validating script locally to make sure it catches different cases.

### Additional notes:

GitHub actions are a pain. https://rhysd.github.io/actionlint/ is helpful, as is tmate, but oh boy.

### DIFF
--- a/lib/run_cucumber.js
+++ b/lib/run_cucumber.js
@@ -51,7 +51,7 @@ function build_config( command_args ) {
   /** Create the cucumber config, run the tests, create and
   *   manipluate the output. */
   const environment = { cwd: process.cwd() };
-  console.log("alkiln-run: environment: %o", environment);
+  console.log(`ALK0@@@ alkiln-run: environment: %o`, environment);
 
   // We'll use the user info to construct the possible paths to /sources
   let id = session_vars.get_user_id();
@@ -86,11 +86,11 @@ function build_config( command_args ) {
     provided: build_config( argv ),
   }, environment);
 
-  console.log("alkiln-run: runConfiguration: %o", runConfiguration);
+  console.log(`ALK0@@@ alkiln-run: runConfiguration: %o`, runConfiguration);
 
-  var path = require('path');
+  var path = require(`path`);
   var codeDir = path.dirname(require.main.filename);
-  console.log("alkiln-run: working directory: %o", codeDir);
+  console.log(`ALK0@@@ alkiln-run: working directory: %o`, codeDir);
   const support = await cucumber_api.loadSupport({ support: {
     requireModules: [ `${ codeDir }/index.js` ],
     requirePaths: [ `${ codeDir }/index.js` ],
@@ -100,7 +100,7 @@ function build_config( command_args ) {
   console.log(`\n\n`);
 
   const { success } = await cucumber_api.runCucumber({...runConfiguration, support});
-  console.log("alkiln-run: success: %b", success);
+  console.log(`ALK0@@@ alkiln-run: success: %b`, success);
   const exitCode = success ? 0 : 1;
   process.exitCode = exitCode;
 
@@ -109,7 +109,7 @@ function build_config( command_args ) {
   let summary_file_name = 'cucumber-report.txt';
   fs.readFile(summary_file_name, 'utf8', (err, data) => {
     if (err) {
-      console.error(err);
+      console.error(`ALK0@@@ ERROR: cucumber error\n`, err);
       return;
     }
 
--- a/lib/run_cucumber.js
+++ b/lib/run_cucumber.js
@@ -79,33 +79,35 @@ function build_config( command_args ) {
   let one_dir_exists = session_vars.set_sources_paths( sources_list );
   if ( !one_dir_exists ) {
     // If we don't throw, it'll be more confusing when no tests run
-    log.error({ type: `run`, code: `ALK0049`, pre: `ALKiln can't find your "sources" folder(s). See above for details.` });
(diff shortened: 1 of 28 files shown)


lib/run_cucumber.js
```diff
@@ -79,46 +79,52 @@ function build_config( command_args ) {
   let one_dir_exists = session_vars.set_sources_paths( sources_list );
   if ( !one_dir_exists ) {
     // If we don't throw, it'll be more confusing when no tests run
-    throw new ReferenceError( `ALKiln ERROR: ALKiln can't find your "sources" folder(s). See above for details.`);
+    log.error({ type: `run`, code: `ALK0049`, pre: `ALKiln can't find your "sources" folder(s). See above for details.` });
+    throw new ReferenceError(`ALK0050 ERROR: ALKiln can't find your "sources" folder(s). See above for details.`);
```
plocket: These were supposed to be the same error code. Gorramit. We _could_ re-use one of those error codes at some future point. That's going to be hard to keep track of. We could just leave it in the dust because the other option is to re-number _all_ the error codes that come after. I may just be too compulsive to avoid it...

lib/scope.js
```diff
@@ -2082,12 +2118,16 @@ module.exports = {
     * @param ensure_all_vars_used {bool} [false] = Default false. If true, errors if any
     *     object in var_data is not used during running of this function.
     */
-    if ( session_vars.get_debug() ) { console.log( `~~~~~~~~~~~~~~ scope.setFields() ~~~~~~~~~~~~~~` ); }
+    log.debug({
+      code: `ALK0139`,
+      type: `info`,
+      pre: `~~~~~~~~~~~~~~ scope.setFields() ~~~~~~~~~~~~~~`
+    });
```
plocket: We added `log.console()` for plain old pretty stuff and we should just use that here.

lib/steps.js
```diff
@@ -418,149 +424,33 @@ When(/^(?:the user gets)?(?:I get)? to ?(?:the)?(?: ?question)?(?: id)? "([^"]+)
...
+        reports.addToReport(scope, { type: `row info`, code: `ALK0068`, value: row.report_str, });
       }
     }
   } else {
-    reports.addToReport(scope, { type: `note`, value: `${ ' '.repeat(2) }All rows were used` });
+    reports.addToReport(scope, { type: `note info`, code: `ALK0069`, value: `${ ' '.repeat(2) }All rows were used` });
   }
 
-  reports.addToReport(scope, { type: `note`, value: `` });
+  reports.addToReport(scope, { type: `note info`, code: `ALK0070`, value: `` });
```
plocket: This is cosmetic and doesn't need a code (it's just adding a new line).

plocket: Then again, we don't have another simple way to add a new line to the report. When we loop through unused rows, we don't know which will be the last line in order to add a `\n` to it. The report is going to add a code to every line in the debug, and this code is meaningless.

lib/run_cucumber.js
plocket: Fixed

lib/scope.js
plocket: That function was silly. This and related stuff fixed.

lib/steps.js
plocket: Fixed

Issue #461, opened 2021-12-29
Create a report for setup or takedown failure (break reports into their own file?)


plocket:
This can happen when their GitHub secrets are wrong, when their server is down, when their interview filename is wrong, etc, so it would be good to let them know what went wrong.

Right now, the developer doesn't get any hints in the artifacts when setup or takedown fails. They would have to go into the GitHub job display and read, and understand, what's going on in there, and that would be an obstacle for many of our target audience.

It may be time to break reporting functionality into its own thing.

We're adding a "log.js" file in v4. Should reports and logs be combined into one file, or are they distinctly separate? When we want to add something to log something, we often want to add exactly that same information to the report.

PR #948, opened 2024-10-02
Make a proper Log class and methods

### LINKED ISSUE #659: Abstract adding to debug_log.txt?
Re-implement this function if it's useful. Maybe it would just be used in `addToReport()`, but maybe we could also then use it elsewhere.

```js
log.add_to_debug_log = function(value) {
  let artifacts_path = session_vars.get_artifacts_path_name();
  fs.appendFileSync(`${ artifacts_path }/${ log.debug_log_file}`, value);
};
```

[We could also just do this in `log.debug()`.]

### LINKED ISSUE #805: Create verbose log where we log absolutely everything
We'd need to replace all the `console.log`s with a function that logged to the console, but also saved in the verbose debug doc.

- [ ] Label each log with its purpose and/or placement
- [ ] Still log to the console
- [ ] Log cucumber output

~~It'd also be nice (but not critical) to have nice indentation that matches, but that might be handled in a different issue.~~ [This would be more appropriate for reports.]

### LINKED ISSUE #924: Remove url that prints with "successfully loaded interview"
That info is redundant. It's already printed with "trying to load" and it's very noisy.

### LINKED ISSUE #925: Decision: Allows log to throw an error
Give a Log instance the ability to throw an error with `log.throw()`. This will replace `log.error()`. There's no reason to log an error we're not throwing and will make sure that errors get saved to the debug log.

~~If we want to manipulate the stack trace to show a better trace for the error: https://stackoverflow.com/a/42755876~~ [You can't manipulate the stack trace that way. If you pass in an error, though, I believe it will keep the stack trace it had where it was created.]

plocket:
Apparently there was still a while to go. I still have decision and design/architecture docs to write, but this is ready for review. What I most want to discuss is how to handle bugs internal to `.debug()`.

I wish this integrated the report more too, but this is already big enough.

In this PR, I have:

### Reason for this PR

The logs have been a bit of a mess for a while. The verbose logs could only be logged live, not saved. Saving to the debug log was a bit add-hoc. There wasn't a good differentiation between console logs and debug logs. This addresses a few of those problems.

Close #805, create verbose-type log
Close #924, remove extra url print
Close #659, abstract adding to debug_log
Close #925, allow a `Log` to throw an error
Address #461, setup and takedown reports

Visually checked various log files.


docs/decisions/006_which_log_level_to_pick_2024_09_5.md
```diff
@@ -0,0 +1,46 @@
...
+`success`:
+  - Only log success once for success of all tests.
+  - Include the `success` icon in the Scenario "Passed" section of the report.
+`throw`:
+  - Use this level to throw errors for the user's tests (or our own internal tests).
+  - Use this if ALKiln itself needs to abort a test or the test suite.
+`error`:
+  - Avoid using `error` explicitly. Only the debug logs should use the `error` level, and only to record critical errors that come in with `throw`.
+`info`:
+  - Log expected errors as `info`. For example, the aborted promises when awaiting navigation.
```
BryceStevenWilley: Really interesting rule for what to use when. Trying to sit with my biases of what INFO is usually used for, and determine if it's worth renaming INFO to something else, like "suspicious" (SUS). Something that, by itself, can be ignored (and doesn't need to be investigated), but with a failure, might be an issue. Usually I'd use warn for that, but I do like warn being only for user focused logs

lib/docassemble/server_install.js
```diff
@@ -23,20 +24,33 @@ const server_install = async () => {
...
+    });
   }
 
   if ( folders.length > 1 ) {
-    log.warn({
-      type: `setup`,
-      code: `ALK0045`,
-      data: `There are ${folders.length} folders at ${path}: "${folders.join('", "')}". ALKiln chose to use the folder "${chosen_folder}".`
-    })
+    log.warn({ code: `ALK0045`, context: `setup`, },
```
BryceStevenWilley: Still working through things, but should `context` be changed to `server_install` too?
```suggestion
    log.warn({ code: `ALK0045`, context: `server_install`, },
```

docs/decisions/reports_and_logs_2024_08_11.md
```diff
@@ -0,0 +1,141 @@
...
+- Create a complex and heavy system to try to abstract absolutely everything
+
+## The actual design
+
+This design still seems a little complex and potentially too coupled with reports.
+
+### System-context diagram
+
+```
+logic code -> logger -> report -----------> console
```
BryceStevenWilley: I get what these diagrams are trying to show, but why are there 3 different ones? Are those options considered?

lib/docassemble/docassemble_api_interface.js
```diff
@@ -61,6 +61,10 @@ const time = require('../utils/time');
 let da_i = {};
 module.exports = da_i;
 
+const artifacts_path = session_vars.get_artifacts_path_name();
+const log = new Log({ path: artifacts_path, context: `api` });
+const context = `da api`;
```
BryceStevenWilley, line 64:
```suggestion
```

Unused variable

lib/docassemble/docassemble_api_interface.js
```diff
@@ -61,6 +61,10 @@ const time = require('../utils/time');
 let da_i = {};
 module.exports = da_i;
 
+const artifacts_path = session_vars.get_artifacts_path_name();
+const log = new Log({ path: artifacts_path, context: `api` });
```
BryceStevenWilley: (haven't gotten through log.js yet, but) `log` is always called with `context: "da api"` in this file; should `context` here just be `da api` here, and that param can be removed from each individual call?

docs/decisions/reports_and_logs_2024_08_11.md
```diff
@@ -0,0 +1,141 @@
...
+  - `AbortError`s of promises that should indeed have been aborted.
+- Should we allow "in-between"s for all parts of the log? All together: pre-everything (new lines, dividers, etc), icon/context/etc., pre-logs, logs (the actual messages to print), pre-data, data, post-data? We can default to nothing for all of those.
+- How do we differentiate between methods that log to the console vs. methods that just store data?
+
+[^1]: Don't go down this rabbit hole. Broadly - with the "Playground" version, users see the results as a web page instead of in the GitHub job console or in the artifacts. The users run the tests on their own server. This has to do with the platform we built ALKiln for - [docassemble](https://docassemble.org/). It gives users a faster iteration cycle.
+
+## Non/anti-goals
+
+- Spend a month on this
+- Create a complex and heavy system to try to abstract absolutely everything
```
BryceStevenWilley, line 25: An additional non-goal that I think you might have; logs that are only understandable if you read the surrounding code. A lot of things can be simplified if you assume that, but it def raises the effort level needed by users quite a bit

BryceStevenWilley (commented):
Some initial thoughts. Didn't get a chance to look through everything, but focused on the design docs and how the new log API is used in Kiln itself.

plocket:
> Didn't get a chance to look through everything, but focused on the design docs and how the new log API is used in Kiln itself.

Don't worry about looking at the code implementation for now. I'm doing a fair amount of refactoring. Thanks, @BryceStevenWilley!

docs/decisions/006_which_log_level_to_pick_2024_09_5.md
plocket: I like this line of thought. No need to reply if you don't have bandwidth, I'll just brainstorm before.

Non-alarming words that can have an internal meaning:
- note
- detail
- cue
- queue
- data
- trace
- dig
- finding
- memo
- review
- intel
- mark
- comms
- signal
- task
- todo

Somewhat alarming words that can have an internal meaning:
- suspicious
- sus
- alert
- notice
- clue
- mark
- symptom
- advise
- caution

We could also require using "info" for things that don't need attention and leave "debug" as our internal signal to examine further. Or maybe, going in a different direction - just "alkiln"/"alk".

docs/decisions/reports_and_logs_2024_08_11.md
plocket: Yes, alternatives. I'm going to have to make some adjustments to this doc and maybe break it out into separate docs.

lib/docassemble/docassemble_api_interface.js
plocket: I think I understand what you mean and I like it. Though I wish I could make it all caps so it could look like a proper constant. That would somewhat defeat the purpose, though.

lib/docassemble/docassemble_api_interface.js
plocket: Took the other avenue suggested in https://github.com/SuffolkLITLab/ALKiln/pull/948/files#diff-227ae832372018230aee8c35c6536e1bbfbf40bd225b35079529a93f6139713e.

docs/decisions/006_which_log_level_to_pick_2024_09_5.md
plocket: Misread this note a bit. I think an alternative word (like "SUS") would be useful for internal warnings, but just want to be clear that an aborted promise is expected and is really just information - when one promise wins a race, the others will, of course, be (correctly) aborted. I added a design doc about the various meanings of "error" that might help clarify this a bit. I sort of want to combine that doc with this one, but I haven't seen how to make it clear.

plocket:
Ready, barring 1 main discussion point (and the docs). That one item - how to handle bugs internal to `.debug()`. They bring up the potential for infinite loops under very specific circumstances. The `max_loops` variable is a safety valve right now, but I hate it.

Thoughts:
1. I really don't like using `max_loops` or some other kind of cut-off. If we use it, we need to make it high enough that useful information doesn't get missed. That said, it just smells bad. Calling `.debug()` internally is just asking for trouble, even if we're calling it with what we believe to be safe arguments.
2. Option - collect the unexpected behavior messages as a list and, at the end, decide whether to save them to the debug log or to console log them. I'm not sure how to pass that data around, though.
3. Option - console log all unexpected behavior. I'm really reluctant to create that much noise for the authors when it's not-critical functionality.

While we mull over that, I'm going to keep plugging away at the design/decision docs.

plocket:
My current experiments to eliminate inner calls to `.debug()` is to replace that call to use a function that abstracts saving straight to the debug file. [If that's impossible, I think logging to the console is a reasonable option at that point.]

docs/decisions/reports_and_logs_2024_08_11.md
plocket: Ongoing struggle to improve clarity, I think. I've (unreasonably optimistically) made a [discussion category for improving quality](https://github.com/SuffolkLITLab/ALKiln/discussions/categories/improving-clarity) for avoiding making tons of issues while getting to put down thoughts about clarifying logs, other output, and maybe documentation (though we do have the repo for the [Assembly Line documentation website](https://assemblyline.suffolklitlab.org/docs/) for that last one probably).

PR #959, opened 2024-11-21
Fix #725, partial filename match for downloading fails

### LINKED ISSUE #725: File download Step with partial name doesn't get provided in output
[Allow ["download" Step](https://assemblyline.suffolklitlab.org/docs/alkiln/writing/#download) to accept partial matches on filenames. This is important because some filenames have random characters in them from docassemble, so an author can't always know what the full file name will be.]

Right now, since there is a way the Step can work (using the full name with the extension), this isn't a high priority, but it would be very nice to have this worked out.

In in-Playground tests, the first test at [Alaska ChidSupport defendant_paths.feature](https://github.com/A2JatAKCourts/docassemble-ChildSupport/blob/d283296a1d05cf7d0a5802618bce3c3991738754/docassemble/ChildSupport/data/sources/defendant_paths.feature) downloads a file. The test will pass with both a filename of `child_support` or `child_support_preview.pdf`, but a downloaded file will only be provided for the latter string. Not sure about GitHub tests yet.

Examples:
```gherkin
Scenario: ...
  # all the other Steps
  # This one doesn't download a file
  And I download "child_support"
  # This one does download a file
  And I download "child_support_preview.pdf"
```

The Step [is supposed to do] a partial match for the file name when looking for the document link. It's confusing that the tests would both pass, but one would correctly provide the file while the other doesn't. One theory I have is that somehow the download name of the file is affected and so it does get downloaded, but our code can't access it because it doesn't have an extension or something.

plocket:
The failure is actually during the comparison - the file was being saved using the partial name the author gave in the Step when it should have been using the file's downloaded name.

Close #725 

Note that our current download method will, in ideal cases, let an author re-name the downloaded file. Since it might not be consistent, we might still avoid implementing this at this point. Also, in ideal cases, the Step does already continue without waiting for the whole file to download, which addresses #492. That happened a while ago. It's worth considering closing that issue at this point. We have to research under what circumstances a file would fail to download with a `fetch` while still being able to download with a click. Maybe with an encrypted interview.

In this PR, I have:

### Reason for this PR

When using the "download" Step, authors use a filename to get the file link from the page. For example, "my_file.pdf". They are also supposed to be able to use a partial file name to match the file name - "my_fi" should work just as well. That seemed to cause problems, though.

Exploring the problem showed that the real problem was when comparing the downloaded PDF to a baseline PDF. An author would download "my_fi" and then compare "my_file.pdf" with "basline_file.pdf". The comparison would fail because "my_file.pdf" was missing.

The problem was that ALKiln was saving the downloaded file using the partial-match name, not the file's actual name. That is, ALKiln would save a file called "my_fi". Then, when ALKiln tried to find "my_file.pdf" for the comparison, it would only find "my_fi" and the comparison would fail.

This PR now uses the name that the download file gives in the `response.headers`.

Note: We might need to improve the message when ALKiln can't find the downloaded file. It wasn't clear about what happened. I might add that to the PR

Also see #725 

Close #725

N/A

### DIFF
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -2781,7 +2781,7 @@ module.exports = {
       await scope.afterStep(scope);
     },  // Ends scope.steps.sign()
 
-    download: async ( scope, filename ) => {
+    download: async ( scope, full_or_partial_file_href ) => {
       /* Taps the link that leads to the given filename to trigger downloading.
       *    and waits till the file has been downloaded before allowing the tests to continue.
       *    WARNING: Cannot download the same file twice in a single scenario.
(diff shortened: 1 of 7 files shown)


lib/scope.js
```diff
@@ -2781,7 +2781,7 @@ module.exports = {
       await scope.afterStep(scope);
     },  // Ends scope.steps.sign()
 
-    download: async ( scope, filename ) => {
+    download: async ( scope, filename_matcher ) => {
```
BryceStevenWilley: `filename_matcher` to me implies that it's some more powerful lambda function or regex or something, and not just a partial string. Maybe `partial_filename` would be better.

BryceStevenWilley (approved):
Looks good to me!  The variable name change is the only thing I see that could improve things

Issue #229, opened 2021-06-06
How to detect navigation?


plocket:
I'm not sure it's possible because sometimes the page reloads the same page before moving on. I think there's another issue for this somewhere. If this wasn't the case, we may be able to use the sought var to test for the uniqueness of the page.

BryceStevenWilley:
https://github.com/SuffolkLITLab/ALKiln/issues/190 in the other issue.

plocket:
Using `page.waitForRequest` and `page.waitForResponse` plus some info about the behavior of the internals has yielded some good results. We're also detecting different invalidation feedback in different ways. We're also experimenting with using puppeteer promises and `Promise.race` with newer puppeteer because it has `AbortController`s for some of their async operations, though not all, so it may get tricky. Here are some da dev suggestions, though I've implemented some of these already in my current efforts:

> A `daPageLoad` event is triggered when there is a GET request to `/interview` or `/start` or `/run`. It [`daPageLoad`] is triggered when a new screen is loaded. A new screen is loaded when the user clicks a Continue button or launches an action by clicking a hyperlink or equivalently when the `action_perform()` JavaScript function runs an action. The screen is also loaded when `reload` causes the screen to refresh. `/checkin` events do not load a new page; if they did the screen would reload every time someone changed the value of a field, and once every six seconds. However it is possible to call `background_response('refresh')` from `check in` code, or use `background_action()` with the second argument `'refresh'` to send a signal for the browser to refresh the screen, which would cause a `daPageLoad` event.
> 
> The browser is not told whether the screen it is showing is influenced by `validation code` resulting in an exception. If the `validation code` attached errors to fields, you could inspect the jQuery Validation Plugin using JavaScript. If it caused a message to appear at the top of the screen, you could query the DOM for the presence of an error message at the top of the screen.

If we can figure out how to abort waiting for the `daPageLoad` jQuery event, we can possibly use that alone. [Alternatively, we could make one listener for the whole session unless that listener gets destroyed. In order to detect that destruction, we may be able to use a named function in the listener and check `jQuery._data( document, "events" ).daPageLoad.includes( func_name )`.]

[...]

PR #1034, opened 2025-11-24
Fix error when getting tapper text

BryceStevenWilley:
There's a weird error that seems to happen because a new error in Chromium? 

```
ProtocolError: Protocol error (DOM.resolveNode): Node with given id does not belong to the document
           at <instance_members_initializer> (/opt/hostedtoolcache/node/18.20.8/x64/lib/node_modules/@suffolklitlab/alkiln/node_modules/puppeteer-core/lib/cjs/puppeteer/common/CallbackRegistry.js:93:14)
           at new Callback (/opt/hostedtoolcache/node/18.20.8/x64/lib/node_modules/@suffolklitlab/alkiln/node_modules/puppeteer-core/lib/cjs/puppeteer/common/CallbackRegistry.js:97:16)
           at CallbackRegistry.create (/opt/hostedtoolcache/node/18.20.8/x64/lib/node_modules/@suffolklitlab/alkiln/node_modules/puppeteer-core/lib/cjs/puppeteer/common/CallbackRegistry.js:22:26)
           at Connection._rawSend (/opt/hostedtoolcache/node/18.20.8/x64/lib/node_modules/@suffolklitlab/alkiln/node_modules/puppeteer-core/lib/cjs/puppeteer/cdp/Connection.js:89:26)
           at CdpCDPSession.send (/opt/hostedtoolcache/node/18.20.8/x64/lib/node_modules/@suffolklitlab/alkiln/node_modules/puppeteer-core/lib/cjs/puppeteer/cdp/CDPSession.js:66:33)
           at IsolatedWorld.adoptBackendNode (/opt/hostedtoolcache/node/18.20.8/x64/lib/node_modules/@suffolklitlab/alkiln/node_modules/puppeteer-core/lib/cjs/puppeteer/cdp/IsolatedWorld.js:109:46)
           at IsolatedWorld.adoptHandle (/opt/hostedtoolcache/node/18.20.8/x64/lib/node_modules/@suffolklitlab/alkiln/node_modules/puppeteer-core/lib/cjs/puppeteer/cdp/IsolatedWorld.js:126:28)
           at async CdpElementHandle.<anonymous> (/opt/hostedtoolcache/node/18.20.8/x64/lib/node_modules/@suffolklitlab/alkiln/node_modules/puppeteer-core/lib/cjs/puppeteer/api/ElementHandle.js:264:60)
           at async Object.tapElementAndStayOnPage (/opt/hostedtoolcache/node/18.20.8/x64/lib/node_modules/@suffolklitlab/alkiln/lib/scope.js:298:29)
           at async Object.tapElementAndNavigate (/opt/hostedtoolcache/node/18.20.8/x64/lib/node_modules/@suffolklitlab/alkiln/lib/scope.js:263:5)
```

More context of the error: https://github.com/SuffolkLITLab/docassemble-ALWeaver/actions/runs/19590236387?pr=953

Some external discussion of it: https://github.com/SeleniumHQ/selenium/issues/15401

This is a little of a hacky fix, but simply not trying to get the text content before clicking the object is fixes the issue.

### DIFF
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -7,6 +7,7 @@ const { AxePuppeteer } = require('@axe-core/puppeteer');
 const safe_filename = require("sanitize-filename");
 // "en" is faster to import: https://github.com/faker-js/faker/issues/1114#issuecomment-1169532948
 const faker = require('@faker-js/faker/locale/en').faker;
+const { v4: uuidv4 } = require('uuid');
 
 // Ours
 const pdf_reader = require('./utils/pdf_reader');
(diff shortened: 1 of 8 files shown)


lib/scope.js
```diff
@@ -259,13 +259,26 @@ module.exports = {
...
+        () => elem.evaluate( (el) => { console.log("adding log in console"); return el.click(); }).then(console.log("logging the 'then'")),
+        () => {
+          return Promise.race([
+            scope.page.waitForNavigation({ waitUntil: 'domcontentloaded' }),
+            new Promise(() => {
+              page_id != scope.page_id
+            }),
+          ]);
+        },
+      ]
```
plocket: The existing issue most relevant to this is https://github.com/SuffolkLITLab/ALKiln/issues/229.

The original error was triggered because the page had already navigated before we got the text of the page. My mistake for misunderstanding the behavior of async execution. The problem is deeper than the following, but the following may still be part of the solution:

`tapElementAndNavigate` passes promises `tapElementAndStayOnPage` (I know, the names don't work. We need to refactor to reduce confusion.) Instead, it should pass a list of functions, then `tapElementAndStayOnPage` should turn that into a list of Promises to use in a `Promise.race()` (as seen in these related lines we've worked on).

That will solve this particular bug but immediately runs into the next bug caused by this undetected navigation - the interview hangs waiting to detect navigation that's already happened.

Navigation detection has always been a problem in this project. I started some work on more comprehensive navigation detection in https://github.com/plocket/ALKilnReview/tree/229_930_navigation. We might be able to combine the work here with the work on that branch, though we may find the work on that branch to be sufficient.

plocket: False alarm, the problem is a bug in the most recent correction. The new `.map` isn't quite right. Something like it still needs to be part of the solution and I'll make a push. Still made progress on navigation (which I'm still going to want to solve at some point), which you can read in the issue I linked above.

lib/scope.js
```diff
@@ -7,6 +7,7 @@ const { AxePuppeteer } = require('@axe-core/puppeteer');
 const safe_filename = require("sanitize-filename");
 // "en" is faster to import: https://github.com/faker-js/faker/issues/1114#issuecomment-1169532948
 const faker = require('@faker-js/faker/locale/en').faker;
+const { v4: uuidv4 } = require('uuid');
```
plocket: Remove unnecessary change

```suggestion
```

plocket:
I only see the expected errors. I'll merge and squash to get rid of some of the cruft.

plocket (approved):
The test that needs to pass is passing, thanks for starting the ball rolling 👍

PR #1041, opened 2026-02-02
Close #1014, ALKiP notifications

### LINKED ISSUE #1014: Allow ALKiln to give notifications to ALKiP to show to authors in the Playground pre-run
We sometimes need authors to update ALKiP (ALKilnInThePlayground) at the same time as they update their ALKiln version in order to get the latest features [or possibly, in future, avoid breaking something]. We need to be able to send messages to the authors when they arrive or are installing a new version of ALKiln, etc. [or new information about their current ALKiP version that would help them].

I'm not sure how we will determine what circumstances should trigger various messages [i.e. what variables do we need from ALKiP]. Please discuss.

[
The very first message will be a test and isn't going to show up for authors because it would only show up if an author had updated ALKiP to a version that would show the message, at which point the author doesn't need a message, so it won't show up.

More functionality details: ALKiP will run a new AlKiln command and send relevant details about the environment and the author's actions. This time around, this would have been triggered if the author was about to update ALKiln to a version => 5.15.x while their ALKiP was at >= 1.3.x. We would have needed:

- The version of ALKiln authors are asking for
- The version of ALKiln authors are using

Will we need more information in the future?

- The version of ALKiP they have in the Playground
- docassemble-os version?
- Version of AssemblyLine if it's there
- Potentially all ALKiln-specific env vars, but I can't see why at the moment
- The paths of files we know we've put on the system...? (runtime_config.json) That might be useful if we ever want to tell authors to get rid of theoretical old files or to tell them they can update to a version of ALKiP that would clean up old files.

]

plocket:
In this PR, I have:

TODO:
* [ ]  Implement newer changes in https://github.com/SuffolkLITLab/ALKiln/tree/1014_breaking_news. Most easily seen in [this repo's network graph](https://github.com/SuffolkLITLab/ALKiln/network). Leave out the changes to do with file generating.

### Reason for this PR

[Early PR for quick top-level sanity check review, failing tests are fine for now.]
Summary: Add notifications to ALKilnInThePlayground (ALKiP) for ALKiln changes
Details:
This is a testing framework. One of the environments in which it runs is on a test author's own server. ALKiP is a  package of mine that helps authors runs these tests on their server. I work hard to ensure backwards compatibility, but it can be useful to keep both up to date with the other and sometimes to notify authors of other changes to their server that might affect how ALKiP is running. It's hard for authors to tell which versions match up with each other and when to update in general.

This ALKiln PR generates HTML that ALKiP can use to add notifications about relevant changes to ALKiln, to the author's server, or about whatever else ALKiln thinks is helpful. We will use notifications sparingly.

[Demo of the visual appearance of a notification](https://codepen.io/plocket/pen/ByoLvvY). Visual preview:

ALKiln_notification_preview

Closes #1014 

Publishing to npm and testing on ALKiP

### DIFF
--- a/lib/breaking_news.js
+++ b/lib/breaking_news.js
@@ -1,0 +1,313 @@
+#!/usr/bin/env node
+
+
+// Run the code when it's called as an executable
+if (require.main === module) {
+    try_to_break_the_news_gently({});
+    return;
+}
+
+// Export the function if it is being imported
+module.exports = { try_to_break_the_news_gently };
+
+
+function try_to_break_the_news_gently() {
+  let news = null;
+  try {
+    news = break_the_news();
+  } catch ( breaking_news_error ) {
+    news = `🖊️ ALK0279 NOTE breaking news: Unable to break any news.\n${ breaking_news_error }\n`;
+  }
+
+  process.stdout.write( news );
+  return news;
+}
+
+
+function break_the_news() {
+  /** Returns a string. It will be either announcements for authors or a
+   *     declaration of the lack of announcements depending on the
+   *     arguments/flags of the command/subprocess.
+   * 
+   * Discuss: Should we integrate Log into here for, e.g., debugging?
+   * 
+   * @returns {str} - The breaking news or a non-news message.
+   * */
+
+  /* Example of process.argv: 
+   * [
+   *   '/Users/me/.nvm/versions/node/v18.17.0/bin/node',
+   *   '/Users/me/code/alkiln/lib/breaking_news.js',
+   *   '--alkiln_desired_version=6.20.0',
+   *   '--alkip_version=1.0.0'
+   * ]
+   * */
+  const argv = require(`minimist`)( process.argv.slice(2) );
+  let {
+    alkiln_desired_version,
+    alkiln_current_version,
+    alkip_version,
+  } = argv;
+
+  let alkiln_desired_semver = get_semver_parts( alkiln_desired_version );
+  let alkiln_current_semver = get_semver_parts( alkiln_current_version );
+  let alkip_semver = get_semver_parts( alkip_version );
+
+  let reasons = {
+    feature: `✨ New feature`,
+    versions: `🔀 Version mismatch`,
+    fix: `🛠️ Fix`,
+  }
+
+  let header = `## Breaking news!`;  // `## Hear ye, hear ye`
+  let subtitle = `*An ALKiln version or two has one or more messages for you*`;
+  let valediction = `🌼 ALKiln`;
+
+  let messages = [
+    // This first one is for testing and will never be seen in real tests.
+    //     Authors will need to update ALKiP > 1.3.x to see messages like this
+    //     and that will disqualify it
+    {
+      body: `If you want to be able to use the [constrained random answers Step](https://assemblyline.suffolklitlab.org/docs/components/ALKiln/writing/#constrained_random) you should update your server's version of ALKilnInThePlayground to be above 1.3.1`,
+      do_include: ( none_are_null( alkiln_desired_semver, alkip_semver )
+        && (
+          semver_a_more_recent_than_b( alkiln_desired_semver, [5, 15, 0] )
+          || semver_a_same_as_b( alkiln_desired_semver, [5, 15, 0] )
+        ) && (
+          semver_a_less_recent_than_b( alkip_semver, [1, 3, 1] )
+          || semver_a_same_as_b( alkip_semver, [1, 3, 1] )
+        )
+      ),
+      code: `ALK0280`,
+      reason: reasons.feature,
+    },
+    {
+      body: `If you want to be able to use the [constrained random answers Step](https://assemblyline.suffolklitlab.org/docs/components/ALKiln/writing/#constrained_random) you should update your server's version of ALKiln to be above 5.15.0`,
+      do_include: ( none_are_null( alkiln_current_semver, alkip_semver )
+        && semver_a_less_recent_than_b( alkiln_current_semver, [5, 15, 0] )
+        && semver_a_more_recent_than_b( alkip_semver, [1, 3, 1] )
+      ),
+      code: `ALK0281`,
+      reason: reasons.feature,
+    },
+  ];
+
+  let body_parts = [];
+  for ( let msg_obj of messages ) {
+    if ( msg_obj.do_include === true ) {
+      let lines = [`### ${ msg_obj.reason } (${ msg_obj.code })`, msg_obj.body];
+      body_parts.push( lines.join(`\n\n` ));
+    }
+  }
+
+  let news = `💡 ALK0278 INFO breaking news: Nothing is breaking. Probably.\n`;
+  if ( body_parts.length > 0 ) {
+    news = [ header, subtitle, ...body_parts, valediction ].join(`\n\n`) + `\n`;
+  }
+
+  return news;
+}
+
+
+// =============
+// Helpers
+// =============
+
+function get_semver_parts( version_str ) {
+  /** Given a semver (https://semver.org) formatted version string, return a
+   *     list of ints, and possibly strings, that make up the parts of that
+   *     version string. Otherwise returns null.
+   * 
+   * Note: This makes assumptions about our semver format:
+   *     - There will be three or less `.` characters.
+   *     - After the final period will come an integer which may be followed by
+   *        - Nothing or
+   *        - A `-` followed by other valid semver characters
+   * 
+   * @example
+   * // returns [3, 1, 15, `feat-news-1`]
+   * get_semver_parts(`3.1.15-feat-news-1`)
+   * @example
+   * // returns null
+   * get_semver_parts(`3.a.15-feat-news-1`)
+   * 
+   * @param {str} version_str - See ./docs/version_formatting_rules.md
+   * 
+   * @returns {[int, int, ...int|str]} - Version as ints and maybe also strs.
+   * */
+  if ( !version_str || typeof( version_str ) !== `string` ) { return null; }
+
+  let v_parts_strs = version_str.split(`.`);
+  let initial_parts = strs_to_ints_or_strs( v_parts_strs );
+
+  if ( initial_parts.length < 3 ) { return null; }
+  if (
+    typeof( initial_parts[0] ) !== `number`
+    || typeof( initial_parts[1] ) !== `number` ) {
+    return null;
+  }
+
+  let version_parts = [ initial_parts[0], initial_parts[1] ];
+
+  // Handle "experimental" versions. E.g. "3.1.15-feat-news"
+  if ( typeof( initial_parts[2] ) === `string`) {
+    let patch_parts = split_around_first_dash( initial_parts[2] );
+    let patch_num = str_to_int_or_str( patch_parts[0] );
+    version_parts.push( patch_num, patch_parts[1] );
+
+  } else if ( typeof( initial_parts[2] ) === `number`) {
+    version_parts.push( initial_parts[2] );
+
+  } else {
+    return null;
+  }
+
+  return version_parts;
+}
+
+
+function strs_to_ints_or_strs( strs ) {
+  /** Given a list of strings, turn each int string into an actual int and
+   *     leaves the rest as-is.
+   * 
+   * @example
+   * // returns [ 2, 10, 1 ]
+   * strs_to_ints_or_strs([ `2`, `10`, `1` ])
+   * 
+   * @example
+   * // returns [ 2, 10, `1-fix-typo` ]
+   * strs_to_ints_or_strs([ `2`, `10`, `1-fix-typo` ])
+   * 
+   * @params {[str]} strs - Strings
+   * 
+   * @returns {[int|str]} - List of ints or strings or both
+   * 
+   * */
+  let ints_and_or_strs = [];
+  for ( let one_str of strs ) {
+    ints_and_or_strs.push( str_to_int_or_str( one_str ));
+  }
+
+  return ints_and_or_strs;
+}
+
+
+function str_to_int_or_str( original_str ) {
+  /** Converts a string to an int if it contains only digits. Otherwise, returns
+   *     the original string.
+   *
+   * @param {str} original_str - The string to convert.
+   * 
+   * @returns {int|str} - An int if reasonable, otherwise the original string.
+   */
+  try {
+
+    if ( includes_non_digit( original_str )) { return original_str; }
+    return parseInt( original_str );
+
+  } catch ( str_to_int_error ) {
+    return original_str;
+  }
+}
+
+
+function includes_non_digit( original_str ) {
+  /** Returns whether the given string contains any non-digit characters.
+   *
+   * @param {str} original_str - The string to check.
+   * 
+   * @returns {bool} - true if the string has non-digit chars, otherwise false.
+   */
+  return original_str.match(/\D/);
+}
+
+
+function split_around_first_dash( original_str ) {
+  /** Splits a string into two parts around the 1st dash (`-`). Excludes the
+   *     dash itself.
+   * 
+   * @example
+   * // returns [ 3, `feat-2` ]
+   * split_around_first_dash( `3-feat-2` )
+   *
+   * @param {str} original_str - The string to be split.
+   * 
+   * @returns {[str, str]} - The str before the 1st dash and the str after.
+   */
+  let parts = original_str.match(/^([^-])-(.+)$/);
+  return [ parts[1], parts[2] ];
+}
+
+
+function none_are_null( ...items ) {
+  /** Returns true if all items are not `null`. Otherwise, returns false.
+   *
+   * @param {...*} items - The items to be checked.
+   * 
+   * @returns {bool} - true if none of the items are null, otherwise false.
+   */
+  return items.every(( item ) => { return item !== null; });
+}
+
+
+function semver_a_more_recent_than_b( semver_a, semver_b ) {
+  /** Compares two semvers to see if the 1st semver is more recent than the 2nd.
+   *
+   * WARNING: This function doesn't differentiate well between different dev
+   *     versions, as described in ./docs/version_formatting_rules.md
+   *
+   * @param {[int|str]} semver_a - The 1st semver (more recent semver?)
+   * @param {[int|str]} semver_b - The 2nd semver.
+   * 
+   * @returns {bool} - true if semver_a is more recent than semver_b, otherwise
+   *     false.
+   */
+  for ( let index = 0; index < semver_a.length; index++ ) {
+    if ( semver_b[ index ] === undefined ) {
+      // a is longer than b
+      return true;
+    }
+    if ( semver_a[ index ] > semver_b[ index ] ) {
+      return true;
+    }
+    if ( semver_a[ index ] < semver_b[ index ] ) {
+      return false;
+    }
+  }
+
+  return false;
+}
+
+function semver_a_less_recent_than_b( semver_a, semver_b ) {
+  /** Compares two semvers to see if the 1st semver is less recent than the 2nd.
+   *
+   * @param {[int|str]} semver_a - The 1st semver (less recent semver?)
+   * @param {[int|str]} semver_b - The 2nd semver.
+   * 
+   * @returns {bool} - true if semver_a is less recent than semver_b, otherwise
+   *     false.
+   */
+  let is_less = true;
+  if ( semver_a_more_recent_than_b( semver_a, semver_b )) {
+    return false;
+  }
+  if ( semver_a_same_as_b( semver_a, semver_b )) {
+    return false;
+  }
+  return true;
+}
+
+
+function semver_a_same_as_b( semver_a, semver_b ) {
+  /** Compares two semver arrays to see if they are the same.
+   *
+   * @param {[int|str]} semver_a - The 1st semver as an array of ints &/or strs.
+   * @param {[int|str]} semver_b - The 2nd semver as an array of ints &/or strs.
+   * 
+   * @returns {bool} - true if semver_a is the same as semver_b, otherwise
+   *     false.
+   * */
+  let str_a = semver_a.join(``);
+  let str_b = semver_b.join(``);
+  return str_a === str_b;
+}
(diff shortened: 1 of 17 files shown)


lib/setup/artifacts.js
```diff
@@ -12,9 +12,15 @@ session_vars.delete_artifacts_path_name();
 // If using `npm run cucumber` (as opposed to the bin commands),
 // use `--` before `--sources=./foo`
 const argv = require(`minimist`)( process.argv.slice(2) );
-const artifacts_path = files.make_artifacts_folder( argv.path );
+
+const artifacts_path = files.make_artifacts_folder();
```
tobynorth, line 16: **nit**: `argv` is no longer used -- consider either cleaning it up or (if this was accidental) restoring it as an arg to `make_artifacts_folder`

lib/steps.js
```diff
@@ -118,6 +121,9 @@ Before(async (scenario) => {
 
   // Start the running "progress bar" for the Scenario
   log.stdout({}, `\nScenario: ${ scenario.pickle.name }: `);
+  // Save Scenario data in Scenario folder
+  // '@alkiln @generated'
+  // console.log( JSON.stringify( scenario ) );
```
tobynorth, line 126: **question/nit**: did you mean to commit these comments?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@suffolklitlab/alkiln",
-  "version": "5.15.4",
+  "version": "5.15.0-feat-news-17",
```
tobynorth, line 3: **question (non-blocking)**: just to satisfy my own curiosity, what's going on with the version number here? (in particular, why did the patch version number drop from 4 to 0?)

lib/steps.js
```diff
@@ -155,6 +161,16 @@ Before(async (scenario) => {
   fs.mkdirSync( scope.paths.scenario );
   // Store path name for this Scenario's individualized report
   scope.paths.scenario_report = `${ scope.paths.scenario }/report.txt`;
+  // If it's a generated Scenario, save the `.feature` file in here
+  // GENERATED_FILES_FOLDER_NAME
+  if ( scenario.gherkinDocument.uri.includes("_alkiln_generated") ) {
```
tobynorth, line 166: **nit**: `_alkiln_generated` is a magic string, ideally would be defined in a more maintainable location. (Where did it come from, anyway? I don't see any other references to it in this PR)

lib/breaking_news.js
```diff
@@ -0,0 +1,1336 @@
...
+    alkip_version,
+  } = {}) {
+  /**
+   * Returns a string representing html for updates about ALKiln and
+   *     ALKilnInThePlayground or about mismatches between ALKiln and
+   *     ALKilnInThePlayground versions or a string representing an error
+   *     object.
+   * 
+   * Note that the function prints the value to the console before returning the
+   *     value. This is the only way ALKilnInThePlayground can get the value.
```
tobynorth, line 88: **question**: why is this? Seems cleaner to use a temp file; is that not feasible for some reason?

lib/breaking_news.js
```diff
@@ -0,0 +1,1336 @@
...
+   * break_the_news({ alkiln_current_version: "5.15.0", alkip_version: "1.3.2" })
+   * break_the_news({})
+   * */
+  let alkiln_desired_semver = get_semver_parts( alkiln_desired_version );
+  let alkiln_current_semver = get_semver_parts( alkiln_current_version );
+  let alkip_semver = get_semver_parts( alkip_version );
+
+  // Hard-coded metadata for "article" (notification) categories
+  // Discuss: Abstract `notification_types`?
+  let notification_types = {
```
tobynorth, line 165: **praise**: charming subheadings, lol. I especially enjoyed the squash one bug ones 😁

lib/breaking_news.js
```diff
@@ -0,0 +1,1336 @@
...
+        ...get_top_masthead_parts(),
+        `<h2 class="flag nameplate">The Furnace</h2>`,
+        ...get_flag_masthead_parts({ publication_volume }),
+        // BREAKING NEWS!!! with kerning adjustments
+        `<div class="banner"><span class="smaller_right">BREA</span><span>K</span><span class='bigger_left bigger_right'>I</span><span class="bigger_left">N</span><span class="smaller_left">G</span> <span class="bigger_right">N</span>E<span>W</span><span class="smallest_left">S!!!</span></div>`,
+        ...articles_section,
+        `<footer class="footer">~ Every person, a universe ~</footer>`,
+      ], `</div>`,
+    ], `</div>`,
+  ];
```
tobynorth, line 286: **suggestion**: consider refactoring hardcoded HTML & CSS into a template and using a templating engine

lib/breaking_news.js
```diff
@@ -0,0 +1,1336 @@
...
+// Helpers
+// =============
+
+function get_relevant_articles_parts({ articles_data }) {
+  /**
+   * Returns a nested list of HTML strings or lists of HTML strings or both for
+   *     the articles for each relevant notification, as well as the most recent
+   *     "publication volume" number.
+   * 
+   * @param {obj} obj - Named arguments
```
tobynorth, line 303: **nit/question**: I don't see `obj` anywhere other than in this docstring. Did you mean to just list `articles_data` (and its contents) here?

lib/breaking_news.js
```diff
@@ -0,0 +1,1336 @@
...
+   * @param {arr} obj.articles_data - List of data for every possible "article"
+   *     where an article is the data for notifications that might be useful to
+   *     an author based on their current versions of ALKiln and
+   *     ALKilnInThePlayground, or for other reasons.
+   * @param {bool} [obj.articles_data[n].is_no_news_notification] - True if this
```
tobynorth, line 308: **question (non-blocking)**: why not just hide the news notification entirely if there's no news?

lib/breaking_news.js
```diff
@@ -0,0 +1,1336 @@
...
+   * @returns {[str]} - Shuffled list of possible subheadings for this article
+   * */
+
+  let notification_types_subheads = [];
+  for ( let one_notification_type of article.notification_types ) {
+    notification_types_subheads = notification_types_subheads.concat( one_notification_type.subheads );
+  }
+  let shuffled_subheads = get_shuffled_copy(notification_types_subheads);
+
+  return shuffled_subheads;
```
tobynorth, line 409: **suggestion**: a simpler way to do this would be:
```suggestion
return get_shuffled_copy(
  article.notification_types.flatMap(notification_type => notification_type.subheads)
);
```

lib/breaking_news.js
```diff
@@ -0,0 +1,1336 @@
...
+  let final_publication_volume = 0;
+  for ( let article of articles_data ) {
+
+    if ( !article.relevant ) { continue; }
+    let there_is_some_news = articles_section.length > 0;
+    if ( there_is_some_news && article.is_no_news_notification ) { continue; }
+
+    let shuffled_subheads = get_shuffled_subheads({ article });
+
+    let final_subhead = `We do what we must because we can`;
```
tobynorth, line 359: 🎵 for the good of all of us ...

lib/breaking_news.js
```diff
@@ -0,0 +1,1336 @@
...
+      if ( subhead.includes(`"`) && used_quotes ) { continue; }
+      if ( subheads_used[ subhead ]) { continue; }
+
+      if ( subhead.includes(`!`)) { used_exclamation = true; }
+      if ( subhead.includes(`?`)) { used_question = true; }
+      if ( subhead.includes(`"`)) { used_quotes = true; }
+
+      final_subhead = subhead;
+      subheads_used[ subhead ] = true;
+    }
```
tobynorth, line 373: **suggestion**: it feels like it would be a little cleaner to refactor this into `get_shuffled_subheads` by just ensuring that a max of 1 subhead with each checked char (and no duplicate subheads) are added to the return array. Then in this function, you could just `pop` or `shift` an element off the array to pass to `get_one_articles_parts`

lib/breaking_news.js
```diff
@@ -0,0 +1,1336 @@
...
+   * @param {obj} obj - Named arguments
+   * @param {[str]} paragraphs - Text of the paragraphs of an article
+   * 
+   * @returns [str] - The same strings except wrapped in the appropriate html
+   * */
+  let html_paragraphs = [];
+  for ( let one_paragraph of paragraphs ) {
+    html_paragraphs.push(`<p class="body_copy">${ one_paragraph }</p>`);
+  }
+  return html_paragraphs;
```
tobynorth, line 484: **suggestion**: this could be simplified to:
```suggestion
  return paragraphs.map(paragraph => `<p class="body_copy">${ paragraph }</p>`);
```

lib/breaking_news.js
```diff
@@ -0,0 +1,1336 @@
...
+    let result = indent_and_flatten_html_list(
+      nested_str_or_list,
+      indent + 1
+    );
+    // // .concat() feels evil because it accepts a string as well
+    // flattened = flattened.concat( result );
+    flattened.push( ...result );
+  }
+  
+  return flattened;
```
tobynorth, line 641: **suggestion**: I think this could be simplified to:
```suggestion
  return list_or_str.flatMap(nested_str_or_list => 
    Array.isArray(nested_str_or_list) 
      ? indent_and_flatten_html_list(nested_str_or_list, indent + 1)
      : `  `.repeat(indent) + nested_str_or_list
  );
```

lib/breaking_news.js
```diff
@@ -0,0 +1,1336 @@
...
+}
+
+
+function get_semver_parts( version_str ) {
```
tobynorth, line 645: **suggestion**: Consider using a 3rd party library to replace all your semver-related functions if you haven't already ([example](https://www.npmjs.com/package/semver)). Since this is neither related to the core ALKiln logic nor (it seems) something ALKiln needs to go outside the semver spec for, probably better to let someone else maintain it

lib/breaking_news.js
```diff
@@ -0,0 +1,1336 @@
...
+   *     and the string of the date when it was changed. If that's not possible,
+   *     return the current version and the current date.
+   * 
+   * @returns {{version: str, date: str}} - A version number & a formatted date
+   * */
+  try {
+    // If we switch to module (import) instead of common (require), use
+    // https://nodejs.org/api/esm.html#importmetafilename for file path
+    let most_recent_file_sha = execSync(
+      `git rev-list -1 HEAD -- "${ __filename }"`
```
tobynorth, line 857: **question**: this is running on users' servers, right? Is it safe to assume they have git installed and have cloned the ALKiln repo locally?...

tobynorth (approved):
Looks good to me! Didn't find any major issues. One high-level, almost certainly out of scope question for my own understanding: it seems like the need for this PR arose out of ALKiln and ALKiP being separate packages -- why is that? (would you ever want to install ALKiP without wanting to install ALKiln?)

And a general question: how are all the logging changes (artifacts.js, validate.js, Log.js) and path changes (files.js, session_vars.js, set_sources_path.js, scope.js) related to the new notifications feature? For clarity, I'd suggest either putting something about these changes in the PR description or moving them to separate PR(s).

lib/setup/artifacts.js
plocket: Great point, thank you!

I realize you have no context, but I'm going to think about this "out loud", so if you do have thoughts I'd appreciate them. tl;dr: ~I think I could bring it back to add flexibility without any particular downside.~ [I think I should remove `argv` considering how `session_vars` finds the config path now, though I'm still not sure I'm a fan.]

If I recall correctly, I had meant to remove it. I think the caller is now meant to make sure the root of the command is where the artifacts folder should be. In ALKiP (a project that uses this lib), for example, the `cwd` argument for `subprocess` ensures the root is in `/tmp`. I'm actually not sure I like it as the behavior is a bit hidden. I remember having trouble wrangling the root of the path for ALKiP for some reason, though, and landing on this solution. ~Seems to me like I could have both without a problem.~ [Going to keep using `cwd` for now, though, since `session_vars` is strictly using `cwd` with no other arguments. It makes sense I suppose. We'd otherwise have to manage an additional env var. See `session_vars.get_runtime_config_path()`.]

[Note to self: Worth adding some of these notes to the docs.]

lib/setup/artifacts.js
```diff
@@ -12,9 +12,15 @@ session_vars.delete_artifacts_path_name();
...
 
-// Leave the config Project name value as it is. Local developers find it
-//    useful to re-use that value.
+const log = new Log({ path: artifacts_path, context: `artifacts` });
+log.info({ code: `ALK0284`, context: `artifacts`,},
+  `Created the artifacts folder at "${ artifacts_path }"`
+);
+
+// This is just for artifacts, not any other config values. Leave the config
+// Project name value as it is. Local developers find it useful to re-use it.
```
plocket, line 26: What does "This" mean, Past Self? I think it meant that saving the artifacts path name here is unique in some way. Maybe because it's updated with every run of the tests. If so, we should write documentation about which config values get changed and when they get changed. Suggestion:

Changing config values:
- Artifacts folder: name is different with each test run since artifacts are unique to each test run.
- Project name: the env vars determine the Project name. Local devs often re-use the same Project. Other environments, like GitHub or ALKiP are one-use and manage their own Project name value.

Remove comment from code here:
```suggestion
const log = new Log({ path: artifacts_path, context: `artifacts` });
log.info({ code: `ALK0284`, context: `artifacts`,},
  `Created the artifacts folder at "${ artifacts_path }"`
);

```

lib/steps.js
plocket: Nice catch!


```suggestion
```

lib/steps.js
plocket: Good catch here too. The name is [already in globals.js](https://github.com/SuffolkLITLab/ALKiln/blob/067d26b0e3704b5223242ba5622bfe3072f698b6/lib/globals.js#L4) and I should be using that instead. I made a variable for it and everything! Looks like I don't need a separate variable for that atm anyway.

```suggestion
  if ( scenario.gherkinDocument.uri.includes(globals.generated_files_folder_name) ) {
```

lib/steps.js
```diff
@@ -155,6 +161,16 @@ Before(async (scenario) => {
   fs.mkdirSync( scope.paths.scenario );
   // Store path name for this Scenario's individualized report
   scope.paths.scenario_report = `${ scope.paths.scenario }/report.txt`;
+  // If it's a generated Scenario, save the `.feature` file in here
+  // GENERATED_FILES_FOLDER_NAME
```
plocket, line 165:
```suggestion
```

lib/steps.js
```diff
@@ -25,6 +25,9 @@ const session_vars = require('./utils/session_vars');
 const files = require('./utils/files' );
 const reports = require(`./utils/reports`);
 const da_api = require('./docassemble/docassemble_api_interface');
+const globals = require(`./globals`);
+
+const GENERATED_FILES_FOLDER_NAME = globals.generated_files_folder_name;
```
plocket, line 30:
```suggestion
```

package.json
plocket: No problem: I just manage branch versions based off of their base branch. When I first diverged from the default branch, the version was `5.15.0`, that's all. I'm curious to hear if you've seen other methods!

plocket:
Thanks for the excellent and detailed review! I've finally started in on this, though it'll stay here a while longer (see below).

> And a general question: how are all the logging changes (artifacts.js, validate.js, Log.js) and path changes (files.js, session_vars.js, set_sources_path.js, scope.js) related to the new notifications feature? For clarity, I'd suggest either putting something about these changes in the PR description or moving them to separate PR(s).

Good insticts! The logging and path changes were indeed meant to be in a separate PR and I'll move them there and deal with those first before addressing the notification feature, so this PR is going to end up sitting on the back burner a little while longer unfortunately. And will probably run into merge conflicts unless I make a whole new PR, which I won't because I don't want to lose this awesome review. Some days you can't have it both ways `¯\_(ツ)_/¯`

> One high-level, almost certainly out of scope question for my own understanding: it seems like the need for this PR arose out of ALKiln and ALKiP being separate packages -- why is that? (would you ever want to install ALKiP without wanting to install ALKiln?)

I _think_ it's understandable without too much further context. There are 4 environments in which ALKiln can run and ALKiP is just one of those. As such, though you'd never want to install ALKiP without ALKiln, you may very well want to install ALKiln without ALKiP. At least, we have some users who do that.

plocket:
Before merging, change workflow files to use feature branch.

lib/breaking_news.js
plocket: You're right, this comment is inaccurate. The idea of creating a file hadn't occurred to me. I am reluctant to do so because file reference wrangling has been an albatross around my neck throughout this project and I still haven't figured out why.

Response: May be subject to cli GUI changes. I'm concerned about future breaking changes.

Response to response: There's no cli, just stdout, and that's purely the HTML. Maybe I should make this a JSON string though to future-proof in case I want to include more info in the future.

Response: Are there any other consumers? Or could there be in the future? It's pretty tightly coupled.

Response to response: Good question. Right now there aren't. If there are in the future, putting this in a file instead wouldn't create a breaking change, so I'm not as concerned about that. It would be easy to keep this grandfathered in as a fallback to a file. It is tightly coupled. I'm not completely comfortable with that.

Response: It's odd there are 2 different repositories.

Response to response: I've been wanting to document this. It's a good point. When people install the ALKiln package, they'd have to install all the ALKiP code as well. It makes the size of the package larger and increases the complexity of the repository. This adds some complexity, but avoids other complexity. Also, they sometimes progress at different rates. That's not ideal, but actually may not be easier to wrangle in one repo. For example, we currently have two ALKiP versions available and we've had to implement the same changes on both. Maybe I'm wrong and the complexity would actually be reduced, but I'm not sure how test that out. What if ALKiP's ALKiln version was different than the repo it's in. I think it's worth re-examining, though.

TODO: Put this string in a JSON object at the very least. Consider storing it in a file.

TODO: Discuss merging ALKiP and ALKiln repos.

lib/breaking_news.js
plocket: Good thought and very attractive. We already have a bunch of dependencies and I feel leery about adding more. Mostly security concerns, esp. with npm attacks right now. Also, complexity. It would depend on what affordances the libraries we chose had. And the time required to research those options and then to keep those packages up to date. Do you have some specific suggestions?

Response: Security and updating, depend-a-bot might be an option. We merged automatically for patch and minor updates and reviewed for major changes.

Response response: Yes, it's something I use on some repos. I can't let it merge changes directly because some of those changes have broken our code in the past. There are some changes to cucumber that the devs don't consider breaking changes that we would break our code. I know from discussions with them. We could review everything every time, but we don't have a lot of people power here. If you're into that, though, I could make you the steward of updates!

Response: I will seriously consider that!

lib/breaking_news.js
plocket: Good question! This is, I think, a quirk of mine for docstrings. In this function, I accept an object as an argument and de-structure it to get articles_data. I choose to represent that in the docstring with obj because it doesn't have a meaningful name of its own. It represents what I call "named arguments" - the keys of the object. Maybe I should rename it to anon! Or something.
Response: There's only one arg. Why not just pass articles_data?
Response response: Maybe that's a good idea. Some thoughts:

1. It's a convention I use throughout my code and I try to stick to it. I'm not completely consistent, so maybe consistency is an insufficient reason.
2. It means this function and the caller will have the same name for the values. I often find that useful to know. I feel it helps with searching for names, updating names, and keeping them in sync. Maybe that's just vibes, though.
3. Bad reason perhaps: Future proofing. I often find myself adding new arguments and I dislike changing the whole data structure of function's args in those cases.

lib/breaking_news.js
plocket: Also a good question! On the other end I do hide it. I do like giving some information as opposed to no information, though. It helps me know the system is working. Like an `ok` signal. Missing information is then a signal that something un-handled has gone wrong.

lib/breaking_news.js
plocket: TODO: consider using a single, non-anonymous-obj, argument now that the code has settled down.

lib/breaking_news.js
plocket: It is reasonable to assume! Thank you for thinking of this, though. `git` is always installed on da servers. ALKiln isn't, but this is ALKiln. If this code is running, then ALKiln is installed.

lib/breaking_news.js
plocket: TODO: Maybe if I change this into a JSON object instead, I could just send an actual `ok` response key. But then I wouldn't get to see all the fun defaults! I will have to think on this.

lib/breaking_news.js
plocket: I like this!

TODO: Implement above!

lib/breaking_news.js
plocket: TIL about `flatMap`s! Iterator discussion IRL. TODO: Consider capitulating to iterators.

lib/breaking_news.js
plocket: TODO: Research! This is an even better candidate for a 3rd party library. So much more fiddly.

lib/breaking_news.js
plocket: I think even if this were compressed/compiled, `npm` would send its version and `git` commits. I think the `git` info is part of the `sha` that guarantees the `npm` version. I'm not completely sure, though.

lib/setup/artifacts.js
plocket: Removed `argv` in related commit: https://github.com/SuffolkLITLab/ALKiln/pull/1048/changes#diff-16b41121da238f1f9ac601b37621476fd1cf84ee11f0a55824bd38036114c153R14

lib/steps.js
plocket: Removed in [related commit](https://github.com/SuffolkLITLab/ALKiln/commit/2f29d9b0b23e8938987276077fb9f25614564a50)

lib/setup/artifacts.js
plocket: I take it back.

> I think I should remove argv considering how session_vars finds the config path now...

The artifacts path root is unrelated to the config path. Glad this conversation is here. It helped me track down my thinking during this transition. Two steps forward, three steps back, yakkity yak.

PR #1045, opened 2026-04-07
Update default docassemblecli version to 0.0.25

BryceStevenWilley:
The current default docassemblecli version is set to 0.0.17. docassemblecli 0.0.25 contains fixes that allow it to upload projects that contain only a `pyproject.toml`, and no `setup.py`, which isn't needed when installing from GitHub or manually with a zip file.

Confirmed this is necessary and working on AssemblyLine's Kiln tests: https://github.com/SuffolkLITLab/docassemble-AssemblyLine/pull/1039/changes/c78befe418f130d3798e7898153bf18b1affc190

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,7 +43,11 @@ Format:
 - 
 -->
 
-<!-- ## [Unreleased] -->
+## [Unreleased]
+
+### Changed
+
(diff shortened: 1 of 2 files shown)


plocket (approved):
Thanks for the update! Tests passing after rerun, LGTM 👍

PR #1046, opened 2026-04-07
Handle curl failures while servers are starting

### LINKED ISSUE #1044: Retry curl failures when waiting for GitHub Action server to start up
Ran into this error when running an GHA server test in Assembly Line:

```
ALK0013 INFO: Time elapsed: 60 seconds
💡 ALK0015 INFO: Response is 404
💡 ALK0017 INFO: The docker container is not ready yet. Will check again in 30 seconds.
💡 ALK0013 INFO: Time elapsed: 90 seconds
Error: Process completed with exit code 56.
Run log="🤕 ALK0022 ERROR: Something went wrong. See above for more details."
```

None of the logs have any detail, but the only line that would return such a specific error code is the call to `curl`, https://github.com/SuffolkLITLab/ALKiln/blob/eceab90f42e34c3d17b72b53fd8765a22827c8d6/action_for_github_server/action.yml#L222. 

Error code 56 is "failure when receiving data from peer", which is a bit generic and has a few different causes, easy to see it happening when the server is just starting up though.

IMO, more of the curl errors should be treated like a 404 error, and that we're still waiting for the server to start. It should at least be retried once.

https://github.com/SuffolkLITLab/docassemble-AssemblyLine/actions/runs/23148630938/attempts/1?pr=1030

rajeswari1301:
In this PR, I have:

* [ ] Added tests for any new features or bug fixes (if relevant)

### Reason for this PR
curl sometimes fails with exit code 56 (failure receiving data from peer) while the docassemble server is still starting up. The script had no handling for curl network-layer failures - it was only checking HTTP response codes, so these transient errors could cause flaky CI failures.

Fixes #1044 

Discussed the approach with Bryce before implementing. Reviewed the retry loop logic manually to confirm the curl exit code is captured correctly and the loop continues as expected when curl fails.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,7 +48,7 @@ Format:
 ## [Unreleased]
 
 ### Fixed
-- Retry on curl exit code 56 (failure receiving data from peer) while waiting for docassemble server to start up. Added set +e to prevent GitHub Actions from killing the acript on curl failures, retries on exit code 56 and exits wuth actual curl exit code on any unexpected error See [#1044](https://github.com/SuffolkLITLab/ALKiln/issues/1044).
+- Handle curl failures while waiting for docassemble server startup. Wrapped the curl call with set +e / set -e so GitHub Actions strict error handling doesn't kill the script on curl failures. Any non-zero curl exit code is treated as "not ready yet" and retries after 30 seconds. See [#1044](https://github.com/SuffolkLITLab/ALKiln/issues/1044).
 
 
 ## [5.15.4] - 2025-11-29
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,7 +48,7 @@ Format:
 ## [Unreleased]
 
 ### Fixed
-- Handle curl failures while waiting for docassemble server startup. Wrapped the curl call with set +e / set -e so GitHub Actions strict error handling doesn't kill the script on curl failures. Any non-zero curl exit code is treated as "not ready yet" and retries after 30 seconds. See [#1044](https://github.com/SuffolkLITLab/ALKiln/issues/1044).
+- Handle curl failures while waiting for docassemble server startup. Wrapped the curl call with set +e / set -e so GitHub Actions does not exit on curl failures. All non-zero curl exit codes are treated as "not ready yet" and retried - this covers transient errors like exit code 56 (connection dropped) and exit code 7 (connection refused) that may come up during startup. Added a comment explaining the permissive approach. Changed response value from "000" to "missing" for clarity in logs. See [#1044](https://github.com/SuffolkLITLab/ALKiln/issues/1044).
 
 
 ## [5.15.4] - 2025-11-29
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,7 +48,7 @@ Format:
 ## [Unreleased]
 
 ### Fixed
-- Retry on curl exit code 56 (failure receiving data from peer) while waiting for docassemble server to start up. See [#1044](https://github.com/SuffolkLITLab/ALKiln/issues/1044).
+- Retry on curl exit code 56 (failure receiving data from peer) while waiting for docassemble server to start up. Added set +e to prevent GitHub Actions from killing the acript on curl failures, retries on exit code 56 and exits wuth actual curl exit code on any unexpected error See [#1044](https://github.com/SuffolkLITLab/ALKiln/issues/1044).
 
 
 ## [5.15.4] - 2025-11-29
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,12 @@ Format:
 
 <!-- ## [Unreleased] -->
 
+## [Unreleased]
+
+### Fixed
+- Retry on curl exit code 56 (failure receiving data from peer) while waiting for docassemble server to start up. See [#1044](https://github.com/SuffolkLITLab/ALKiln/issues/1044).
+
+
 ## [5.15.4] - 2025-11-29
 
 ### Changed
--- a/action_for_github_server/action.yml
+++ b/action_for_github_server/action.yml
@@ -198,7 +198,6 @@ runs:
       - name: "💡 ALK0012 INFO: Waiting for server to start"
         shell: bash
         run: |
-          set +e
           log="💡 ALK0012 INFO: Waiting for server to start"
           echo -e "\n$log" >> "${{ env.DEBUG_LOGS_PATH }}"
 
@@ -220,13 +219,13 @@ runs:
             fi
 
             # Set the variable to a new string value in each iteration
+            set +e
             response=$(curl -o /dev/null -s -w "%{http_code}" http://localhost:80/health_check?ready=1)
             status=$?
+            set -e
 
-            if [[ $status -eq 56 ]]; then
+            if [[ $status -ne 0 ]]; then
               response="000"
-            else
-              exit $status
             fi
 
             log4="💡 ALK0015 INFO: Response is $response (curl exit code: $status)"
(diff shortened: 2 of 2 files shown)


action_for_github_server/action.yml
BryceStevenWilley, line 201: Remembering that GitHub Actions sets the shell script to exit early on any error code (`set -e`), which is why we were able to see the 56 error code. We'll need to also revert that GitHub Actions behavior with this as well:

```bash
set +e
```

action_for_github_server/action.yml
```diff
@@ -207,28 +207,34 @@ runs:
...
 
             # Set the variable to a new string value in each iteration
-            response=$(curl -o /dev/null -s -w "%{http_code}\n" http://localhost:80/health_check?ready=1)
-            log4="💡 ALK0015 INFO: Response is $response"
+            response=$(curl -o /dev/null -s -w "%{http_code}" http://localhost:80/health_check?ready=1)
+            status=$?
+
+            if [[ $status -eq 56 ]]; then
+              response="000"
+            fi
```
BryceStevenWilley, line 232: With `set +e`, the script won't automatically fail anymore, so we'll have to still manually exit on non-56 statuses.

```suggestion
            else
                exit $status
            fi
```

BryceStevenWilley (changes requested):
Some changes that I think are needed to make this work.

action_for_github_server/action.yml
plocket: [Edit: Look further down where I expressed a different sentiment.]

I think the check here is too permissive. Non-0, non-56 error codes should still exit with the error code. I think the logic we're looking for is something like this:

If the status is 0, continue
If the status 56, set the response to "000"
Otherwise, exit with the status code.

Regarding a response code of "000": I admit I don't have a better suggestion for a response code. I think "000" is actually a decent option. I'm not sure this is a valid response code, so we may want to consider that, but I do like that "000" can tell us we created it ourselves. I'll go with whatever you decide here.

action_for_github_server/action.yml
```diff
@@ -207,28 +207,36 @@ runs:
...
             echo -e "\n$log4" >> "${{ env.DEBUG_LOGS_PATH }}" && echo -e "$log4"
 
             # Check if the string includes the desired text
-            if [[ $response == *"200"* ]]; then
+            if [[ $response == "200" ]]; then
               log5="💡 ALK0016 SUCCESS: Success! The docker container is ready!"
               echo -e "\n$log5" >> "${{ env.DEBUG_LOGS_PATH }}" && echo -e "$log5"
             else
-              log6="💡 ALK0017 INFO: The docker container is not ready yet. Will check again in 30 seconds."
+              log6="💡 ALK0017 INFO: The docker container is not ready yet (HTTP=$response, curl exit code=$status). Will check again in 30 seconds."
```
plocket, line 242: I'm not opposed to including this additional information, so we can keep this change. I just want to note that this information is logged above on line 231 with `log4="💡 ALK0015 INFO: Response is $response (curl exit code: $status)"` in case you didn't notice it. By the time this log happens, we will have seen `0015` already. Thank you for the addition of the status code, btw!

CHANGELOG.md
```diff
@@ -45,6 +45,12 @@ Format:
 
 <!-- ## [Unreleased] -->
 
+## [Unreleased]
+
+### Fixed
+- Handle curl failures while waiting for docassemble server startup. Wrapped the curl call with set +e / set -e so GitHub Actions strict error handling doesn't kill the script on curl failures. Any non-zero curl exit code is treated as "not ready yet" and retries after 30 seconds. See [#1044](https://github.com/SuffolkLITLab/ALKiln/issues/1044).
```
plocket: When you update the code above to be less permissive, also update this comment to include the detail about exit code 56. Thank you!

action_for_github_server/action.yml
plocket: Actually, let me propose something different because I do want to log "ALK0015" before exiting if possible. I propose (in pseudo code):

If status == 56, response = "000"  # Make sure `response` has a value
Log ALK0015  # Make sure values are logged
If status != 0 and status != 56, exit with status  # Exit if needed

Note that the log comes in the middle - after we make sure `response` has a value. That way we can always log before exiting.

plocket (changes requested):
Thanks for the changes! I think we need logic that's a little less permissive for the status code check. Other than that, the rest of the items from our discussion are up to you. Great work!

action_for_github_server/action.yml
BryceStevenWilley: For what it's worth when working on this I believe @rajeswari1301 ran into at least one other error code that were also transient and related to the server starting up. There are likely to be quite a few of them if we looked through the docs comprehensively.

plocket: It's a good point to bring up, though I'm not sure about that error you mentioned specifically (see below). Maybe the question I should be asking is this: are there status errors it would be bad to skip? I'm not familiar with the status errors that could come up in this context. I can't see what could possibly be that highly concerning, though. Famous last words?

If nothing could go significantly wrong, the only down-side is making authors wait a full 2 minutes while their tests churn for no reason. That's not particularly fun, but re-running the tests costs time too. Also, people usually go do other things while tests are running, which means re-trying is probably the bigger time saver in practical terms.

We don't lose the status error because we already log it. We have a time limit on the operation so it won't loop forever. It's a sandbox, so no harm can be done to an author's servers. I can't see how harm could be done to the repo itself either.

Ok, I'm convinced! If we can't [find or come up with a serious ramification](https://curl.se/libcurl/c/libcurl-errors.html#CURLcode), let's skip errors. [There are only about 100 and I took a look. See notes below.]

I still like the idea of changing `$response` on a non-zero status code. What do folks think of changing `"000"` to `"missing"`? That reads well in the log and, after some thought, if it's not going to be a valid status code I'm thinking it's worth making it more expressive.

---

Incidental note:

> an into at least one other error code that were also transient and related to the server starting up

Was the error [this one](https://github.com/SuffolkLITLab/ALKiln/actions/runs/24144620820/job/70455092595)? If so, it was because the script exited after a non-56 (which was `0`) and the server wasn't ready yet. The test then failed when it tried to `dainstall`.

plocket: Also, how do folks feel about adding a comment or a design doc describing why we're ignoring all status code errors. I suspect future me will be surprised that we're being so permissive.

plocket: tldr; Go ahead and skip all status codes!

I've read all the errors. The only problem I see is the potential that it's possible that an error code could be useful if 1. we introduce a bug in the code in a future edit or 2. da changes something and it breaks our code. I'm ok with letting those situations resolve after 2 minutes instead of immediately.

rajeswari1301: Thanks for the discussion @plocket ! And thanks @BryceStevenWilley for bringing up the other error code -the loop was exiting before ALK0015 even logged, which pointed to a connection side failure during startup, likely exit code 7 (failed to connect)

My approach of skipping all non zero codes was intentional for this reason, there are likely several transient codes that come up during server startup, and since we already log them and have a timeout, skipping them all is safer than trying to list out each one.

Also agree with @plocket 's suggestion to add a comment explaining why we're being permissive that'll definitely help. And glad we landed on skipping all codes, it keeps it simple and your reasoning makes total sense.

Will update:
- "000" to "missing"
- Add a comment explaining the permissive approach
- Update CHANGELOG to mention the permissive handling of all non-zero curl exit codes

BryceStevenWilley:
I think the test failures here are also due to https://github.com/SuffolkLITLab/docassemble-ALToolbox/pull/339, which is breaking everything. Re-running those now, hopefully they should succeed now that fix is pushed to PyPI.

plocket (approved):
Looks good to me! Thanks for the thorough review process and for reducing the flakiness of the tests. I recommend a squash and merge in this case.

plocket:
@BryceStevenWilley, I think you need to approve since you requested changes.

BryceStevenWilley (approved):

PR #1054, opened 2026-05-14
Protect against malicious pre-install/post-install npm scripts. Close…

### LINKED ISSUE #1053: Strengthen `npm install` security in our actions
Considering recent problems with `npm install` running malicious `preinstall` and `postinstall` scripts, we changed how we installed packages in [ALKiP](https://github.com/SuffolkLITLab/docassemble-ALKilnInThePlayground/). We should do the same for our action. It's true they're not on the author's server, but they can still do harm by exfiltrating GitHub secrets.

---

Edit:

### More details about this change:

We specifically used the `-min-release-age=7` and `--ignore-scripts` arguments with `npm install`. [Update: We're now debugging new trouble with running puppeteer's script. The previously working solution is no longer working.]

~~[Then, because our puppeteer dependency needs to run a script (and no other of our dependencies), we run only the script for puppeteer with `npm rebuild -g puppeteer`.]~~

~~Then, because our puppeteer dependency needs to run a script (and no other of our dependencies), we run only the script for puppeteer with `npm rebuild -g puppeteer --update-binary`.~~

### Details about why we added those `npm install` arguments:

Life is full of risks and software is one of them. Sometimes we can write and run code in ways that help lower the risks software brings.

`npm` "supply chain attacks" is something we weren't aware of and they have been increasing, affecting hundreds and thousands of packages and systems. This is when someone who has the ability to edit a package secretly adds malicious code to a package without anyone else noticing. When someone does this in a commonly used package, it can affect everyone who uses that malicious package and anyone who uses a package that uses that malicious package and so on. Our systems are all heavily reliant on open source packages. Open source packages are packages made by people who are not paid to make those packages. They are often overworked and under-supported. They often lack the resources to properly examine every change contributors make.

#### `--min-release-age=7`

`--min-release-age=7` is a broader kind of protection. It reduces our risk of all malicious code, not just pre- and post- install scripts (see below). It makes sure I only install a new version of a package if it was released more than 7 days ago. Most malicious code is found shortly after it gets added to a package's code. It might be a couple of days. It's often a few hours or less. Once that malicious code is detected, package owners remove the version with the problem and developers can no longer install it by accident. Making sure we avoid using the very latest code helps us avoid installing a lot of malicious code.

[Here is an article about the `--minimum-release-age` argument as of May 2026 (on web.archive.org)](https://web.archive.org/web/20260508094631/https://daniakash.com/posts/simplest-supply-chain-defense/).

> A 7-day package delay would have blocked installs in most short-lived malicious publish attacks from the last 8 years
> ...the poisoned versions were live for about 4 hours before npm pulled them.
> \- [https://daniakash.com/posts/simplest-supply-chain-defense/ on web.archive.org](https://web.archive.org/web/20260508094631/https://daniakash.com/posts/simplest-supply-chain-defense/)

[https://daniakash.com/posts/simplest-supply-chain-defense/ on web.archive.org](https://web.archive.org/web/20260508094631/https://daniakash.com/posts/simplest-supply-chain-defense/) also contains some general information about the risks of malicious code.

#### `--ignore-scripts`

One main way malicious actors can implement supply chain attacks is by adding, or editing, the `preinstall` and `postinstall` scripts of existing popular packages. Those scripts run even without a developer running in the developer's code. All they have to do is install the package. It's very sneaky because many people aren't aware of `preinstall` and `postinstall` scripts and aren't aware this code can automatically run without asking first.

`--ignore-scripts` is one way to combat this. If, as a developer, I search all my packages for those scripts during a time when I'm pretty sure there is no malicious code in those scripts, then I can note exactly which packages I need those scripts for and why. I can decide which scripts I want to keep using, what I can ignore, and whether I should switch to another package because of various risk factors, like how sure I am that this code gets thoroughly reviewed. I can then choose to ignore all other scripts that get added to my existing packages and only run the scripts for the packages that I've approved. If another package suddenly gets a new pre/post-install script, my `npm` installation will ignore it. Later, if I use the package and get an error because I ignored the script, I can look into the new script in details and decide whether I want to run it. It just makes it easier to review code more efficiently and effectively.

[Here is an `npm` article about the `--ignore-scripts` flag](https://web.archive.org/web/20260327115726/https://www.nodejs-security.com/blog/npm-ignore-scripts-best-practices-as-security-mitigation-for-malicious-packages).

> A common supply chain security issue with npm packages is the capability of malicious or compromised npm packages to execute arbitrary code or OS commands during the package installation process.
> The npm package manager (the CLI, not the registry) has a feature called lifecycle hooks...One of the most commonly used lifecycle hooks is the `postinstall` script (but there’s also `preinstall` for example). This script is executed after a package is installed and can be used to perform additional setup or configuration tasks.
> The npm CLI has a flag called `ignore-scripts` that can be used to prevent the execution of any lifecycle hooks defined in the package.json file of the packages you are installing.
> \- npm

For people who develop with npm, this mostly helps with packages that add new pre- and post- install scripts, or packages that I newly install. It doesn't help with packages I already use that already have those scripts. You do what you can.

plocket:
… #1053. Very quick review. Same as for ALKiP npm security. Personal note that unit tests were flaky, which I haven't seen before.

In this PR, I have:

* Action can't be tested in this way ~~Added tests for any new features or bug fixes (if relevant)~~

### Reason for this PR

As simple as it says

Closes #1053

Ran workflows

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,10 @@ Format:
 - Print error correctly when missing sources paths
 - Changed the default version of docassemblecli to 0.0.25
 
+### Security
+
+- Security: Gave our main action's `npm install` call the arguments ` --ignore-scripts --min-release-age=7` to improve security against malicious `pre-install` and `post-install` scripts. See [#1053](https://github.com/SuffolkLITLab/ALKiln/issues/1053).
+
 ## [5.15.4] - 2025-11-29
 
 ### Changed
--- a/action.yml
+++ b/action.yml
@@ -120,7 +120,9 @@ runs:
       run: |
         log="💡 ALK0026 INFO: Install ALKiln directly from npm"
         echo -e "\n$log" >> "${{ env.DEBUG_PATH }}"
-        npm install -g "@suffolklitlab/alkiln@$ALKILN_VERSION"
+        npm install --ignore-scripts --min-release-age=7 -g "@suffolklitlab/alkiln@$ALKILN_VERSION"
+        npm rebuild -g puppeteer --update-binary
+
 
     # Files prep
     - shell: bash


BryceStevenWilley (approved):
Looks good to me!

PR #1064, opened 2026-05-27
Use npm min-age-release only when "internal tests" flag is `true`. See #1055

### LINKED ISSUE #1062: Should we turn off GitHub+You `push` tests for the Suffolk dev server?
They run on the Playground server, which means that if we run tests too close together tests mess each other up by causing the server to reload. [More about GitHub+You tests](https://assemblyline.suffolklitlab.org/docs/components/ALKiln/setup/#githubNyou-details). We could save them for pull requests and manual triggers.

If we want to still use tests here to keep track of our dev server issues, we could add a cron job and that'll have less of a chance of interfering.

plocket:
Closed by #1093

Also added warnings if the repo seems to be using the flag incorrectly/unsafely. When merged, this should incorrectly (in some respects) pass and thus prove that this [action] will work for our users. As we develop a fix for the bug, it may fail again, and thus prove that this [action] will work for us internally.

[Also closes #1062 . Since the PR was already getting rid of `push`, I figured I might as well add the cron job. It's for Sun at 1am.]

In this PR, I have:

* Not yet sure how we can test these properly. We'd need to be able to examine previous action console output. ~~[ ] Added tests for any new features or bug fixes (if relevant)~~
* Again, not an npm release, just a GitHub release. We may want to develop a policy for this. ~~[ ] Added my changes to the CHANGELOG.md at the top, under the "Unreleased" section~~

### Reason for this PR

Allow author's tests to use the npm min release age security feature. Allow us to test more recent versions of ALKiln.

#1055
Closes #1062 

[Visually confirmed at least one use case shows warnings](https://github.com/SuffolkLITLab/ALKiln/actions/runs/26520493132/job/78109808962#step:7:373).


action.yml
```diff
@@ -115,14 +119,42 @@ runs:
...
       run: |
         log="💡 ALK0026 INFO: Install ALKiln directly from npm"
+        repo_name="${{ github.event.repository.name }}"
+        if [[ $repo_name == *"ALKiln"* ]]; then
+          echo "⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️ YOU MAY BE MISSING THE _ALKILN_INTERNAL_TESTS FLAG ⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️"
+        fi
         echo -e "\n$log" >> "${{ env.DEBUG_PATH }}"
+        if [[ $repo_name == *"ALKiln"* ]]; then
+          echo "⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️ YOU MAY BE MISSING THE _ALKILN_INTERNAL_TESTS FLAG ⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️"
+        fi
```
BryceStevenWilley, line 139: What's the reason for printing this line twice?

action.yml
```diff
@@ -115,14 +119,42 @@ runs:
...
+      run: |
+        log="🔎 ALK0279 WARNING: Install **recent** version of ALKiln directly from npm"
+        repo_name="${{ github.event.repository.name }}"
+        if [[ $repo_name != *"ALKiln"* ]]; then
+          echo "⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️ YOU MAY BE USING AN UNSAFE VERSION OF ALKILN ⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️"
+        fi
+        echo -e "\n$log" >> "${{ env.DEBUG_PATH }}"
+        if [[ $repo_name != *"ALKiln"* ]]; then
+          echo "⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️ YOU MAY BE USING AN UNSAFE VERSION OF ALKILN ⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️"
+        fi
```
BryceStevenWilley, line 158: Same here: not sure why we need to print the same line twice, before and after the debug path gets logged?

BryceStevenWilley (approved):
Looks good to me. My only thing is that I'm not sure why we print the same line twice, in both actions. Am I missing something about where the second line goes?

action.yml
plocket: It's just to bring more attention to those lines. I can remove if desired.

plocket:
~~Now fails correctly~~ and closes #1062. Since the PR was already getting rid of `push`, I figured I might as well add the cron job. It's for Sun at 1am.

plocket:
It has stopped failing correctly as it was before. When I removed `push` (an unrelated change), puppeteer stopped building Chrome. I'm looking into it on this branch: https://github.com/SuffolkLITLab/ALKiln/tree/refs/heads/fix_puppeteer_scripts.

Issue #1088, opened 2026-05-30
Node/puppeteer can't find Chrome


plocket:
Moved from #1055.

### Next steps 06/01 15:15

#1093 addressed the immediate problems named here, so this issue is now an enhancement, not a priority or bug. To fully resolve, I'd like to see if we can get to a point where we're not trying to [locate the browser executable ourselves](https://github.com/SuffolkLITLab/ALKiln/blob/6838da231c1088c930ef63be11b3df3885bc00d9/action.yml#L126-L143). My next suggestion is to work out all the issues that new versions of puppeteer would cause and resolve those up to the most recent (minus 7 days) version of puppeteer.

### Decision

We are going to automate ALKiln internal security with `.npmrc` and [still use `npm install` because `npm ci` can't be used for a global package] ~~use `npm ci`~~ both in our own code and in consumer-facing code. Since our dependencies are pinned with npm shrinkwrap, out consumers will benefit from those measures as well. ALKiln's code itself doesn't have the kind of pipeline (that is, thousands of consumers or more) that would catch breaches in the same way that hugely popular packages do, so a 7-day delay in the version of ALKiln itself has very little impact, so we will remove `--min-age-release` from the action, and maybe from ALKiP as well. ~~We may also remove `--ignore-scripts` from consumer-run code for the same reasons.~~ [Our own `--ignore-scripts` won't protect consumers from malicious scripts, so we will continue to use `--ignore-scripts` in code that others consume.]

### Summary 05/29, 13:30
[...]

PR #1093, opened 2026-06-01
Fix #1055, address #1088, action outputs and pipeline issues...

### LINKED ISSUE #1055: Action test files outputs causing error when used in workflow
Unblocked

**05/30 19:40**: Blocked by #1088, without which our tests will never reach the point at which this issue becomes relevant.

Evolving situation updates, most recent first:

**05/30 19:40** (yes, the same time as the above)

Back to the erroring output paths, my suspicion is that we need to restore `argv`. in `artifacts.js`. In a confused fugue one day, I removed the command line arg that allows a process to specify a path for the artifacts folder. [The fugue is recorded in this PR comment chain](https://github.com/SuffolkLITLab/ALKiln/pull/1041/changes#r3329245379) (recording it here because that PR will disappear when we break it into multiple parts).

I need to test the fix in other environments before I publish the final fix.

I appreciate all the solutions people/LLMs have offered for suppressing the error, but we really need to solve the bug itself.

**05/28 10:55**

I believe we're now running into an [unrelated issue where Chrome isn't downloading](https://github.com/SuffolkLITLab/ALKiln/actions/runs/26581857228/job/78316775950#step:5:310). We will take care of that in #1088.

**05/28 9:56**

Plans:

I'm planning to suppress errors in OUTPUT creation, though I believe there's only so much we can do. I plan to, in the event of an error, try make a fake archives folder with fake files that say there was an error. That will let consumers of the output get _some_ kind of output. If that fails, I can give empty strings to the values, but that will probably cause confusion for consumers if it leads to the root of the action and I believe it would still fail when they try to read files. I could, alternatively, put nothing in the outputs at all. I'm not sure what that would end up doing.

Alternatively, we could switch to reading the files ourselves and giving the consumers the text output of those files. If that were the case, we could have a fallback to give the authors safe strings about failures. I think that would be more sure of success because there are less fiddly bits [for both us and the consumers]. Current consumers would have to change their code. I believe we only have one or two consumers at the moment, though, and maybe we could help them manage that.

[Stackoverflow about suppressing errors in Bash](https://stackoverflow.com/questions/22009364/is-there-a-try-catch-command-in-bash). ([Web archive version of Stackoverflow post](https://web.archive.org/web/20260227081227/https://stackoverflow.com/questions/22009364/is-there-a-try-catch-command-in-bash))
[Post on gotchas about suppressing Bash errors](https://mywiki.wooledge.org/BashFAQ/105), though some (most?) of them wouldn't be a problem here. ([Web archive version of gotchas](https://web.archive.org/web/20260515025008/https://mywiki.wooledge.org/BashFAQ/105))

**05/27 10:30**

Notes:

Idea: 2 new `inputs`
`_ALKILN_INTERNAL_TESTS`
`_ALKLIN_SECURITY_SENSITIVE_NPM_ARGS`
1. With “internal tests”, warn if min release age is not present (not just default).
2. With “npm args” warn if there is no min-release-age or missing --ignore-scripts.

Or `_ALKILN_IGNORE_MIN_AGE` instead of `_ALKLIN_SECURITY_SENSITIVE_NPM_ARGS`

- [x] Or just have the internal flag and add flexibility later if needed.

**05/27 10:00**

I've at least figured out why the initial tests passed while the new ones fail. We do have a bug in our new code (which I still need to fully ferret out). The earlier tests passed because `--min-release-age` forced our GitHub tests to test with an older version of our code, which didn't contain the bug.

The bug has to do with the folder in which our artifacts get created. I believe it's being created at the same level as the GitHub artifacts folder. I think we need to manage the `cwd` there with more care. Switching it temporarily, then switching back.

The symptom is that the `action.yml` is finding 2 folders when looking the artifacts folder path to give to the user.

---

Original:

An ALKiln test run failed. See the action at https://github.com/SuffolkLITLab/ALKiln/actions/runs/26467595245.

Potentially due to a change in GitHub?

This happens only when we try to use the outputs. [If so, it's unsurprising that] our passing tests [didn't] catch the problem. [Development branch at [https://github.com/SuffolkLITLab/ALKiln/tree/fix_action_outputs](https://github.com/SuffolkLITLab/ALKiln/tree/fix_action_outputs). [GitHub page containing action output example](https://docs.github.com/en/actions/tutorials/create-actions/create-a-composite-action).]

TODO:
- [x] Add code to use test files output regardless of success or failure, though avoid making an issue when successful. [See if that fails too.]
- [x] Figure out why we have no artifacts on test failure.
- [x] Fix bug.

> [8](https://github.com/SuffolkLITLab/ALKiln/actions/runs/26467595245/job/77931978483#step:7:899)
> 💡 ALK0184 INFO: Prepare ALKiln output file paths
> Test results artifacts folder is alkiln-2026-05-26 at 18h41m00sUTC/
> alkiln-2026-05-26 at 18h41m23sUTC/
> alkiln-2026-05-26 at 18h41m00sUTC/
> alkiln-2026-05-26 at 18h41m23sUTC/report.txt is missing
> alkiln-2026-05-26 at 18h41m00sUTC/
> alkiln-2026-05-26 at 18h41m23sUTC/report_log.txt is missing
> alkiln-2026-05-26 at 18h41m00sUTC/
> alkiln-2026-05-26 at 18h41m23sUTC/unexpected_results.txt is missing
> alkiln-2026-05-26 at 18h41m00sUTC/
> alkiln-2026-05-26 at 18h41m23sUTC/debug_log.txt is missing
> Error: Unable to process file command 'output' successfully.
> Error: Invalid format 'alkiln-2026-05-26 at 18h41m23sUTC/'
> 
> ### Results:
> 
? unexpected_results.txt does not exist. Something other than the interview tests may have failed.

### LINKED ISSUE #1062: Should we turn off GitHub+You `push` tests for the Suffolk dev server?

plocket:
In this PR, I have:

* N/A, previous tests are sufficient ~~Added tests for any new features or bug fixes (if relevant)~~

### Reason for this PR

Replaces/closes #1064, which was attempting to address the same problems.

- Fixes https://github.com/SuffolkLITLab/ALKiln/issues/1055, restore custom artifacts path argument to allow environments like action.yml to make and reuse their own artifacts path value in environments where it's hard to pass around return values. Closes #1055.
- Addresses #1088, new issues with puppeteer installs in GitHub actions being
  unable to find Chrome executables.
- Closes #1062, avoids testing on our dev server with every push. Uses cron job instead.

Replaces/closes #1064 
Fixes #1055
Closes #1062
Addresses #1088

[See details in the issues linked above. Many actions run on many branches with many variations of the code. It's a lot to parse through, but we've found the complexity of these problems hard to simplify.]


BryceStevenWilley (approved):
LGTM!

